PR #2, opened 2022-02-21
Add identity form flow

ibolton336:


pkg/client/public/locales/es/translation_old.json
```diff
@@ -0,0 +1,16 @@
...
+    "dialog": {
+        "title": {
+            "newIdentity": "",
+            "updateIdentity": ""
+        }
+    },
+    "message": {
+        "selectIdentityType": ""
+    }
+}
```
gildub, line 16: `translation_old.json` files don't have to be checked in anymore as they have been added to `.gitignore`

gildub (commented):
Looks good.

Just a couple of things.
When editing a credential the values for type and below are blank, not sure if that's WIP or to be addressed by following PR.
Also see inline a comment about translation_old.json files

gildub (approved):

PR #4, opened 2022-02-22
Fix entrypoint server dir

fbladilo:
- Exec node call should be within pkg/server instead of server

@gildub Not sure if we should be using `exec npm run -s start` instead of `exec node index.js` to start the server there , is there any difference?

### DIFF
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -38,5 +38,5 @@ fi
 #   # envsubst '${SSO_REALM} ${SSO_CLIENT_ID}' < ./build/keycloak.json.template > ./build/keycloak.json
 # fi
 
-cd server
+cd pkg/server
 exec node index.js


gildub:
@fbladilo, 

Thanks for catching that.

The "pkg" path is not really necessary and will ultimately disappear on the target.
Because we're not using npm workspace we have to handle each pkg sub-project (for client and server) and that makes it more cumbersome than it needs to be.
We'll remove the pkg from the target later on.

Regarding `exec npm run -s start` vs `exec node index.js`, npm is always installed along node as part of the nodejs bundle, so we could use npm meanwhile in production we target nodejs runtime to run the Express server.

gildub (approved):

gildub (approved):
LGTM

PR #10, opened 2022-02-24
Settings api

ibolton336:
Add api support for mvn,svn, & git settings configuration from the repositories nav section.

### DIFF
--- a/pkg/client/src/app/api/rest.tsx
+++ b/pkg/client/src/app/api/rest.tsx
@@ -61,7 +61,8 @@ export const IDENTITIES = "/identities";
 
 export const SETTINGS = "/settings";
 
-const headers = { Accept: "application/hal+json" };
+const legacyHeaders = { Accept: "application/hal+json" };
+const headers = { Accept: "application/json" };
 
 type Direction = "asc" | "desc";
 
@@ -132,7 +133,9 @@ export const getBusinessServices = (
   };
 
   const query: string[] = buildQuery(params);
-  return APIClient.get(`${BUSINESS_SERVICES}?${query.join("&")}`, { headers });
+  return APIClient.get(`${BUSINESS_SERVICES}?${query.join("&")}`, {
+    legacyHeaders,
+  });
 };
 
 export const deleteBusinessService = (id: number | string): AxiosPromise => {
@@ -214,7 +217,7 @@ export const getStakeholders = (
   };
 
   const query: string[] = buildQuery(params);
-  return APIClient.get(`${STAKEHOLDERS}?${query.join("&")}`, { headers });
+  return APIClient.get(`${STAKEHOLDERS}?${query.join("&")}`, { legacyHeaders });
 };
 
 export const createJobFunction = (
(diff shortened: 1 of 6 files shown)


pkg/client/src/app/pages/repositories/Mvn.tsx
```diff
@@ -36,7 +104,7 @@ export const RepositoriesMvn: React.FunctionComponent = () => {
       </PageSection>
       <PageSection>
         <Card>
-          <CardBody>
+          {/* <CardBody>
```
gildub, line 107: Is this needed ?

pkg/client/src/app/pages/repositories/Mvn.tsx
```diff
@@ -48,14 +116,14 @@ export const RepositoriesMvn: React.FunctionComponent = () => {
             <Button variant="link" isInline>
               Clear repository
             </Button>
-          </CardBody>
+          </CardBody> */}
           <CardBody>
```
gildub, line 120: Don't we need error handling here ?

pkg/client/src/app/pages/repositories/Svn.tsx
```diff
@@ -35,7 +72,7 @@ export const RepositoriesSvn: React.FunctionComponent = () => {
               className="repo"
               label="Consume insecure Subversion repositories"
               aria-label="Insecure Subversion Repositories"
-              isChecked={isInsecure}
+              isChecked={svnInsecureSetting === true ? true : false}
               onChange={onChange}
             />
           </CardBody>
```
gildub, line 78: Don't we need error handling here too ?

gildub (commented):
Please see inline.
The rest.tsx file is becoming really difficult to follow don't you think it might be time to make it more modular?
Git.tsx, Svn.tsx and Mvn.tsx are very similar, where the first 2 ones are the same and seems good candidates for a share component.

pkg/client/src/app/api/rest.tsx
```diff
@@ -753,6 +756,7 @@ export const getBulkCopyAssessment = (
 };
 
 const identityHeaders = { Accept: "application/json" };
+const settingsHeaders = { Accept: "application/json" };
```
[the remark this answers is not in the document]
ibolton336: +! sounds good

pkg/client/src/app/pages/repositories/Mvn.tsx
ibolton336: I will add a TODO for a stretch goal for this. As jeff mentioned.

pkg/client/src/app/pages/repositories/Mvn.tsx
ibolton336: Good point

ibolton336:
> 

I agree but I don't think we have the bandwidth to take on tech debt for now. Let's revisit when things calm down if that makes sense?

> Please see inline. The rest.tsx file is becoming really difficult to follow don't you think it might be time to make it more modular? Git.tsx, Svn.tsx and Mvn.tsx are very similar, where the first 2 ones are the same and seems good candidates for a share component.

gildub (approved):
LGTM, besides the technical debt we're adding here.

ibolton336:
> LGTM, besides the technical debt we're adding here.

Not sure I added any here since those files were already in existence & not modular.

PR #23, opened 2022-02-28
add any types for identity form & add useFetchIdentities hook

ibolton336:


gildub (changes requested):
Value and fieldName seems to be strings
Isn't schema of type formik schema ?
Could you please change them accordingly ?

gildub (approved):
Not sure about some comments but we can address them later 
otherwise LGTM.

PR #25, opened 2022-03-02
Integrate UI pagination into global app table component

ibolton336:
- Converts identity to new approach using existing app table control
- Removes server side pagination from entire app on presentation layer side. Placeholders added for future work. Identities is done just need to propagate across the rest of the app. 
- Includes filter,sort, pagination for stakeholders/stakeholders groups/jobfunctions/tags/applications. TBD Reports?


gildub (approved):
LGTM

PR #27, opened 2022-03-07
Update identity form to include default xsd file to check against

ibolton336:


pkg/client/src/app/pages/identities/components/identity-form.tsx
```diff
@@ -29,6 +29,8 @@ import {
 import "./identity-form.css";
 const xmllint = require("xmllint");
 const { XMLValidator } = require("fast-xml-parser");
+const schema = require('./schema.xsd');
```
gildub, line 32: Could we preferably use import instead ?
You might need to add entry to src/typings.d.ts

gildub (changes requested):
I just tested adding a new 'credentials -> Maven settings file'
When I upload a settings.xml file, I always get a warning saying You should select a valid settings.xml file.
Meanwhile it doesn't block me from saving the credential.
When editing, the type (source, maven, proxy) is not defined and the maven file is not there.

pkg/client/src/app/pages/identities/components/identity-form.tsx
```diff
@@ -96,6 +148,14 @@ export const IdentityForm: React.FC<IdentityFormProps> = ({
     description: string()
       .trim()
       .max(250, t("validation.maxLength", { length: 250 })),
+    kind: object().shape({
+      value: string().min(1, "value required").required(),
+      toString: object().required(),
+    }),
```
gildub, line 154: This is a detail, `.shape` could be removed as this is equivalent:
```
kind: object().shape({
      value: string().min(1, "value required").required(),
      toString: object().required(),
    }),
```
Actually the whole block can be replaced with:

``` 
const validationSchema = object({
    name: string()
      .trim()
      .required(t("validation.required"))
      .min(3, t("validation.minLength", { length: 3 }))
      .max(120, t("validation.maxLength", { length: 120 })),
    description: string()
      .trim()
      .max(250, t("validation.maxLength", { length: 250 })),
    kind: object().shape({
      value: string().min(1, "value required").required(),
      toString: object().required(),
    }),
    settings: string().when("kind.value", {
      is: "maven",
      then: string().required("Must upload xml settings file"),
    }),
  });

```

pkg/client/src/app/pages/identities/components/identity-form.tsx
```diff
@@ -74,17 +126,17 @@ export const IdentityForm: React.FC<IdentityFormProps> = ({
     encrypted: identity?.encrypted || "",
     id: identity?.id || 0,
     key: identity?.key || "",
-    kind: { value: "", toString: () => "" },
-    userCredentials: { value: "", toString: () => "" },
+    kind: kindInitialValue,
```
gildub, line 129: I might be missing something but why kind cannot be just initialized with `""` ?
Having the field mandatory should suffice because the user can only select its value through a drop-down box which we control the value of. In which case we wouldn't need `toString()`

pkg/client/src/app/pages/identities/components/identity-form.tsx
```diff
@@ -330,7 +386,7 @@ export const IdentityForm: React.FC<IdentityFormProps> = ({
                   },
                   {
                     value: "scm",
-                    toString: () => `SCM Private Key/Passphrase`,
+                    toString: () => `source Private Key/Passphrase`,
```
gildub, line 389: Shouldn't it be Source instead of source ?

gildub (changes requested):
See inline.
When testing I found out:

* After creation of a new credential the type is not displayed in the credential table even after page refresh.
* When editing a proxy type of credential the first time after it's creation, the type is not displayed (not selected) and when the type is 'proxy', the user/password details are blank. Entering the user/passwd a second time and saving it works.
* When editing a maven type credential, changing the attached file doesn't work, one can select another file and save but the modification is not accounted for.
* When creating another maven type credential, doesn't work properly as the xml file name displayed is the one of the first created maven type credential saved  
* All the type select drop-down box are showing an unnecessary cancel (x) icon.

pkg/client/src/app/pages/identities/components/identity-form.tsx
ibolton336: > This is a detail, `.shape` could be removed as this is equivalent:
> 
> ```
> kind: object().shape({
>       value: string().min(1, "value required").required(),
>       toString: object().required(),
>     }),
> ```
> 
Is this not what is already there? Did you mean to remove "shape()" in your example ablove?
> Actually the whole block can be replaced with:
> 
> ```
> const validationSchema = object({
>     name: string()
>       .trim()
>       .required(t("validation.required"))
>       .min(3, t("validation.minLength", { length: 3 }))
>       .max(120, t("validation.maxLength", { length: 120 })),
>     description: string()
>       .trim()
>       .max(250, t("validation.maxLength", { length: 250 })),
>     kind: object().shape({
>       value: string().min(1, "value required").required(),
>       toString: object().required(),
>     }),
>     settings: string().when("kind.value", {
>       is: "maven",
>       then: string().required("Must upload xml settings file"),
>     }),
>   });
> ```

gildub: Ah yes, you're right, in first example without shape.

PR #30, opened 2022-03-10
Custom rules table with xml files uploads and windup parsing

gildub:
A valid xml file containing windup rules is entirely processed to provide source, target and rules tags.

Notes:
- Upload File - Multiple component has the implementation for checking the file has a valid xml content and that the xml is valid per the requirements. Meanwhile it only logs it for now. Passing the file status back to is coming in following PR.
- Need to implement delete action (trash icon)

image

image


pkg/client/src/app/pages/applications/analysis-wizard/components/add-custom-rules.tsx
```diff
@@ -71,37 +77,73 @@ export const AddCustomRules: React.FunctionComponent<IAddCustomRules> = ({
...
+      allowBooleanAttributes: true,
+    });
+
+    if (validationObject === true) {
+      isXML = true;
+      if (isSchemaValid(payload)) {
+        isXSD = true;
+      }
+    }
+    console.log("XML :", isXML);
```
ibolton336, line 107: remove

pkg/client/src/app/pages/applications/analysis-wizard/components/add-custom-rules.tsx
```diff
@@ -71,37 +77,73 @@ export const AddCustomRules: React.FunctionComponent<IAddCustomRules> = ({
...
+
+    let isXML = false;
+    let isXSD = false;
+
+    const validationObject = XMLValidator.validate(payload, {
+      allowBooleanAttributes: true,
+    });
+
+    if (validationObject === true) {
+      isXML = true;
```
ibolton336, line 102: Should we use react state for these values?

pkg/client/src/app/pages/applications/analysis-wizard/components/add-custom-rules.tsx
```diff
@@ -71,37 +77,73 @@ export const AddCustomRules: React.FunctionComponent<IAddCustomRules> = ({
...
   const handleReadSuccess = (data: string, file: File) => {
-    setReadFileData((prevReadFiles) => [
-      ...prevReadFiles,
-      { data, fileName: file.name, loadResult: "success" },
-    ]);
+    validateXMLFile(data);
+
+    const fileList = [
+      ...readFileData,
+      { data, fileName: file.name, loadResult: "success" } as IReadFile,
```
ibolton336, line 129: Why are we asserting here?

pkg/client/src/app/pages/applications/analysis-wizard/custom-rules.tsx
```diff
@@ -32,15 +32,71 @@ import {
...
+      }
+
+      return [
+        {
+          name: file.fileName,
+          source: source,
+          target: target,
+          total: rulesCount,
+        },
+      ] as Rule[];
```
ibolton336, line 90: Can we set a return type instead of asserting here?

pkg/client/src/app/pages/applications/analysis-wizard/custom-rules.tsx
```diff
@@ -32,15 +32,71 @@ import {
...
+    };
+
+    let rules: Rule[] = [];
+    customRulesFiles.forEach((file) => {
+      if (file.data) rules = [...rules, ...getRules(file)];
```
ibolton336, line 95: Can we use curly brackets for if statements just for consistency's sake.

ibolton336 (changes requested):

pkg/client/src/app/pages/applications/analysis-wizard/custom-rules.tsx
gildub: This is proper js/ts syntax and don't think there is an eslint rule to enforce that anyway.

ibolton336: Ok. I understand that it is valid - I just believe that we should be consistent with either using them or not. I would argue that it makes code more readable to have curly brackets. Semicolons are not syntactically required, but we add them for legibility.

pkg/client/src/app/pages/applications/analysis-wizard/components/add-custom-rules.tsx
gildub: Because the object is created on the fly. 
I suppose I could create a sub-type or assert

pkg/client/src/app/pages/applications/analysis-wizard/custom-rules.tsx
gildub: Yes we can. But why create an object just to return it then ?

ibolton336: https://medium.com/swlh/why-you-should-avoid-type-assertions-in-typescript-5494e3d04dd

pkg/client/src/app/pages/applications/analysis-wizard/components/add-custom-rules.tsx
ibolton336: https://medium.com/swlh/why-you-should-avoid-type-assertions-in-typescript-5494e3d04dd

pkg/client/src/app/pages/applications/analysis-wizard/custom-rules.tsx
gildub: I agree we could argue about it for a long time as this is quite opinionated. 
We could be consistent by agreeing to have no brackets when is only one statement.
It effectively makes the code more readable which is paramount along with consistency.
Please let me know if you being consistent that way. In which case we can use brackets for conditional statements that are constituted of single statement.

pkg/client/src/app/pages/applications/analysis-wizard/components/add-custom-rules.tsx
gildub: This is temporary until we inject the xml parsing inside into the file load validation. 
Let me wrap it with a comment.

pkg/client/src/app/pages/applications/analysis-wizard/components/add-custom-rules.tsx
gildub: There is no need for state as the validation process is not transient across rendering.

pkg/client/src/app/pages/applications/analysis-wizard/components/add-custom-rules.tsx
gildub: That makes sense and this is really important and trying to write lighter code defeated the purpose of type in this case. Thank you for bringing my attention to this.

ibolton336 (approved):

ibolton336:
LGTM

PR #31, opened 2022-03-14
Introduce react-query for proxy fetch and mutate

ibolton336:
- Introduce react-query as a pattern for proxies. Can adopt this pattern for usage across the app as api interactions become more complex.


gildub (changes requested):
Looks good, there is an installation issue because `pkg/client/package-lock.json` is version 1.
Please let's keep it in version 2.

gildub (changes requested):
Testing show the https proxy switch behavior is broken, can't do anything with it.

gildub (approved):
Looks good.
There is an existing issue with proxies credentials being saved but not reflected in UI.
Let's create an issue.

PR #33, opened 2022-03-15
Advanced options

gildub:
image

image


pkg/client/src/app/pages/applications/analysis-wizard/analysis-wizard.tsx
```diff
@@ -205,11 +197,25 @@ export const AnalysisWizard: React.FunctionComponent<IAnalysisWizard> = ({
           component: (
             <CustomRules
               customRulesFiles={customRulesFiles}
-              setValue={(val) => setValue("customRulesFiles", val)}
+              sources={sources}
+              setCustomRulesFiles={(val) => setValue("customRulesFiles", val)}
```
ibolton336, line 201: Can we change these function names & values to something more descriptive than setValue, val ?

ibolton336 (changes requested):

pkg/client/src/app/pages/applications/analysis-wizard/options.tsx
```diff
@@ -31,24 +120,97 @@ export const Options: React.FunctionComponent<IOptions> = ({ targets }) => {
...
+              setSelectSourcesOpen(!isSelectSourcesOpen);
+            }}
+            onClear={() => {
+              setSources([]);
+            }}
+          >
+            {defaultSources.map((source, index) => (
+              <SelectOption key={index} component="button" value={source} />
+            ))}
+          </Select>
```
ibolton336, line 170:
```suggestion
           <SimpleSelect
            variant={SelectVariant.typeaheadMulti}
            aria-label="Select targets"
            selections={targets}
            isOpen={isSelectTargetsOpen}
            onSelect={(e:any, selection:string) => {
              if (!targets.includes(selection))
                setTargets([...targets, selection]);
              else setTargets(targets.filter((target) => target !== selection));
              setSelectTargetsOpen(!isSelectTargetsOpen);
            }}
            onToggle={() => {
              setSelectTargetsOpen(!isSelectTargetsOpen);
            }}
            onClear={() => {
              setTargets([]);
            }}
          >
            {defaultTargets.map((target, index) => (
              <SelectOption key={index} component="button" value={target} />
            ))}
          </SimpleSelect>
```

pkg/client/src/app/pages/applications/analysis-wizard/options.tsx
```diff
@@ -1,27 +1,116 @@
 import * as React from "react";
```
ibolton336, line 1: The file name 'options' is somewhat terse. I think we should at least change to options-step or something similar to convey what it is referring to. Tough to follow what is going into this file.

ibolton336 (changes requested):

pkg/client/src/app/pages/applications/analysis-wizard/options.tsx
```diff
@@ -31,24 +120,97 @@ export const Options: React.FunctionComponent<IOptions> = ({ targets }) => {
...
             }}
           >
-            {targetsOptions}
+            {defaultTargets.map((target, index) => (
+              <SelectOption key={index} component="button" value={target} />
+            ))}
           </Select>
         </FormGroup>
+        <FormGroup label="Sources" fieldId="sources">
+          <Select
```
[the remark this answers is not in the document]
gildub: Sure. I can't see the difference in the example you provided below but that would be clearer when I add it.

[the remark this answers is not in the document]
gildub: Actually the change to SimpleSelect requires some refactoring. Since this is working fine then we can create an issue for a follow-up PR.

pkg/client/src/app/pages/applications/analysis-wizard/options.tsx
```diff
@@ -1,27 +1,116 @@
...
+  "websphere",
+];
+
+export const Options: React.FunctionComponent<IOptions> = ({
+  targets,
+  sources,
+  setTargets,
+  excludedRulesTags,
+  setExcludedRulesTags,
+  setSources,
```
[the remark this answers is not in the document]
gildub: This is working right now, I'm more than happy to do that in follow up PR, but the priority is to get the wizard working.

pkg/client/src/app/pages/applications/analysis-wizard/analysis-wizard.tsx
```diff
@@ -153,34 +142,27 @@ export const AnalysisWizard: React.FunctionComponent<IAnalysisWizard> = ({
...
         },
         {
           name: "Set targets",
-          component: (
-            <SetTargets
-              targets={targets}
-              setTargets={(val) => setValue("targets", val)}
-            />
-          ),
+          component: <SetTargets targets={targets} />,
```
ibolton336, line 165: Looking much better! Can we pull the actual form values from the context too to avoid this prop bubbling  overhead? Os there a way to type the form values pulled in from the context api?

ibolton336 (changes requested):

pkg/client/src/app/pages/applications/analysis-wizard/analysis-wizard.tsx
```diff
@@ -202,35 +162,33 @@ export const AnalysisWizard: React.FunctionComponent<IAnalysisWizard> = ({
...
       name: "Review",
-      component: <Review applications={applications} getValues={getValues} />,
+      component: <Review applications={applications} />,
       nextButtonText: "Run",
     },
   ];
-
-  console.log(watch());
-
+  console.log(methods.watch());
```
ibolton336, line 179: remove

pkg/client/src/app/pages/applications/analysis-wizard/custom-rules.tsx
```diff
@@ -97,7 +96,7 @@ export const CustomRules: React.FunctionComponent<ICustomRules> = ({
     });
 
     return rules.flat();
-  }, [customRulesFiles]);
+  }, [customRulesFiles, sources, setValue]);
```
ibolton336, line 99: Why add the set value handler here?

pkg/client/src/app/pages/applications/analysis-wizard/review.tsx
```diff
@@ -10,28 +10,35 @@ import {
...
+    "depsAll",
+    "Application and all dependencies, including known Open Source libraries",
+  ],
+  ["depsSelect", "list of packages to be analyzed manually"],
+]);
+
+export const Review: React.FunctionComponent<IReview> = ({ applications }) => {
+  const { getValues } = useFormContext();
+  const mode: string = getValues("mode");
+  const targets: string[] = getValues("targets");
```
ibolton336, line 34: Is there a centralized location where we can put these type definitions? 
```suggestion
  const { getValues, setValue } = useFormContext<IAnalysisWizardFormValues>();
  const { targets, sources } = getValues();```

**Review by ibolton336 — CHANGES_REQUESTED** [review]

File: pkg/client/src/app/pages/applications/analysis-wizard/analysis-wizard.tsx [hunk 10]
```diff
@@ -27,13 +27,12 @@ interface IAnalysisWizard {
 export interface IFormValues {
```
ibolton336, line 27: Update to IAnalysisWizardFormValues so we can reuse this export with the context api.

ibolton336 (changes requested):

pkg/client/src/app/pages/applications/analysis-wizard/custom-rules.tsx
gildub: Because it's used above in the useMemo().

ibolton336: isn't it already defined in the enclosing scope? No need to add the function definition to this array as it is not ever changing.

gildub: I get an error if we don't:
`React Hook React.useMemo has a missing dependency: 'setValue'. Either include it or remove the dependency array.eslintreact-hooks/exhaustive-deps`

ibolton336 (approved):
Nice job integrating with the better dev flow offered by react-hook-forms 👍

PR #36, opened 2022-03-16
Update add app modal

ibolton336:
- Update repository fields on the add app modal form 
- Add conditional validation for the application repo url field
- Disables binary fields until the api field exists for binary.


gildub (approved):
LGTM

PR #39, opened 2022-03-16
Fix proxy identity initial values

ibolton336:
- Show selected identity when loading existing modified proxies 
- Update associated identity when de-selecting switch for identities


gildub (approved):
LGTM

PR #40, opened 2022-03-17
Update repo url validation

ibolton336:
- Allow URL without branch or path but check valid url if any input exists for url field.


jortel (approved):
LGTM

PR #43, opened 2022-03-19
Custom file handler

ibolton336:


pkg/client/src/app/api/models.tsx
```diff
@@ -342,9 +342,9 @@ export interface TaskData {
     withDeps: boolean;
     artifact: string;
   };
-  targets: string[];
-  sources: string[];
-  scope: {
+  targets?: string[];
+  sources?: string[];
+  scope?: {
```
gildub, line 347: I don't think this is needed. Actually the backend has those fields mandatory.
* targets cannot be empty.
* sources can be empty but it's an empty array in that case.
* scope is mandatory where withKnown is false by default and packages.includes = [] by default.

pkg/client/src/app/pages/applications/analysis-wizard/analysis-wizard.tsx
```diff
@@ -198,49 +245,143 @@ export const AnalysisWizard: React.FunctionComponent<IAnalysisWizard> = ({
...
+  const getNextStep = (activeStep: any, callback?: any) => {
+    if (activeStep.id === 1 && mode === "Upload a local binary") {
+      setTimeout(() => {
+        callback();
+      });
+    } else {
+      setTimeout(() => {
+        callback();
+      });
+    }
```
mturley, line 372: Both the if/else branches here do the same thing?

pkg/client/src/app/pages/applications/analysis-wizard/analysis-wizard.tsx
```diff
@@ -63,11 +72,15 @@ export const AnalysisWizard: React.FunctionComponent<IAnalysisWizard> = ({
...
+    if (!isInitTasks) {
+      initTasks();
+    }
 
-  const [isInitTasks, setInitTasks] = React.useState(false);
-  const [createdTasks, setCreatedTasks] = React.useState<Array<Task>>([]);
+    return () => {
+      setInitTasks(false);
+    };
+  }, []);
```
gildub, line 83: I don't see why we need useEffect (I actually initially tried that path but came back from it).
I React state isInitTasks allows us to flag if the task(s) have been created and if not then this is done only once.

I actually get an error on here: React Hook useEffect has missing dependencies: 'initTasks' and 'isInitTasks'. Either include them or remove the dependency array.

pkg/client/src/app/pages/applications/analysis-wizard/analysis-wizard.tsx
```diff
@@ -36,8 +44,9 @@ export interface IAnalysisWizardFormValues {
   withKnown: string;
   includedPackages: string[];
   excludedPackages: string[];
-  customRulesFiles: IReadFile[];
+  customRulesFiles: any;
```
gildub, line 47: Why change for any ?

ibolton336: Was running into this when having a nested ts definition within the form values. Will adjust to include definition within the form values interface. https://github.com/react-hook-form/react-hook-form/discussions/7764

pkg/client/src/app/pages/applications/analysis-wizard/analysis-wizard.tsx
ibolton336: Ok I will revert the change

pkg/client/src/app/pages/applications/analysis-wizard/analysis-wizard.tsx
```diff
@@ -91,6 +104,24 @@ export const AnalysisWizard: React.FunctionComponent<IAnalysisWizard> = ({
...
+  const {
+    mode,
+    targets,
+    sources,
+    withKnown,
+    includedPackages,
+    excludedPackages,
+    customRulesFiles,
+    excludedRulesTags,
+    createdTasks,
```
gildub, line 121: `createdTasks` are only used in this file and we can pass the task Id to `UploadBinary` which always handles a single application case anyway.

pkg/client/src/app/pages/applications/analysis-wizard/components/upload-binary.tsx
```diff
@@ -11,21 +11,54 @@ import {
...
+import { IAnalysisWizardFormValues, IReadFile } from "../analysis-wizard";
+import { Task } from "@app/api/models";
 
-export interface IReadFile {
-  fileName: string;
-  data?: string;
-  loadResult?: "danger" | "success";
-  loadError?: DOMException;
+interface IUploadBinary {
+  createdTasks: Array<Task>;
```
gildub, line 20: We only need the ID of the task involved.
=

gildub (approved):
LGTM

PR #44, opened 2022-03-21
Set ouptut, mode.artifact and rules path uploads

gildub:
Completely set up custom rules:
- Uploads custom rules files
- Makes rules optional for task.data

Adds generic rest request for file uploads to be used by both binary and custom rules

Harmonize notifications for Task creation (wizard initialization), custom rules uploads, task data set (update) and submission.


ibolton336 (approved):
LGTM

PR #46, opened 2022-03-21
Loosen types on field values to fix ts error

ibolton336:
Still need to find a fix for deep type checking against field values. This fixes the circular dep error that keeps popping up. https://github.com/react-hook-form/react-hook-form/discussions/7764

### DIFF
--- a/pkg/client/src/app/pages/applications/analysis-wizard/set-mode.tsx
+++ b/pkg/client/src/app/pages/applications/analysis-wizard/set-mode.tsx
@@ -6,7 +6,7 @@ import {
   SelectOption,
   SelectVariant,
 } from "@patternfly/react-core";
-import { useFormContext } from "react-hook-form";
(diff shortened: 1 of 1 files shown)


pkg/client/src/app/pages/applications/analysis-wizard/set-mode.tsx
```diff
@@ -6,7 +6,7 @@ import {
   SelectOption,
   SelectVariant,
 } from "@patternfly/react-core";
-import { useFormContext } from "react-hook-form";
+import { FieldValues, useFormContext } from "react-hook-form";
```
gildub, line 9: This not used.

gildub: Never mind we can fix it during next pass.

gildub (approved):
LGTM

PR #48, opened 2022-03-21
Fix proxy save button disabled state when value changes 

ibolton336:
Previously, the save button was not changing when updating values/saving. This was in part due to proxies not being properly refetched after a mutation.


pkg/client/src/app/pages/proxies/proxy-form.tsx
```diff
@@ -313,6 +315,19 @@ export const ProxyForm: React.FC<ProxyFormProps> = ({
     formik.handleChange(event);
   };
 
+  const [hasUpdate, setHasUpdate] = useState(false);
+  useEffect(() => {
+    console.log(formik.values, httpProxy);
```
gildub, line 320: Do we need to keep that ?

PR #51, opened 2022-03-23
Display report when analysis is succeeded

gildub:
- Provide link to windup report when Analysis is successful
- Filters tasks with only latest created task when several task exist for a same application.
- Blocks Analyze button when selected applications are already engaged in "Created", "Running" or "Ready"
- Harmonize Task state and matching Analysis state displayed by UI
- Renames IconStatusAssessment to IconStatus as it's shared by several components


ibolton336 (approved):
LGTM

PR #61, opened 2022-03-29
Bug fix for showing kind on identity edit

ibolton336:


gildub (commented):
When editing a source control the user credentials are not matching.
Also I cannot upload a Maven file anymore, I get a xml validation error

ibolton336:
> When editing a source control the user credentials are not matching. Also I cannot upload a Maven file anymore, I get a xml validation error

The validation error is because it does not match the schema provided.

Issue #63, opened 2022-03-30
Allow single user mode


fbladilo:
We need UI to honor the AUTH_REQUIRED pod environment variable , essentially this will cause Tackle to enable/disable keycloak auth and allow a user to just run as admin without auth. 

When AUTH_REQUIRED is set (false) at the operator level keycloak will not be deployed , so environment variables SSO_* should not be required when AUTH_REQUIRED is set to false.

Also , for naming consistency the SSO_* UI env variables should match what the hub is using : 
```
            - name: KEYCLOAK_REALM
              value: "{{ keycloak_sso_realm }}"
            - name: KEYCLOAK_HOST
              value: "{{ keycloak_sso_host }}"
            - name: KEYCLOAK_CLIENT_ID
              value: "{{ keycloak_sso_hub_client_id }}"
            - name: KEYCLOAK_CLIENT_SECRET
              valueFrom:
                secretKeyRef:
                  name: "{{ hub_secret_name }}"
                  key: keycloak_client_secret
```
For reference also see : https://github.com/konveyor/tackle2-operator/issues/33

PR #64, opened 2022-03-30
Single user ui changes

ibolton336:
- Update env var names
- Conditionally render the keycloak provider wrapper based on an env var. 
- Updated the PR to follow the forklift env var approach for supporting encrypted env vars in production loaded at runtime & loading development mode env vars at build time.
Pending operator changes: https://github.com/konveyor/tackle2-ui/issues/63

Depends upon https://github.com/konveyor/tackle2-operator/pull/43


fbladilo:
@ibolton336 I don't think we are using KEYCLOAK_REALM or KEYCLOAK_CLIENT_ID that I could find , perhaps it would make sense to remove them if they are useless. Also , KEYCLOAK_HOST makes sense more as KEYCLOAK_SERVER_URL , still follows naming but I think better represents what it is , since it is a URL that contains the host.

fbladilo (commented):
We need a conditional here for AUTH_REQUIRED , basically : 
```
if [[ $AUTH_REQUIRED == "true" ]]; then

	if [[ -z "$KEYCLOAK_SERVER_URL" ]]; then
	  echo "You must provide KEYCLOAK_SERVER_URL environment variable" 1>&2
	  exit 1
	fi
fi
```
I believe that should work fine since we need to check for KEYCLOAK_HOST only when AUTH_REQUIRED evals true

gildub:
@fbladilo, please let us know when you have an operator PR covering the change from  SSO_SERVER_URL to KEYCLOAK_SERVER_URL

entrypoint.sh
```diff
@@ -12,20 +12,11 @@ if [[ -z "$PATHFINDER_URL" ]]; then
...
-  echo "You must provide SSO_REALM environment variable" 1>&2
-  exit 1
-fi
-
-if [[ -z "$SSO_SERVER_URL" ]]; then
-  echo "You must provide SSO_SERVER_URL environment variable" 1>&2
-  exit 1
+if [[ $AUTH_REQUIRED == "true" ]]; then
+  if [[ -z "$KEYCLOAK_SERVER_URL" ]]; then
+    echo "You must provide SSO_SERVER_URL environment variable" 1>&2
```
gildub, line 17: Actually we need to change the message itself as well.

entrypoint.sh
```diff
@@ -12,20 +12,11 @@ if [[ -z "$PATHFINDER_URL" ]]; then
   exit 1
 fi
 
-if [[ -z "$SSO_REALM" ]]; then
-  echo "You must provide SSO_REALM environment variable" 1>&2
-  exit 1
-fi
```
gildub, line 18: @ibolton336,

KEYCLOAK_REALM or KEYCLOAK_CLIENT_ID are legacy names which have been replaced by SSO_REALM and SSO_CLIENT.

At the moment we're using hard coded values in https://github.com/konveyor/tackle2-ui/blob/main/pkg/client/src/app/keycloak.ts.

Tackle 1 was generating a json file bundled with the app:
 https://github.com/konveyor/tackle-ui/blob/main/public/keycloak.json.template

Since those value are provided dynamically by the operator I think we  need Express to serve them.

entrypoint.sh
```diff
@@ -12,20 +12,11 @@ if [[ -z "$PATHFINDER_URL" ]]; then
...
+  if [[ -z "$KEYCLOAK_SERVER_URL" ]]; then
+    echo "You must provide KEYCLOAK_SERVER_URL environment variable" 1>&2
+    exit 1
+  fi
 fi
-
-if [[ -z "$SSO_CLIENT_ID" ]]; then
-  echo "You must provide SSO_CLIENT_ID environment variable" 1>&2
-  exit 1
-fi
```
gildub, line 28: I think we need to keep this for same reason as above.

gildub (commented):
We need to keep SSO_CLIENT_ID and SSO_REALM. See inline.

entrypoint.sh
fbladilo: @gildub Do we have to keep the naming as SSO_xxxx or can they be replaced by KEYCLOAK_CLIENT_ID and KEYCLOAK_REALM ? The reason I ask is because we are getting rid of all SSO naming everywhere else.

gildub: @fbladilo, Yes that's absolutely fine. Better have it consistent across the board

entrypoint.sh
```diff
@@ -12,20 +12,21 @@ if [[ -z "$PATHFINDER_URL" ]]; then
...
   exit 1
 fi
 
-if [[ -z "$SSO_SERVER_URL" ]]; then
-  echo "You must provide SSO_SERVER_URL environment variable" 1>&2
-  exit 1
-fi
-
-if [[ -z "$SSO_CLIENT_ID" ]]; then
+if [[ -z "$KEYCLOAK_CLIENT_ID" ]]; then
```
fbladilo, line 20: KEYCLOAK_CLIENT_ID and KEYCLOAK_REALM should be contained within the AUTH_REQUIRED conditional below, operator will only set these variables in the pod when AUTH_REQUIRED is true, otherwise they will be ignored but the entrypoint should not fail (AUTH_REQUIRED=false case).

pkg/client/public/locales/es/translation.json
```diff
@@ -233,7 +230,6 @@
         "risk": "Riesgo",
         "scheduled": "Programado",
         "select": "Seleccione",
-        "selected": "Seleccionado",
```
gildub, line 236: The commit doesn't need this change

pkg/client/public/locales/en/translation.json
```diff
@@ -233,7 +230,6 @@
         "risk": "Risk",
         "scheduled": "Scheduled",
         "select": "Select",
-        "selected": "Selected",
```
gildub, line 236: The commit doesn't need this change

pkg/client/public/locales/en/translation.json
```diff
@@ -37,9 +37,6 @@
         "save": "Save",
         "saveAndReview": "Save and review",
         "saveAsDraft": "Save as draft",
-        "selectAll": "Select all",
-        "selectNone": "Select none",
-        "selectPage": "Select page",
```
gildub, line 42: The commit doesn't need this change

pkg/client/tsconfig.json
```diff
@@ -29,6 +29,7 @@
     },
     "skipLibCheck": true,
     "strict": true, // Includes alwaysStrict and noImplicitAny
+    "noImplicitAny": false,
```
gildub, line 32: We need to avoid going on that road.

gildub (changes requested):
As we've discussed it would be preferable to use the Express API approach.

pkg/client/tsconfig.json
ibolton336: This is the approach that forklift-ui takes. Need this to manipulate the global window object.

ibolton336:
> As we've discussed it would be preferable to use the Express API approach.

This approach is the same as the forklift-ui approach. We discussed it would be preferable to use the same approach across the apps. Adding in an asynchronous call when we are bootstrapping the app could cause some unforeseen issues. This PR just implements the proven template variable approach that Mike recommended. This way we can keep our strict typing for the env var constant.

gildub:
@ibolton336, the way Forklift/MTV addressed its needs is definitively not a standard across the apps.
I would rather have a better long term solution.
Let's discuss it.

pkg/client/config/webpack.common.js
```diff
@@ -170,9 +172,26 @@ module.exports = (env) => {
...
+        ...(env !== 'production' 
+          ? {
+              // In dev mode, populate window._env at build time
+              filename: 'index.html',
+              template: path.resolve(__dirname, "../public/index.html.ejs"),
+              favicon: path.resolve(__dirname, "../public/favicon.ico"),
+              templateParameters: {
+                _env: helpers.getEncodedEnv(),
+              },
+            }
```
gildub, line 189: Why not having that block in webpack.dev.js ?

pkg/client/config/webpack.common.js
```diff
@@ -170,9 +172,26 @@ module.exports = (env) => {
...
+          : {
+              // In real prod mode, populate window._env at run time with express
+              filename: 'index.html.ejs',
+              template: `!!raw-loader!${path.resolve(__dirname, '../public/index.html.ejs')}`,
+            }),
```
gildub, line 194: And that one in webpack.prod.js ?

gildub (commented):
See inline

gildub (approved):
Let's move forward and adjust later if needed. Thanks

PR #65, opened 2022-03-30
Wizard steps and validation for mode and scope steps

gildub:


pkg/client/src/app/pages/applications/analysis-wizard/analysis-wizard.tsx
```diff
@@ -80,7 +84,6 @@ export const AnalysisWizard: React.FunctionComponent<IAnalysisWizard> = ({
 
   const [isInitTasks, setInitTasks] = React.useState(false);
   const [createdTasks, setCreatedTasks] = React.useState<Array<Task>>([]);
-  const isMutating = useIsMutating();
```
ibolton336, line 83: Need to integrate this into the upload binary step so that the next button is disabled when uploading a binary.

ibolton336 (changes requested):

pkg/client/src/app/pages/applications/analysis-wizard/analysis-wizard.tsx
```diff
@@ -277,85 +283,52 @@ export const AnalysisWizard: React.FunctionComponent<IAnalysisWizard> = ({
             />
           ),
           canJumpTo: stepIdReached >= stepId.AnalysisMode,
+          enableNext: isModeValid(),
```
ibolton336, line 286:
```suggestion
          enableNext: isModeValid() || isMutating,
```

pkg/client/src/app/pages/applications/analysis-wizard/analysis-wizard.tsx
```diff
@@ -97,17 +100,57 @@ export const AnalysisWizard: React.FunctionComponent<IAnalysisWizard> = ({
...
+      (application) =>
+        application.repository &&
+        application.repository.url !== "" &&
+        application.identities &&
+        application.identities.length > 0 &&
+        hasIdentity(application, "maven")
+    );
 
   const { handleSubmit, watch, reset } = methods;
+  console.log(watch());
```
ibolton336, line 153:
```suggestion
  React.useEffect(() => {
    const subscription = watch(() => {});
    return () => {
      subscription.unsubscribe();
    };
  }, [watch]);```

**Review by ibolton336 — CHANGES_REQUESTED** [review]

File: pkg/client/src/app/pages/applications/analysis-wizard/analysis-wizard.tsx (hunk 3 above)
Comment (ibolton336) [reply pkg/client/src/app/pages/applications/analysis-wizard/analysis-wizard.tsx:153]: We should use the recommended usage of watch shown here: https://react-hook-form.com/api/useform/watch/

File: pkg/client/src/app/pages/applications/analysis-wizard/analysis-wizard.tsx [hunk 4]
```diff
@@ -371,7 +344,6 @@ export const AnalysisWizard: React.FunctionComponent<IAnalysisWizard> = ({
         navAriaLabel={`${title} steps`}
         mainAriaLabel={`${title} content`}
         steps={steps}
```
ibolton336, line 346: Seeing a type error on steps here.

pkg/client/src/app/pages/applications/analysis-wizard/analysis-wizard.tsx
gildub: That's a good point.

pkg/client/src/app/pages/applications/analysis-wizard/analysis-wizard.tsx
gildub: Actually there is no need for watch(), we use it for it's side effects.
If you remember not using it freezed the radio buttons on the `SetScope` page.
Unfortunately setting a subscription doesn't work either as this is creating a circular reference error.
There seems to be limitation of this library with advanced type, which we might be hitting when using string arrays, not sure, but this is really annoying, see  https://github.com/react-hook-form/react-hook-form/issues/4670 and https://github.com/react-hook-form/react-hook-form/issues/4704 discussions about it.
In the meantime we decide what to do I'll revert to console.log(watch).

ibolton336: Yeah I am not sure what to do about the circular reference error either. Looks like the creator/maintainer recommends using simple string types for all form value fields until submit time? Not sure if that is even feasible with our application. 

Ok if you do not need it we can remove it entirely. And if we do need it, let's use some better syntax than console.log is all I ask. Like the docs suggest:

const watchAllFields = watch(); // when pass nothing as argument, you are watching everything

pkg/client/src/app/pages/applications/analysis-wizard/analysis-wizard.tsx
gildub: I can't see that error in my env.

pkg/client/src/app/pages/applications/analysis-wizard/analysis-wizard.tsx
gildub: We need it until we fix the related issue blocking the SetScope component.
I'm fine with a dummy watchAllFields

gildub:
@ibolton336, the first step is only valid if binary or source code is defined on involved application(s) or also a binary upload in the single application case. I'm adding the warning message next.

pkg/client/src/app/pages/applications/analysis-wizard/analysis-wizard.tsx
```diff
@@ -110,18 +112,7 @@ export const AnalysisWizard: React.FunctionComponent<IAnalysisWizard> = ({
...
-    fetchIdentities();
-  }, [fetchIdentities]);
-
-  const hasIdentity = (application: Application, kind: string): boolean =>
-    !!application.identities?.some((appIdentity) =>
-      identities?.find(
-        (identity) => appIdentity.id === identity.id && identity.kind === kind
-      )
-    );
+  React.useEffect(() => fetchIdentities(), [fetchIdentities]);
```
ibolton336, line 115: What do we gain from having a use effect here? One of the perks of using react-query is that it eliminates the need for useEffect. This might be a bit of an anti-pattern. If we want to have a reusable function for filtering identities, I would suggest adding a utility function rather than a hook here.

ibolton336 (changes requested):

pkg/client/src/app/shared/hooks/useHasIdentity.ts
```diff
@@ -0,0 +1,20 @@
+import { Application } from "@app/api/models";
+import { useFetchIdentities } from "@app/shared/hooks/useFetchIdentities";
+
+interface IUseHasIdentity {
+  hasIdentity: (application: Application, kind: string) => boolean;
+  fetchIdentities: () => void;
+}
+
+export const useHasIdentity = (): IUseHasIdentity => {
+  const { identities, fetchIdentities } = useFetchIdentities();
```
ibolton336, line 10: I think you could pass in the identities to this hook as a param rather than calling this here.

pkg/client/src/app/pages/applications/analysis-wizard/analysis-wizard.tsx
```diff
@@ -77,10 +82,12 @@ export const AnalysisWizard: React.FunctionComponent<IAnalysisWizard> = ({
 }: IAnalysisWizard) => {
   const title = "Application analysis";
   const dispatch = useDispatch();
+  const isMutating = useIsMutating();
 
   const [isInitTasks, setInitTasks] = React.useState(false);
   const [createdTasks, setCreatedTasks] = React.useState<Array<Task>>([]);
-  const isMutating = useIsMutating();
+
+  const { hasIdentity, fetchIdentities } = useHasIdentity();
```
ibolton336, line 90: Could pass in the identities here & create a reusable filter hook that is extensible to other types if you wanted. Otherwise, I think the useHasIdentity may be better off as a one off utility function since it isn't really reused.

ibolton336 (changes requested):

gildub:
This PR has been replaced with https://github.com/konveyor/tackle2-ui/pull/67

PR #67, opened 2022-04-01
Applications analysis table, wizard: Taskgroup queries, warning, fixes and more

gildub:
- Replaces `/hub/tasks` queries with `/hub/taskgroups`
- Replaces legacy fetch hooks with react queries hooks
- Taskgroup deletion when wizard is canceled     
- Analysis status now get proper refresh
- Wizard steps chaining
- SetMode will display a warning if involved application don't fit the analysis requirements (source/binary/credentials)
- Targets: new jakarta-ee, openliberty, and openjdk11targets and standard corporate svg icon black for upstream
- Setscope has first radio button as default
- Adds ReactQueryDevtools
- Fixes "Function component cannot be given refs..." warning in log on wizard SetMode step


pkg/client/src/app/pages/applications/applications-table-analyze/applications-table-analyze.tsx
```diff
@@ -208,7 +190,7 @@ export const ApplicationsTableAnalyze: React.FC = () => {
     }
 
     closeApplicationModal();
-    refreshTable();
+    // refreshTable();
```
ibolton336, line 193: Do we need this?

pkg/client/src/app/queries/tasks.ts
```diff
@@ -14,24 +14,31 @@ export const useFetchTasks = (
...
+              const others = uniqLatestTasks.filter((t) => t.id !== aTask.id);
+              uniqLatestTasks = [...others, task];
+            } else uniqLatestTasks.push(task);
+          });
+          setTasks(uniqLatestTasks);
+        })
+        .catch((error) => {
+          console.log("error, ", error);
+        }),
+    { refetchInterval: 1000 * 60 }
```
ibolton336, line 41: Maybe we want to fetch more often and push it out if we a see a performance hit?

ibolton336:
After running through an analysis wizard, I am not seeing my applications list update. I had a failed analysis & had to navigate away and come back for the status to show up.

ibolton336:
I think we would need to update the analysis-wizard onSubmit to use a mutate tasks query. Then we could invalidate the applications query like we currently do for proxies.

ibolton336 (changes requested):

pkg/client/src/app/queries/tasks.ts
ibolton336: Can also disable query if a condition is happening like a wizard modal being open etc. https://react-query.tanstack.com/guides/disabling-queries

pkg/client/src/app/queries/applications.ts
```diff
@@ -0,0 +1,57 @@
...
+import { deleteApplication, getApplications } from "@app/api/rest";
+
+export interface IFetchState {
+  applications: Application[];
+  isFetching: boolean;
+  fetchError: any;
+  refetch: () => void;
+}
+
+export const useFetchApplications = (
```
ibolton336, line 14: Nice job porting this over to applications! I think this will really help us.

pkg/client/src/app/pages/applications/applications-table-analyze/applications-table-analyze.tsx
gildub: Taken care of

ibolton336:
A few questions: 
- Have we implemented the warning within the wizard? 
- Seeing the analyze button disabled after attempting a failed analysis. Seeing some weird state after a bulk analysis & not sure what the outcome of the analysis was. Only one analysis shows failed and can't click the analyze button again. 
Screen Shot 2022-04-05 at 8 21 46 AM

ibolton336:
When starting a bulk analysis, I see the tasks hanging at scheduled with no info updating after several minutes 

Screen Shot 2022-04-05 at 8 32 02 PM

ibolton336:
Seeing this warning in the console:
Screen Shot 2022-04-05 at 9 24 47 PM

ibolton336:
Just jotting some notes down here with some questions. Maybe some are candidates for future issues. 
1) There is no default value on the scope page & next is enabled 
2) Seems like the api is getting pegged with tasks and application requests every second. Can we control this polling logic a little tighter? 
3) It is unclear why the analyze button is disabled. 
4) No information when completing the wizard without valid tasks. 
5) Sometimes when I re-open the wizard, my entire browser crashes and I can't recover the session.

gildub:
@ibolton336, since this PR is now way ahead of PR#65, I'm addressing some of the issues in this one.

I'm updating the description of this PR with additions/fixes

ibolton336 (approved):
VISACK

PR #69, opened 2022-04-06
Wizard: SetTargets to remember selection

gildub:
Makes sure targets selected are properly active when going back and forth between wizard steps.


ibolton336 (approved):
VISACK

PR #73, opened 2022-04-07
Remove legacy headers

ibolton336:
- Removes HAL headers for imports and dependencies.


gildub (changes requested):
Since HAL is not used entirely let's remove all reference to HAL

PR #74, opened 2022-04-07
Wizard: Single app requires `taskgroups` to upload file

gildub:
Fixes the binary file upload for single application analysis case.

Recent change using taskgroups API needs to upload file with `taskgroups` and not through `tasks` API.

### DIFF
--- a/pkg/client/src/app/queries/applications.ts
+++ b/pkg/client/src/app/queries/applications.ts
@@ -15,17 +15,14 @@ export const useFetchApplications = (
   defaultIsFetching: boolean = false
 ): IFetchState => {
   const [applications, setApplications] = useState<Application[]>([]);
-  const { isLoading, error, refetch } = useQuery(
-    "applications",
-    () =>
-      getApplications()
-        .then(({ data }) => {
-          setApplications(data);
-        })
-        .catch((error) => {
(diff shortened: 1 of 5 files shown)


pkg/client/src/app/queries/applications.ts
```diff
@@ -15,14 +15,17 @@ export const useFetchApplications = (
...
-      })
-      .catch((error) => {
-        console.log("error, ", error);
-      })
+  const { isLoading, error, refetch } = useQuery(
+    "applications",
+    () =>
+      getApplications()
+        .then(({ data }) => {
+          setApplications(data);
```
ibolton336, line 23: Take a look at using the queryClient here instead of react state.  

[query client link ](https://react-query.tanstack.com/reference/QueryClientProvider)const queryClient = useQueryClient()

ibolton336 (changes requested):

pkg/client/src/app/queries/applications.ts
gildub: This is actually a mistake. That change has nothing to do with this patch.
Although that's something to discuss separately.

ibolton336 (approved):

PR #76, opened 2022-04-07
Tag color fix for app expanded view

ibolton336:
- Show correct tag colors in app expanded view.
https://issues.redhat.com/browse/TACKLE-420

### DIFF
--- a/pkg/client/src/app/pages/applications/components/application-tags/application-tags.tsx
+++ b/pkg/client/src/app/pages/applications/components/application-tags/application-tags.tsx
@@ -44,7 +44,7 @@ export const ApplicationTags: React.FC<ApplicationTagsProps> = ({
           const newTagTypes: Map<number, TagType> = new Map();
           const newTags: Map<number, Tag[]> = new Map();
 
-          const validResponses = tags.reduce((prev, current) => {
+          const tagValidResponses = tags.reduce((prev, current) => {
             if (current) {
               return [...prev, current.data];
             } else {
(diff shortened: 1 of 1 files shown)


pkg/client/src/app/pages/applications/components/application-tags/application-tags.tsx
```diff
@@ -44,38 +44,58 @@ export const ApplicationTags: React.FC<ApplicationTagsProps> = ({
...
-
-              // Tags
-              newTags.set(tagType.id!, [
-                ...(newTags.get(tagType.id!) || []),
-                e,
-              ]);
-            }
+          Promise.all(
+            tagValidResponses.map((tag) =>
+              getTagTypeById(tag?.tagType?.id || 0)
```
gildub, line 62: Why not use React queries ?

gildub (commented):
It seems tags are still showing in gray only color.

pkg/client/src/app/pages/applications/components/application-tags/application-tags.tsx
ibolton336: This is a bug fix for the existing code. Sure we can move to queries. But this PR aims to fix the bug with the current code.

ibolton336:
> It seems tags are still showing in gray only color.
Maybe try to pull down the latest changes - Where are you seeing the tags with gray only color?
Screen Shot 2022-04-08 at 8 23 04 AM

pkg/client/src/app/pages/applications/components/application-tags/application-tags.tsx
gildub: Ah yes of course it's already using legacy query, never mind.

gildub (approved):
Latest pull worked for me.

PR #77, opened 2022-04-07
Add create/update user to identity request.

ibolton336:
https://issues.redhat.com/browse/TACKLE-424

Pending backend changes to allow user to be retrieved via api on identities get requests..

Also, this will need an env var check for isAuthenticated whenever that merges.


gildub:
Could we please refer backend PR dependency in the description ?

jortel (changes requested):
This is obsoleted by: https://github.com/konveyor/tackle2-ui/pull/273 and should be closed.

PR #78, opened 2022-04-08
Add inline form validation for duplicates & convert identities to queries

ibolton336:
https://issues.redhat.com/browse/TACKLE-427 

- Adds inline form validation
- converts identities to queries for cache usage & general flow enhancements


gildub (approved):
Looks good.

PR #79, opened 2022-04-08
Analysis: custom rules uploadfile with cleaner queries

gildub:
- Use React query for custom rules files upload
- Consolidate analysis queries and prepare for `queryClient.invalidateQuery()` and `queryClient.setQueryData()` for handling data synchronization
- Also benefit from type inference

### DIFF
--- a/pkg/client/src/app/pages/applications/analysis-wizard/analysis-wizard.tsx
+++ b/pkg/client/src/app/pages/applications/analysis-wizard/analysis-wizard.tsx
@@ -1,18 +1,8 @@
 import * as React from "react";
+import { useIsMutating } from "react-query";
 import { FieldValues, FormProvider, useForm } from "react-hook-form";
-import {
-  Application,
-  Identity,
-  TaskData,
-  Taskgroup,
-  TaskgroupTask,
-} from "@app/api/models";
-import "./wizard.css";
 import { yupResolver } from "@hookform/resolvers/yup";
 import * as yup from "yup";
-import { useState } from "react";
-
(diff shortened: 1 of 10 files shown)


pkg/client/src/app/pages/applications/analysis-wizard/analysis-wizard.tsx
```diff
@@ -265,18 +270,17 @@ export const AnalysisWizard: React.FunctionComponent<IAnalysisWizard> = ({
 
     if (createdTaskgroup) {
       const taskgroup = setTaskgroup(createdTaskgroup, data);
-      Promise.all(
```
ibolton336, line 268: How were you able to get rid of the Promise.all here? Wondering because I was seeing some behavior issues resulting from promises coming back asynchronously instead of all promises resolving at once. I do not have any rules files to test this - just a note from my experience with the tag fetching in my recent PR.

ibolton336 (approved):

pkg/client/src/app/pages/applications/analysis-wizard/analysis-wizard.tsx
gildub: I replaced it with react query mutation, this is easier now that we're using Taskgroups, each rules file is uploaded only once for the group.

For testing, there are xml files available in "pkg/client/__mocks__/" folder 

I also realized the custom rules file upload needs to be move to the UploadFile component (triggered in the SetOptions step). 
But that can be a separate PR. Basically it would be better to upload the valid files as the user handles them.

And the other

PR #84, opened 2022-04-11
Convert applications to react-query and fix url field clear bug

ibolton336:
- When attempting to clear the URL field, the user would get a silent failure. This fixes this and fixes the refetch delay for app loading after mutation.


gildub (approved):
LGTM

PR #86, opened 2022-04-11
Fix application property overwriting when adding credentials 

ibolton336:
- Spreads app properties to the payload when attaching a credential to an application
- Fixes credential properties overwritten in app edit form


gildub (approved):
LGTM

PR #90, opened 2022-04-12
Fix delete application not showing up in table

ibolton336:
- Adds delete mutation for apps & migrates assessment view to use query fetch


gildub (approved):
LGTM

PR #100, opened 2022-04-13
Add shared filter for applications tables

ibolton336:
Adds shared filter logic across analysis/assessments. Also fixes broken assessments application table filter.


gildub (approved):
Good one sorted out !

PR #104, opened 2022-04-14
Add toggle for pw visibility

ibolton336:


gildub (commented):
That works well for credential creation but not for updating.

ibolton336:
@gildub Just pushed a change to not show initial placeholder values in the UI. We don't have acces to the user/pw fields of existing credentials at the moment.

gildub (approved):
LGTM

PR #105, opened 2022-04-14
Fix identities sort and filter

ibolton336:
[Credentials : Sorting on Type and Name field does not work](https://issues.redhat.com/browse/TACKLE-431) 
https://issues.redhat.com/browse/TACKLE-429


jortel (approved):
lgtm

PR #113, opened 2022-04-19
Fix tab loading bug and color filter bug

ibolton336:
https://issues.redhat.com/browse/TACKLE-441 

- Shows correct controls tab when reloading
- Fixes color filter issue in tags table


jortel (approved):
lgtm

PR #132, opened 2022-04-25
Update review datamodel and fix sort

ibolton336:
https://issues.redhat.com/browse/TACKLE-452
https://issues.redhat.com/browse/TACKLE-453


gildub:
The sort issue is still not working.
Why remove tests files ? Is it possible to amend them ?
I think we should split this PR in two to address the related issues separately.

gildub:
@ibolton336, I understand, let's keep them together. 
Regarding tests we decided to not add unit tests but to keep existing one when possible. 
Is that an issue here ?

gildub:
@ibolton336, sort works but the review comments are not displayed. 
Also would it be possible to keep some part of the tests files ?

ibolton336:
> @ibolton336, sort works but the review comments are not displayed. Also would it be possible to keep some part of the tests files ?

I tried to keep the tests, but the tests removed are testing the legacy sort/filter functions that have been removed. Maybe we can add in some tests for our lib-ui component filter logic to cover these cases.

gildub (approved):
Working just fine now.
I understand re legacy tests.

PR #142, opened 2022-04-28
Edit identity bug

ibolton336:
Supports back-end changes to identity objects. Now shows username & password when editing credentials


jortel (approved):
lgtm

PR #147, opened 2022-04-29
Download import summaries csv

gildub:


pkg/client/src/app/pages/applications/manage-imports/manage-imports.tsx
```diff
@@ -351,9 +351,9 @@ export const ManageImports: React.FC = () => {
                     <KebabDropdown
                       dropdownItems={[
                         <DropdownItem
-                          key="download-csv-sample"
+                          key="download-csv-template"
                           component={
-                            <a href="/sample_application_import.csv" download>
```
aufi, line 356: It looks it is needed fix a route or an existence of the sample CSV file. "hub/importsummaries/download" path downloads previously uploaded CSV Import file which could be different to the sample one

aufi (commented):
Not sure what is the correct path of the sample file ```public/src``` or ```public/``` (as originally and updated in https://github.com/konveyor/tackle2-ui/pull/148)

gildub:
@aufi , the template file path is `pkg/client/public/` then the bundler (webpack) will copy it to the `dist/` directory.

ibolton336 (approved):

aufi:
Thanks for info :+1:

PR #150, opened 2022-04-29
Fix mvn creds not updating

ibolton336:
https://issues.redhat.com/browse/TACKLE-514

### DIFF
--- a/pkg/client/src/app/api/rest.tsx
+++ b/pkg/client/src/app/api/rest.tsx
@@ -255,6 +255,18 @@ export const updateApplication = (
   return APIClient.put(`${APPLICATIONS}/${obj.id}`, obj);
 };
 
+export const updateAllApplications = (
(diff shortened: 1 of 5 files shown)


gildub (changes requested):
In the application expanded area the credentials is not showing up when maven credentials are added to an application.

ibolton336:
> In the application expanded area the credentials is not showing up when maven credentials are added to an application.

Sounds like a separate issue.

gildub:
Oh I might be wrong meanwhile the related issue says "Maven creds not displayed upon expanding an application"

ibolton336:
> Oh I might be wrong meanwhile the related issue (Tackle-514) says "Maven creds not displayed upon expanding an application"

The screenshot seems to suggest it is with the "expanded" edit credentials modal.

gildub (approved):
LGTM

PR #153, opened 2022-05-02
Fix imports

gildub:
- Add refresh interval for `importSummaries`, to avoid reloading page for latest
- Displays import report (the page was freezing) 
- Fix delete warning message

https://issues.redhat.com/browse/TACKLE-530
https://issues.redhat.com/browse/TACKLE-515

### DIFF
--- a/pkg/client/src/app/pages/applications/manage-imports/manage-imports.tsx
+++ b/pkg/client/src/app/pages/applications/manage-imports/manage-imports.tsx
@@ -88,10 +88,11 @@ export const ManageImports: React.FC = () => {
     dispatch(confirmDialogActions.closeDialog());
   };
 
-  const onDeleteImportSummaryError = (error: AxiosError) => {
+  const onDeleteImportSummaryError = (error: Error | null) => {
     dispatch(confirmDialogActions.closeDialog());
-    dispatch(alertActions.addDanger(getAxiosErrorMessage(error)));
+    if (error) dispatch(alertActions.addDanger(error.message));
   };
+
   const { mutate: deleteImportSummary } = useDeleteImportSummaryMutation(
     onDeleteImportSummarySuccess,
     onDeleteImportSummaryError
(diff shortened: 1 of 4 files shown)


pkg/client/src/app/queries/imports.ts
```diff
@@ -42,16 +41,16 @@ export const useFetchImportSummaries = () => {
...
-      .then(({ data }) => {
-        setImportSummaries(data);
-        return data;
-      })
-      .catch((error) => {
-        console.log("error, ", error);
-        return error;
-      })
+  const { isLoading, error, refetch } = useQuery(
+    "importsummaries",
```
ibolton336, line 45: Looking good! One small thing -  We should define this as a constant and export it as we do in the other queries. Thanks!

ibolton336 (changes requested):

pkg/client/src/app/queries/imports.ts
gildub: I'm more than happy to do that when the key is reused which is not the case anymore.

pkg/client/src/app/pages/applications/manage-imports/manage-imports.tsx
```diff
@@ -269,6 +267,7 @@ export const ManageImports: React.FC = () => {
   const deleteRow = (row: ApplicationImportSummary) => {
     dispatch(
       confirmDialogActions.openDialog({
+        // t("terms.tag(s)")
```
ibolton336, line 270: Are we using this?

pkg/client/src/app/queries/imports.ts
ibolton336: If we ever need to reference the imports query, it would be nice to have it available. Seems silly to remove.

ibolton336 (dismissed):

pkg/client/src/app/pages/applications/manage-imports/manage-imports.tsx
gildub: Yest this is work around otherwise i18n is going to hit us with translations.json files.

ibolton336 (commented):
LGTM - thanks Gilles

ibolton336 (approved):

PR #167, opened 2022-05-06
Allow admin view if no auth

ibolton336:
- Follow up to single user changes: 
- Allows admin view when auth is not enabled


gildub (approved):
LGTM

PR #179, opened 2022-05-09
Bug 504: Add tag filter for applications

ibolton336:
https://issues.redhat.com/browse/TACKLE-504


gildub (approved):
LGTM, just a small issue is the select drop-down stays active after a selection, but that's no blocker and can be addressed after.

PR #187, opened 2022-05-11
Update padding to display Report adoption correctly

aufi:
There was reported overlapping of application circle and adoption candidate chart description. Updating paddings to make title text readable.

Note, this is my first commit to the Tackle2 UI, I'm happy for every possible feedback. (the change might be naive, but looks to me to solving the reported issue). Sample screenshots attached in comments.

https://issues.redhat.com/browse/TACKLE-529

PR #189, opened 2022-05-11
Fix adoption graph title spacing

aufi:
Follow-up for https://github.com/konveyor/tackle2-ui/pull/187 fixing space under title after got feedback from QE.


aufi:
Example:
image

PR #207, opened 2022-05-12
Fix CSV template dependencies format

aufi:
CSV application import can include rows with RecordType=2 dependencies
which should not have trailing empty columns. (this should fit to
Tackle1)


aufi:
Related to https://github.com/konveyor/tackle2-hub/pull/88, hopefully this is the final version :-)

jortel (approved):
LGTM

PR #217, opened 2022-05-13
Poll tasks for clean completion

ibolton336:
- Adds a new hook to handle POST/FETCH for clean functionality
- Adds a confirmation dialog modal to confirm the clean action
image

### DIFF
--- a/pkg/client/src/app/pages/repositories/Mvn.tsx
+++ b/pkg/client/src/app/pages/repositories/Mvn.tsx
@@ -2,6 +2,7 @@ import * as React from "react";
 import {
   Alert,
   Button,
+  ButtonVariant,
   Card,
   CardBody,
   Form,
@@ -26,9 +27,15 @@ import {
   useCleanRepositoryMutation,
   useFetchVolumes,
 } from "@app/queries/volumes";
+import { useFetchTasks } from "@app/queries/tasks";
+import { useDispatch } from "react-redux";
+import { confirmDialogActions } from "@app/store/confirmDialog";
 
 export const RepositoriesMvn: React.FunctionComponent = () => {
   const { t } = useTranslation();
+  // Redux
+  const dispatch = useDispatch();
+
   const [forcedSettingError, setForcedSettingError] =
     React.useState<AxiosError>();
   const [insecureSettingError, setInsecureSettingError] =
@@ -107,9 +114,10 @@ export const RepositoriesMvn: React.FunctionComponent = () => {
   }, [refreshMvnForcedSetting]);
 
   const { volumes, refetch } = useFetchVolumes();
+  const { tasks } = useFetchTasks();
   const [storageValue, setStorageValue] = useState<string>();
   const [currCleanId, setCurrCleanId] = useState<number>(0);
-  const [isCleanDisabled, setIsCleanDisabled] = useState<boolean>(true);
+  const [isCleanDisabled, setIsCleanDisabled] = useState<boolean>(false);
 
   useEffect(() => {
     const thisVol = volumes.find((vol) => vol.name === "m2");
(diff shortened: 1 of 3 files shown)


ibolton336:
https://issues.redhat.com/browse/TACKLE-588

pkg/client/src/app/pages/repositories/Mvn.tsx
```diff
@@ -107,9 +114,10 @@ export const RepositoriesMvn: React.FunctionComponent = () => {
   }, [refreshMvnForcedSetting]);
 
   const { volumes, refetch } = useFetchVolumes();
+  const { tasks } = useFetchTasks();
   const [storageValue, setStorageValue] = useState<string>();
   const [currCleanId, setCurrCleanId] = useState<number>(0);
-  const [isCleanDisabled, setIsCleanDisabled] = useState<boolean>(true);
+  const [isCleanDisabled, setIsCleanDisabled] = useState<boolean>(false);
```
mturley, line 120: If the only time `isCleanDisabled` is true is while the `isLoading` property of the mutation is also true, we can probably consider this redundant and remove it and let `isLoading` be the only thing that disables the clean button right?

Edit: ah, I see now that `isLoading` isn't what we want here because that's the polling state, not the entire process of waiting for the clean to finish. Hmm.

pkg/client/src/app/pages/repositories/Mvn.tsx
```diff
@@ -164,8 +188,7 @@ export const RepositoriesMvn: React.FunctionComponent = () => {
                   isInline
```
mturley, line 188: While we're in here can we remove this `isInline` prop on the clear button? It'll get a natural padding that prevents it from being right up against the text field like it is now.

pkg/client/src/app/queries/volumes.ts
```diff
@@ -59,3 +40,54 @@ export const useFetchVolumes = (): IVolumeFetchState => {
     refetch,
   };
 };
+
+export const useCleanRepositoryMutation = ({ onSuccess, onError }) => {
+  const [processId, setProcessId] = useState(null);
+  const [stop, setStop] = useState(false);
```
mturley, line 46: ~I think we could use a more descriptive name than `stop` here like `isPollingDisabled` maybe? Also it seems like it should be defaulting to true?~

~Or maybe it would be more readable if we inverted it and called it something like `isPollingTaskStatus`, started it as false and set it true when the task starts and false when it's done.~

Actually, I think maybe you don't need this `stop` state at all. See below.

pkg/client/src/app/queries/volumes.ts
```diff
@@ -59,3 +40,54 @@ export const useFetchVolumes = (): IVolumeFetchState => {
...
+
+  // Mutation to kick off the clean task
+  const { mutate } = useMutation(cleanRepository, {
+    onMutate: () => {
```
mturley, line 50: Why `setStop(false)` here and not when `setProcessId` is called in `onSuccess`?

pkg/client/src/app/queries/volumes.ts
```diff
@@ -59,3 +40,54 @@ export const useFetchVolumes = (): IVolumeFetchState => {
...
+        onError();
+      },
+      enabled: processId != null,
+      refetchInterval: stop ? false : 5000,
+      refetchIntervalInBackground: true,
+      refetchOnWindowFocus: false,
+    }
+  );
+
+  return { mutate, data, isLoading };
```
mturley, line 92: Thinking through the state and return values here, I wonder if there's a cleaner way. Right now:
* useCleanRepositoryMutation keeps track of the process ID and whether it should be polling, and returns only the loading state of the polling request (which will be true momentarily every 5 seconds and false otherwise)
* The component above has state for `isCleanDisabled` that should be true while the clean is "running" (it hasn't seen the task succeed/fail yet) and is controlled via the `onSuccess` and `onError` callbacks in here.

Looking at how `useMutation` itself handles status, I wonder if instead you can have this hook own the state of whether the clean process is done (so you can use that as the source of truth for disabling the button instead of having more state up above). I think it could all be driven just from the `processId` state.

Consider this:
* We should only be polling the task when `processId` is not null, so that works for both `enabled` and `refetchInterval` in the useQuery and we don't need the `stop` state.
* We should only disable the "clear repository" button when we're kicking off or polling the task, so we could grab the `isLoading` from the `useMutation` call and call it `isProcessIdLoading`, and this whole hook could return `isCleaning: isProcessIdLoading || !!processId` from here and use that to disable the button instead of having the `isCleanDisabled` state in the component above.

You could still use your `onSuccess` and `onError` to cause the `refetch();` to happen when it's done, but I think `processId` is the only state you need for this whole deal.

mturley (changes requested):
A couple thoughts on how to simplify state here, and one nit about style.

pkg/client/src/app/queries/volumes.ts
mturley: Actually, holy crap, you might not need any `useState` AT ALL here: the process id is already in the `data` returned by `useMutation`, and you can reset the mutation by calling the `reset` function it returns. So instead of `processId` being a `useState`, you could have `const processId = data?.id || null;`, which will be null until the mutation resolves and go back to being null when you `reset` the mutation. All the state you need is already part of `useMutation`!!

pkg/client/src/app/queries/volumes.ts
```diff
@@ -59,3 +40,49 @@ export const useFetchVolumes = (): IVolumeFetchState => {
...
+        reset();
+        onError();
+      }
+    },
+    onError: (error) => {
+      console.error(error);
+      reset();
+      onError();
+    },
+    enabled: processId != null,
```
mturley, line 77: Nit: use `!==` instead of `!=`

mturley (approved):
Other than the `!==` operator nitpick, LGTM!

gildub:
@ibolton336, do you mind fixing last comment from @mturley ?

PR #219, opened 2022-05-13
Fix application import errors list

aufi:
Application import errors list showed all imports from all ImportSummaries and both sucessful&errored.

Updating query to ask API only for Imports for given ImportSummary and valid field is false (as API is expecting it).

Fixes https://issues.redhat.com/browse/TACKLE-589


aufi:
Example before this change:
image

With change in this PR:
image

gildub (approved):
LGTM

PR #221, opened 2022-05-13
TACKLE-537: Fix validation on encrypted Maven settings field so we can save edits, fix SCM private key field to behave consistently

mturley:
Resolves https://issues.redhat.com/projects/TACKLE/issues/TACKLE-537.

The bug here was caused by the Yup validation of the Maven settings field being unaware that its value is encrypted when prefilling for editing. The field was failing validation because its encrypted form cannot be parsed as XML, and that validation error was not appearing on the screen because the form is configured to only update the validation errors when a field changes (not when its initial value is set -- this is done so that "this field is required" doesn't appear on all required fields when a form first renders).

The solution is to short circuit the validation if the field's value has not changed relative to the credential being edited. If it was valid before, we assume it is still valid. If the field has been changed, we assume the user is replacing it with cleartext XML, so it validates as normal.

I noticed a few related quirks while fixing this that I fixed while I'm in here:
* For both file upload fields in this form (this one and the SCM private key field) the Clear buttons were not actually clearing the field when in edit mode, but resetting to the saved value. I fixed them so the Clear button actually fully clears the field.
* The SCM private key field was not showing [Encrypted] in its body, so if you are editing a credential with this field, it will show you the encrypted key (confusing and inconsistent). I changed it so it behaves the same as the Maven settings file (shows [Encrypted] if there is a value in the field).
* The SCM private key fields was never triggering validation errors, because it was not using the `onChange` event (it was directly calling `setValue` which circumvents the event). react-hook-form doesn't allow the form to submit because the field is invalid, but it doesn't display the reason to the user because it doesn't realize the value was changed by the user. I switched this field to use `onChange` like the Maven settings file field does, so now it properly shows the validation error when you change it. (it only validates that it is required, so this only matters when it is cleared).

### DIFF
--- a/pkg/client/src/app/pages/identities/components/identity-form.tsx
+++ b/pkg/client/src/app/pages/identities/components/identity-form.tsx
@@ -525,7 +525,8 @@ export const IdentityForm: React.FC<IdentityFormProps> = ({
                 label={
                   "Upload your [SCM Private Key] file or paste its contents below."
                 }
-                helperTextInvalid="You should select a private key file."
+                validated={getValidatedFromError(errors.key)}
+                helperTextInvalid={!isLoading && errors?.key?.message}
                 isRequired
                 //TODO: PKI crypto validation
                 // validated={isFileRejected ? "error" : "default"}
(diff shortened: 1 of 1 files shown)


ibolton336 (approved):
Everything looks great - thanks for going the extra mile here to clean things up.

PR #222, opened 2022-05-13
Fix applications filter

ibolton336:
- Moves select/expand state into shared applications filter. This allows selected/expanded state to respect filter/sort logic


mturley (approved):
LGTM

PR #233, opened 2022-05-16
TACKLE-538: Only allow toggling password visibility if the password is not encrypted (contains a fresh user-entered value)

mturley:
Resolves https://issues.redhat.com/browse/TACKLE-538


ibolton336 (approved):
LGTM

PR #235, opened 2022-05-17
TACKLE-611: Cascade assessment delete when app is deleted

ibolton336:
https://issues.redhat.com/projects/TACKLE/issues/TACKLE-611

### DIFF
--- a/pkg/client/src/app/queries/applications.ts
+++ b/pkg/client/src/app/queries/applications.ts
@@ -6,7 +6,7 @@ import {
   useQueryClient,
 } from "react-query";
 
-import { Application } from "@app/api/models";
+import { Application, Assessment } from "@app/api/models";
 import {
   createApplication,
   deleteApplication,
@@ -16,6 +16,7 @@ import {
 } from "@app/api/rest";
 import { AxiosError } from "axios";
 import { reviewsQueryKey } from "./reviews";
+import { useDeleteAssessmentMutation } from "./assessments";
 
 export interface IApplicationMutateState {
   mutate: any;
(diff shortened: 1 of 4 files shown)


mturley:
What drives the underlying storage of these applications/assessments? They aren't k8s CRs, right? I wonder if this API has a mechanism similar to k8s `ownerReferences`, so the assessments can be created with metadata saying they are owned by the application and have the backend auto-delete them when the app is deleted.

ibolton336:
> What drives the underlying storage of these applications/assessments? They aren't k8s CRs, right? I wonder if this API has a mechanism similar to k8s `ownerReferences`, so the assessments can be created with metadata saying they are owned by the application and have the backend auto-delete them when the app is deleted. That way the UI doesn't have to be concerned with these relationships.

Good question - I think there is a backend golang mechanism to do just that. The issue here is that these requests span across multiple APIs & the solution is not fully fleshed out on the backend yet so Jeff requested that we do this temporarily on the UI side. See comments https://issues.redhat.com/projects/TACKLE/issues/TACKLE-611

mturley:
Got ya, ok. Thanks.

pkg/client/src/app/queries/applications.ts
```diff
@@ -113,13 +109,19 @@ export const useCreateApplicationMutation = (
...
 ) => {
   const queryClient = useQueryClient();
-  return useMutation(deleteApplication, {
-    onSuccess: () => {
+  const { mutate: mutateAssessments } = useDeleteAssessmentMutation();
+  return useMutation(({ id }: { id: number }) => deleteApplication(id), {
+    onSuccess: (_, { id }) => {
       onSuccess();
       queryClient.invalidateQueries(ApplicationsQueryKey);
+      const assessment = getApplicationAssessment(id);
```
mturley, line 121: @ibolton336 When I try to test this, I'm getting an error on this line when I delete an application:
```
TypeError: getApplicationAssessment is not a function
    at Object.onSuccess (applications.ts?33ea:121:26)
    at eval (mutation.js?832d:92:1)
```

ibolton336: Looks like i only fixed it for the assessments page & this function is reused in the analysis page also. Fixing @mturley

gildub (commented):
LGTM.

gildub (approved):

PR #239, opened 2022-05-17
TACKLE-625:  Correct RBAC rules for applications pages

ibolton336:
Adds granularity for read/write scopes across both applications tables

https://issues.redhat.com/browse/TACKLE-625


gildub (commented):
The "Assess" and "Review" button are still not visible when logged as migrator.

gildub (approved):
LGTM

PR #259, opened 2022-05-19
Don't trim trailing whitespace for key upload

ibolton336:


jortel (approved):
LGTM
Trim causes issues with trailing \n in OPENSSH keys.

PR #260, opened 2022-05-23
TACKLE-575: Uses exposed ingress limit to validate uploaded binaries

ibolton336:
image

Closes https://issues.redhat.com/browse/TACKLE-575?filter=12395291


pkg/client/src/app/Constants.ts
```diff
@@ -19,9 +19,11 @@ export interface IEnvVars {
   KEYCLOAK_REALM: string;
   KEYCLOAK_CLIENT_ID: string;
   KEYCLOAK_SERVER_URL: string;
+  UI_INGRESS_PROXY_BODY_SIZE: string;
 }
 export const ENV: IEnvVars = window["_env"] || {};
 export const isAuthRequired = ENV.AUTH_REQUIRED !== "false";
+export const uploadLimit = ENV.UI_INGRESS_PROXY_BODY_SIZE || "500m";
```
amastbau, line 26: I think MB or MiB is preferred.wdty?
@ibolton336

amastbau (changes requested):

pkg/client/src/app/Constants.ts
ibolton336: @amastbau https://kubernetes.github.io/ingress-nginx/user-guide/nginx-configuration/annotations/#custom-max-body-size Going off of the format used from the ingress annotation here. It seems that they use 'm' to indicate megabytes & not sure there is a way to change it.

ibolton336: @jortel @fbladilo

mturley:
@ibolton336 while you're in here can you set the `variant` of the modal to be medium or large? It looks funky taking up the full screen width for a short message.

mturley (approved):

PR #261, opened 2022-05-23
TACKLE-512: Refetch application assessments and reviews properly when copying/discarding (+ upgrade TypeScript)

mturley:
https://issues.redhat.com/browse/TACKLE-512

There were two problems here:
* We weren't triggering a refetch of assessments when completing the discard action in the table.
* Application assessments were being fetched and stored in state local to the table using `useFetchMultiple`, but the bulk copy is an asynchronous task that is later watched elsewhere, so there is no way for the table to know when it is complete and trigger a refetch of this local state.

To solve these, I refactored the assessments fetch logic to use react-query's `useQueries` so that it is stored in the global `queryClient` cache. That way, we can invalidate those queries from wherever we want, which I've done from the notifications container (because that was already watching for this bulk copy to complete. Also, since we rely on the reference to the application review that is present on the application object itself, I also invalidate the applications query there.

Other cleanup includes:
* I made the loading state of the Assessment column a little cleaner: now it will show a spinner instead of just a blank cell, and it will only show that on the initial fetch.
* I removed unused imports from files I touched (habit... maybe shouldn't do this)
* When looking at our other `useQuery` usage I found a number of places where we were duplicating the `data` into separate state, so I simplified those queries to just use the data in the cache directly. Technically this could avoid bugs in the future if we were to manipulate that cache manually, but that's an antipattern anyway. This is just a bit cleaner.

Note: This PR also upgrades TypeScript to ^4.1.0 for tests to pass, because type inference for `useQueries` is incorrect in earlier TS versions. See https://react-query.tanstack.com/typescript

### DIFF
--- a/pkg/client/package-lock.json
+++ b/pkg/client/package-lock.json
@@ -106,7 +106,7 @@
         "ts-jest": "^27.0.3",
         "ts-loader": "^9.2.6",
         "tsconfig-paths-webpack-plugin": "^3.5.2",
-        "typescript": "4.0.3",
+        "typescript": "^4.1.0",
         "url-loader": "^4.1.1",
         "webpack": "^5.67.0",
         "webpack-cli": "^4.9.2",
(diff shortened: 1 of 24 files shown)


ibolton336 (changes requested):
Looking great - just some unit tests failing then should be g2g.

PR #272, opened 2022-05-27
Remove applications cannot be analyzed message for upload binary mode

ibolton336:
- This warning does not apply on  the upload binary screen. The next button being disabled should be enough to notify that they have not uploaded a binary. 
image


mturley (approved):
LGTM

PR #273, opened 2022-05-30
Enable CreateUser field in Identities

aufi:
Replacing User placeholder with the API provided createUser field.

Related to https://github.com/konveyor/tackle2-hub/pull/103 but can be merged before it (the field would be empty until Hub API provides the username).

https://issues.redhat.com/browse/TACKLE-424

### DIFF
--- a/pkg/client/src/app/pages/identities/identities.tsx
+++ b/pkg/client/src/app/pages/identities/identities.tsx
@@ -235,7 +235,9 @@ export const Identities: React.FunctionComponent = () => {
           ),
         },
         {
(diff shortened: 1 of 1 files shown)


aufi:
Example screenshot:
image

Current solution supports username only (```admin``` in this case), which is part of auth token and doesn't require model Keycloak user in Tackle2 database.

aufi:
> @aufi Do you mind adding the createdBy field to the filter categories as a part of this change?

Yes, is it possible add it in follow-up PR?

ibolton336:
Great - thanks!

ibolton336 (approved):
LGTM

PR #279, opened 2022-06-01
TACKLE-647: Update proxy form save button state when changing enabled/disabled

ibolton336:
Instead of mutating api state directly when toggling enabled state, move the isProxyEnabled state to form state.

This PR checks the form value against the proxy api value to determine if the enabled state has changed. Also - this removes the api call to update enabled state on the switch button click event.
https://issues.redhat.com/browse/TACKLE-647


mturley:
Note: per @jortel, not to be merged to main before 2.0.0 release branch is cut, since the issue is not a blocker for this release.

jortel:
This can now be merged to main.

gildub (approved):
LGTM

PR #282, opened 2022-06-07
Fix missing ENV vars in dev mode: remove unused index.html so it will always be generated from index.html.ejs in dev mode

mturley:
I noticed while troubleshooting an RBAC-related issue that when running the UI via webpack-dev-server, `window._env` was never getting populated and so the UI could not read any environment variables (such as `AUTH_REQUIRED`). It appears this is because we had a `public/index.html` file checked in which did not have the `<script>` which injects `window._env`, and so webpack-dev-server was just serving that instead of generating its own from `index.html.ejs` as it should. Simply deleting that file seems to solve the problem.

This PR also applies Prettier style rules to index.html.ejs.


mturley:
Converted to draft so we don't accidentally merge before GA.

gildub (approved):
LGTM

PR #284, opened 2022-06-07
TACKLE-661: Ensure `checkAccess` is called even when token is undefined, remove unnecessary references to isAuthRequired

mturley:
Follow-up to https://github.com/konveyor/tackle2-ui/pull/283.

Now that the `isAuthRequired` check is part of `checkAccess()`, any place in the UI that was checking both no longer needs to care about `isAuthRequired`. This PR removes it from those places. To ensure we are always calling `checkAccess()` even when the token is undefined, it also replaces guards like:

```ts
let userRoles = token?.realm_access?.roles,
    access = userRoles && checkAccess(userRoles, roles);
```
with:
```ts
let userRoles = token?.realm_access?.roles || [],
    access = checkAccess(userRoles, roles);
```

If the token is defined this makes no difference, but if the token is undefined the `userRoles` array will be `[]` and the `checkAccess` function will still be called (when called with an empty array, it effectively turns into `return !isAuthRequired`).

Needed for: https://issues.redhat.com/browse/TACKLE-661

### DIFF
--- a/pkg/client/src/app/common/RouteWrapper.tsx
+++ b/pkg/client/src/app/common/RouteWrapper.tsx
@@ -18,13 +18,13 @@ export const RouteWrapper = ({
   exact,
 }: IRouteWrapperProps) => {
   const token = keycloak.tokenParsed || undefined;
-  let userRoles = token?.realm_access?.roles,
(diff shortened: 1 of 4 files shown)


jortel (approved):
LGTM

PR #285, opened 2022-06-07
TACKLE-661: Ensure `checkAccess` is called even when token is undefined (missed cases from PR #284)

mturley:
See #284. This solves the same problem described there in a couple more places.

Needed for https://issues.redhat.com/browse/TACKLE-661


jortel (approved):
LGTM

PR #240, opened 2022-05-17
TACKLE-613: Reset local storage on logout for persona storage

ibolton336:
https://issues.redhat.com/projects/TACKLE/issues/TACKLE-613

PR #288, opened 2022-06-08
TACKLE-659: Ensure selected persona will get cleared on logout

mturley:
Resolves https://issues.redhat.com/browse/TACKLE-659

This was previously fixed in https://github.com/konveyor/tackle2-ui/pull/240, but it turns out there is a race condition there. The `keycloak.logout().then(() => { ... });` promise handler sometimes does not run because the logout triggers a page redirect, which can happen before the promise resolves.

To fix this, we clear the persona from localStorage before we initiate the logout. However, we can't simply do this by using the setter provided by LocalStorageContext, because this triggers a React state change which switches the persona dropdown to the default (jarring to see happen while the page is logging out). Instead, this PR directly removes the stored value, circumventing LocalStorageContext so we prevent any React rendering based on this value changing.

Technically speaking it is an antipattern to directly manipulate localStorage that is also synced with React state via LocalStorageContext, but in this case we are about to unload the entire page, so the app is not left with this state out of sync in a way that could cause problems.


aufi (approved):
Gilles might provide better feedback, but looks good to me!

gildub (approved):
LGTM

PR #291, opened 2022-06-09
TACKLE-667: Properly restrict application assessment write actions by RBAC scope

mturley:
Resolves part of https://issues.redhat.com/browse/TACKLE-667.

The other part of that issue is a problem with the RBAC scopes defined in the migrator role (migrators have write access to dependencies, they shouldn't). @jortel is following up on that part.


ibolton336 (approved):
LGTM

PR #295, opened 2022-06-14
TACKLE-665: AddCustomRules: use customFileHandler

gildub:
Use updated version of "file upload multiple" from PF
Use `customFileHandler` prop to have a granular control of duplicate files and XML validation (against schema).
Fixes ability to remove a file from upload list. See https://issues.redhat.com/browse/TACKLE-665 for details.

Note: Latest "file upload multiple" doesn't use a modal to warn about unsupported file selected for upload because files without the correct extension are simply ignored.

### DIFF
--- a/pkg/client/src/app/pages/applications/analysis-wizard/components/add-custom-rules.tsx
+++ b/pkg/client/src/app/pages/applications/analysis-wizard/components/add-custom-rules.tsx
@@ -20,7 +20,7 @@ interface IAddCustomRules {
   readFileData: IReadFile[];
   setReadFileData: (setReadFile: React.SetStateAction<IReadFile[]>) => void;
 }
-interface IParseXMLFileStatus {
+interface IParsedXMLFileStatus {
   state: "valid" | "error";
   message?: string;
 }
@@ -50,7 +50,7 @@ export const AddCustomRules: React.FunctionComponent<IAddCustomRules> = ({
     }
   };
 
-  const validateXMLFile = (data: string): IParseXMLFileStatus => {
+  const validateXMLFile = (data: string): IParsedXMLFileStatus => {
     // Filter out "data:text/xml;base64," from data
     const payload = atob(data.substring(21));
     const validationObject = XMLValidator.validate(payload, {
--- a/pkg/client/src/app/pages/applications/analysis-wizard/components/add-custom-rules.tsx
+++ b/pkg/client/src/app/pages/applications/analysis-wizard/components/add-custom-rules.tsx
@@ -33,7 +33,6 @@ export const AddCustomRules: React.FunctionComponent<IAddCustomRules> = ({
   const [error, setError] = React.useState("");
   const [currentFiles, setCurrentFiles] = React.useState<File[]>([]);
   const [showStatus, setShowStatus] = React.useState(false);
-  const [statusIcon, setStatusIcon] = React.useState("inProgress");
 
   // only show the status component once a file has been uploaded, but keep the status list component itself even if all files are removed
   if (!showStatus && currentFiles.length > 0) {
@@ -41,15 +40,15 @@ export const AddCustomRules: React.FunctionComponent<IAddCustomRules> = ({
   }
 
   // determine the icon that should be shown for the overall status list
-  React.useEffect(() => {
+  const setStatus = () => {
     if (readFileData.length < currentFiles.length) {
-      setStatusIcon("inProgress");
+      return "inProgress";
     } else if (readFileData.every((file) => file.loadResult === "success")) {
-      setStatusIcon("success");
+      return "success";
     } else {
-      setStatusIcon("danger");
+      return "danger";
     }
-  }, [readFileData, currentFiles]);
+  };
 
   const validateXMLFile = (data: string): IParseXMLFileStatus => {
     // Filter out "data:text/xml;base64," from data
@@ -223,7 +222,7 @@ export const AddCustomRules: React.FunctionComponent<IAddCustomRules> = ({
         {showStatus && (
           <MultipleFileUploadStatus
             statusToggleText={`${successfullyReadFileCount} of ${currentFiles.length} files uploaded`}
-            statusToggleIcon={statusIcon}
+            statusToggleIcon={setStatus()}
           >
             {currentFiles.map((file) => (
               <MultipleFileUploadStatusItem
--- a/pkg/client/src/app/pages/applications/analysis-wizard/components/add-custom-rules.tsx
+++ b/pkg/client/src/app/pages/applications/analysis-wizard/components/add-custom-rules.tsx
@@ -119,8 +119,9 @@ export const AddCustomRules: React.FunctionComponent<IAddCustomRules> = ({
           if (validateXMLFile(data as string))
             handleReadSuccess(data as string, file);
           else {
-            // TODO propagate xml validation error
-            const error = new DOMException("Not a valid XML");
+            const error = new DOMException(
+              `File ${file.name} is not a valid XML`
+            );
             handleReadFail(error, 100, file);
           }
         }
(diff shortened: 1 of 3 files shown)


pkg/client/src/app/pages/applications/analysis-wizard/components/add-custom-rules.tsx
```diff
@@ -17,114 +16,178 @@ import { IReadFile } from "../analysis-wizard";
...
+    setShowStatus(true);
+  }
 
-  const getStatusIcon = () => {
-    if (isLoading) {
-      return "inProgress";
-    } else if (currentFiles.every((file) => file.loadResult === "success")) {
-      return "success";
+  // determine the icon that should be shown for the overall status list
+  React.useEffect(() => {
```
mturley, line 34: @gildub it looks like you're reversing a lot of the changes from https://github.com/konveyor/tackle2-ui/pull/264, such as restoring these pieces of redundant/duplicated state and logic that was moved from effects to render-time. Is there a way we can address the issue without losing those code quality improvements? This feels like a bit of a step backwards. At a minimum we definitely shouldn't reintroduce state that is only updated from changes in props (they can simply be plain variables).

If it's because there were a lot of conflicts, we could look at resolving them more cleanly together if you're up for that? I wonder if we can look at a diff of your changes before #264 (do you have those in your reflog from before a rebase, maybe?) to see how we can reapply them to the new code more surgically.

cc @ibolton336

gildub: Of course, I'll organize a meeting. Thanks

ibolton336:
@gildub 
image

Seems as if we are Uploading the file and storing it without enabling/disabling the Add button. Can we integrate this functionality with the Add button? I think this is a regression from main. Thx.

pkg/client/src/app/pages/applications/analysis-wizard/components/add-custom-rules.tsx
gildub: @mturley, regarding this `useEffect()`, it actually comes from latest PF for file upload multiple (and still moving being in beta):
https://codesandbox.io/s/nvslhe.

gildub:
@ibolton336, actually the (requirement)[https://github.com/konveyor/enhancements/tree/master/enhancements/tackle/tackle-windup-integration#custom-rules] says: 
"If any of the files is not valid clicking on "Add" will result in only the valid files being added to the custom rules list".

ibolton336:
> @ibolton336, actually the [requirement](https://github.com/konveyor/enhancements/tree/master/enhancements/tackle/tackle-windup-integration#custom-rules) says: "If any of the files is not valid clicking on "Add" will result in only the valid files being added to the custom rules list".
> 
> Although it make sense when several files are selected for upload, it doesn't really ideal when there is only one file selected.


:+1:  So we need to add some logic for the Add button state as it stands sounds like.

gildub:
@ibolton336, done, button is disabled when there not at least one valid and uploaded file.

pkg/client/src/app/pages/applications/analysis-wizard/components/add-custom-rules.tsx
```diff
@@ -119,8 +119,9 @@ export const AddCustomRules: React.FunctionComponent<IAddCustomRules> = ({
           if (validateXMLFile(data as string))
             handleReadSuccess(data as string, file);
           else {
-            // TODO propagate xml validation error
-            const error = new DOMException("Not a valid XML");
+            const error = new DOMException(
+              `File ${file.name} is not a valid XML`
```
ibolton336, line 123: Small tweak - can we pull out the XML error message from the validation library itself here? Seems like that would be useful info to display.

ibolton336 (changes requested):

pkg/client/src/app/pages/applications/analysis-wizard/components/add-custom-rules.tsx
gildub: Yeah that's a good idea as it seems to be a limitation of initial requirements.  
I saw you used it in your patch and that's on my to do list

pkg/client/src/app/pages/applications/analysis-wizard/components/add-custom-rules.tsx
```diff
@@ -116,11 +116,11 @@ export const AddCustomRules: React.FunctionComponent<IAddCustomRules> = ({
           );
           handleReadFail(error, 100, file);
         } else {
-          if (validateXMLFile(data as string))
-            handleReadSuccess(data as string, file);
+          const validatedXML = validateXMLFile(data as string);
+          if (validatedXML === "") handleReadSuccess(data as string, file);
```
ibolton336, line 120: IMO, This is tough to follow. Can we add an area to store the error and use that to conditionally validate the read success? Just something a little more clear would be nice. 

For example, I think it would be helpful to return an object that has the error message to pull off of if it exists & any other relevant validation info rather than returning a string.

ibolton336 (changes requested):

pkg/client/src/app/pages/applications/analysis-wizard/components/add-custom-rules.tsx
gildub: Good idea, thanks, done.

pkg/client/src/app/pages/applications/analysis-wizard/components/add-custom-rules.tsx
mturley: @gildub that's interesting, thanks for pointing that out. I think that is a flaw in that PF example to be honest. I'll make a note to follow up with them about that, it's definitely an antipattern. If our code works without it we should remove it IMO.

pkg/client/src/app/pages/applications/analysis-wizard/custom-rules.tsx
```diff
@@ -191,6 +181,9 @@ export const CustomRules: React.FunctionComponent = () => {
     });
   });
 
+  console.log("customRulesFiles", customRulesFiles);
+  console.log("readFileData", readFileData);
```
ibolton336, line 185: Please Remove.

gildub: Done

pkg/client/src/app/pages/applications/analysis-wizard/components/add-custom-rules.tsx
gildub: @mturley, ok thanks for feedback, I'll look into it.

pkg/client/src/app/pages/applications/analysis-wizard/components/add-custom-rules.tsx
```diff
@@ -14,117 +13,191 @@ import { XMLValidator } from "fast-xml-parser";
...
-        const newReadFile: IReadFile = {
-          fileName: file.name,
-          loadResult: "success",
-          file: file,
-        };
-        currFiles.push(newReadFile);
-      }
-    }
+  const handleFile = (file: File) => {
+    readFile(file)
```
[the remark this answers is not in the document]
gildub: Yes we can, thanks for pointing it out.

ibolton336:
I am still able to upload duplicate rule files in some cases:

image


image

If I upload a rule file, close the modal, and then reopen the add custom rule file modal & add the same rule file - I see a few bugs. See screen shots.

gildub:
@ibolton336, I've fixed the issue when closing the modal. Thanks for testing.

ibolton336:
> @ibolton336, I've fixed the issue when closing the modal. Thanks for testing.

image

I am still seeing the count increment when adding and then closing the add modal. I see 2 of 1 upon re-opening the modal etc.

ibolton336 (changes requested):
See comment.

gildub:
@ibolton336, I cannot reproduce the count increment using latest patch. Could you please check ?

PR #300, opened 2022-06-17
TACKLE-671: Application form - Packaging Tooltip

gildub:
https://issues.redhat.com/browse/TACKLE-671

### DIFF
--- a/pkg/client/public/locales/en/translation.json
+++ b/pkg/client/public/locales/en/translation.json
@@ -107,6 +107,7 @@
         "appNotAssesedTitle": "Assessment has not been completed",
         "appNotAssessedBody": "In order to review an application it must be assessed first. Please assess the application",
         "assessmentStakeholderHeader": "Select the stakeholder(s) or stakeholder group(s) associated with this assessment.",
+        "binaryPackaging": "If undefined Packaging will default to 'jar'.",
         "continueConfirmation": "Yes, continue",
         "copyAssessmentAndReviewBody": "Some of the selected target applications have an in-progress or complete assessment/review. By continuing, the existing assessment(s)/review(s) will be replaced by the copied assessment/review. Do you wish to continue?",
         "copyAssessmentAndReviewQuestion": "Copy assessment and review?",
--- a/pkg/client/public/locales/es/translation.json
+++ b/pkg/client/public/locales/es/translation.json
@@ -107,6 +107,7 @@
         "appNotAssesedTitle": "La evaluación no se ha completado",
         "appNotAssessedBody": "Para revisar una aplicación, esta debe de ser evaluada primero. Por favor evalúe la aplicación",
         "assessmentStakeholderHeader": "Seleccione a los interesados o grupo de interesados asociados a esta evaluación.",
+        "binaryPackaging": "Cuando no esté definido, se establecerá por defecto en 'jar'.",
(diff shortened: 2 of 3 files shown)


pkg/client/public/locales/en/translation.json
```diff
@@ -107,6 +107,7 @@
         "appNotAssesedTitle": "Assessment has not been completed",
         "appNotAssessedBody": "In order to review an application it must be assessed first. Please assess the application",
         "assessmentStakeholderHeader": "Select the stakeholder(s) or stakeholder group(s) associated with this assessment.",
+        "binaryPackaging": "If undefined Packaging will default to 'jar'.",
```
ibolton336, line 110: A minor nit here - I think this should read: 

"Packaging will default to JAR if left empty." 

or something like that.

ibolton336 (changes requested):

ibolton336 (changes requested):
Can we add a question mark icon next to the field label and add the tooltip or popover to that?  This UX seems odd in that the tooltip is floating to the left of the input box outside of the form and that there is no indication that there is a tooltip present until hovering over the field. 
image

ibolton336:
> @ibolton336, the related issue (TACKLE-671) clearly lacks requirements. To avoid going back and forth we need to make sure requirements are validated by UXD team. The tooltip position is now centered above by default. I don't think a question mark makes sense here but again let's get UXD to determine that. The latter can be added in a separate PR.

No problem. Lets wait for the UXD input before moving forward just so we have something to go off of for future tooltips.

gildub:
@ibolton336, thank you for the follow-up with UXD, I have replaced `tooltip` with `popover`.

ibolton336 (changes requested):
Something isn't looking quite right here with the spacing and icon size. 
image


Also, I think I remember the popover on MTC having a black background? I could be wrong there but seems like the popover is hard to read against the form background.

gildub:
@ibolton336, please re-review with dark popover in place.

ibolton336:
> @ibolton336, please re-review with dark popover in place.

image


I think the form spacing may be the cuplrit here, but the label seems cramped underneath the above input box. Also, the icon size still doesn't seem to be scaled correctly. Would love to nail this so we can reuse across the app. Maybe even create a label with icon local component if that makes sense?

gildub:
@ibolton336, I updated it wrapping Icon with span and css properties and also replaced Icon with `QuestionCircleIcon`

gildub:
@ibolton336, I don't know why we have to go all the way to define CSS at top level as we didn't need such approach with tackle. Anyway it looks much better. Thanks


[image](https://user-images.githubusercontent.com/1901741/178672890-6203204f-6783-444a-8707-2cd699fcf7a5.png)

ibolton336:
> @ibolton336, I don't know why we have to go all the way to define CSS at top level as we didn't need such approach with tackle. Anyway it looks much better. Thanks
> 
> image

Global CSS is useful if we have modifier classes that are available across widely used components. That way we don't have to import the css again in each file. Is there another approach you had in mind?

gildub:
@ibolton336, I understand the global approach, and use local approach, but that's the thing, in forklift, I don't think we're changing anything locally but as you mentioned we might be using a different PF version.

PR #301, opened 2022-06-22
Add unit test support for react query driven components

ibolton336:
This PR adds a way to wrap the react-testing-library render method with the necessary providers to mock react-query driven components. Fortunately most of the unit test setup is fully reusable since Carlos was using react-resting-library already and axios-mock-adapter is already pulled in which handles the api mocks for us. All we have to do now is create new tests for our forms in the manner demonstrated in this PR.  This goal of this PR is to get us up and running with react-testing library. More granular tests are to come. 

Also, I went down a bit of a rabit hole evaluating MSW vs simple jest mocks vs axios-mock-adapter. It seems as though MSW is the most customizeable of the 3 options for API stubbing/mocking & it may be beneficial to switch to it should be run into any issues with axios-mock-adapter. I like the ethos of MSW being driven more at the network layer than the application layer. This may help us if we run into some issues testing RBAC down the road. 

It is also worth noting that most of the existing testing libraries we are using are on old versions that will require some severe breaking changes to the existing tests if we were to bump them. I reverted most of my new work in an effort to keep the older tests around. useAssessApplication seems particularly useful, but the majority of the other tests are outdated since we have moved away from using the components that they are testing. The app is due for a cleanup of some of the older hooks that are no longer in use. 

https://issues.redhat.com/browse/TACKLE-678 - Added a test case to cover this issue.


mturley (approved):
LGTM! I totally agree about MSW being the best approach to mocking, I'm eyeing it for other projects as well. The only thing it is currently lacking is websocket support, but they are working on that!

PR #307, opened 2022-06-30
TACKLE-558: Task cancelation

gildub:
https://issues.redhat.com/browse/TACKLE-558


ibolton336:
@gildub Per discussion with @jortel, a task in a scheduled state should be able to be cancelled also.

ibolton336:
> @ibolton336, tasks that are "Created" (Scheduled), "Running" (InProgress) or "Ready" (Scheduled) can be cancelled.

Can we cancel pending /scheduled tasks?

PR #315, opened 2022-07-13
Tech debt: Remove unused code

ibolton336:
- Removed redundant fetch hooks that have been replaced by react-query
- Converts remaining legacy fetch hooks to react-query. There were only a handful: jobFunctions,business services, delete tag/tagtype


gildub (approved):
LGTM

PR #332, opened 2022-07-30
TACKLE-698 - About: Version 2.1

gildub:
image

### DIFF
--- a/pkg/client/src/app/layout/AppAboutModal/tests/__snapshots__/AppAboutModal.test.tsx.snap
+++ b/pkg/client/src/app/layout/AppAboutModal/tests/__snapshots__/AppAboutModal.test.tsx.snap
@@ -113,7 +113,7 @@ exports[`AppAboutModal 1`] = `
         <TextListItem
           component="dd"
         >
-          2.0
+          2.1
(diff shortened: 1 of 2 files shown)


pkg/client/src/app/layout/AppAboutModal/tests/__snapshots__/AppAboutModal.test.tsx.snap
```diff
@@ -113,7 +113,7 @@ exports[`AppAboutModal 1`] = `
         <TextListItem
           component="dd"
         >
-          2.0
+          2.1
```
ibolton336, line 116: Can we use the constant here as well?

ibolton336 (changes requested):

pkg/client/src/app/layout/AppAboutModal/tests/__snapshots__/AppAboutModal.test.tsx.snap
gildub: I don't think that's helpful since snapshots are automatically generated.

ibolton336 (approved):

PR #333, opened 2022-08-01
Add custom IDs for select toggle

ibolton336:
image


gildub (approved):
Looks good and works for me.

PR #343, opened 2022-08-09
Add missing selectors for dropdown toggles

ibolton336:
Add dropdown toggle selectors for QE

### DIFF
--- a/pkg/client/src/app/pages/applications/analysis-wizard/set-mode.tsx
+++ b/pkg/client/src/app/pages/applications/analysis-wizard/set-mode.tsx
@@ -67,8 +67,9 @@ export const SetMode: React.FunctionComponent<ISetMode> = ({
       <FormGroup label={t("wizard.label.analysisSource")} fieldId="sourceType">
         <SimpleSelect
           toggleAriaLabel="analysis-mode-toggle"
+          toggleId="analysis-mode-toggle"
           variant={SelectVariant.single}
-          aria-label="analysis-mode-select"
+          aria-label="Select user perspective"
           value={toOptionLike(mode, options)}
           onChange={(selection) => {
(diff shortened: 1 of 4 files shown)


pkg/client/src/app/pages/applications/analysis-wizard/set-mode.tsx
```diff
@@ -66,8 +66,9 @@ export const SetMode: React.FunctionComponent<ISetMode> = ({
       </TextContent>
       <FormGroup label={t("wizard.label.analysisSource")} fieldId="sourceType">
         <SimpleSelect
+          toggleAriaLabel="analysis-mode-toggle"
           variant={SelectVariant.single}
-          aria-label="Select user perspective"
+          aria-label="analysis-mode-select"
```
mturley, line 71: FYI, `aria-label` should be a human-readable sentence string, not an identifier. It's what gets read out by screen readers for blind users. "Select analysis mode" or something is actually best here.

For identifiers like this for QE, you can use the `toggleId` prop (SimpleSelect extends most props of PF's Select, and `toggleId` will put an id on the `<button>` used to toggle the select menu).

mturley (changes requested):
Some feedback: we should leave `aria-label` as a human-readable phrase, and use `toggleId` for giving selectors to QE on SimpleSelect / Select.

pkg/client/src/app/pages/applications/analysis-wizard/set-mode.tsx
ibolton336: Thanks Mike - Updated to use toggleID for QE & fixed the aria-label.

mturley (approved):
LGTM, we can take a closer look at the rest of our aria-labels later on if we get to it.

PR #345, opened 2022-08-09
Add Applications and deps tooltip to Import page

aufi:
Application CSV import page imports counts records have two expected types - Application and Dependency.

In order to not confuse user with misleading amount of applications, adding tooltip to related table columns as suggested by Nandini.

Fixes https://issues.redhat.com/browse/TACKLE-631

### DIFF
--- a/pkg/client/src/app/pages/applications/manage-imports/manage-imports.tsx
+++ b/pkg/client/src/app/pages/applications/manage-imports/manage-imports.tsx
@@ -148,8 +148,14 @@ export const ManageImports: React.FC = () => {
       cellTransforms: [truncate],
     },
     { title: t("terms.status"), transforms: [cellWidth(10)] },
-    { title: t("terms.accepted"), transforms: [cellWidth(10)] },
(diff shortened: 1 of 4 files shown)


aufi:
Screenshot:
image

pkg/client/src/app/pages/applications/manage-imports/manage-imports.tsx
```diff
@@ -148,8 +148,14 @@ export const ManageImports: React.FC = () => {
       cellTransforms: [truncate],
     },
     { title: t("terms.status"), transforms: [cellWidth(10)] },
-    { title: t("terms.accepted"), transforms: [cellWidth(10)] },
```
aufi, line 151: I couldn't set less than 10 to the transforms[cellWidth], the 10 value overlapped with the label, after removal looked the same even after resizing window. Please recommend better solution if this would't be accepted, thanks!

ibolton336 (approved):

PR #347, opened 2022-08-10
TACKLE-734: Re-add missing delete tag functions

ibolton336:
Add back in delete functionality which was inadvertently removed.

### DIFF
--- a/pkg/client/src/app/pages/controls/tags/tags.tsx
+++ b/pkg/client/src/app/pages/controls/tags/tags.tsx
@@ -32,7 +32,6 @@ import {
 } from "@app/shared/components";
 
 import { dedupeFunction, getAxiosErrorMessage } from "@app/utils/utils";
-import { deleteTag, deleteTagType } from "@app/api/rest";
 import { Tag, TagType } from "@app/api/models";
 
 import { NewTagTypeModal } from "./components/new-tag-type-modal";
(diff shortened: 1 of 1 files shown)


mturley (approved):
LGTM.

Looking at why this happened: It seems `deleteTagType()` shouldn't let you pass a function at all, and it looks like the culprit is the `any` type on the `mutate` function [here](https://github.com/konveyor/tackle2-ui/blob/039893e45176e55bf23a09df7aefebaea2a8e135/pkg/client/src/app/queries/tags.ts#L26-L30):
```ts
export interface ITagTypeMutateState {
  mutate: any;
  isLoading: boolean;
  error: any;
}
```

So it might be a good idea to look through the codebase and remove `any` wherever we can, especially when dealing with API client stuff. Honestly, with the react-query stuff, you can often get pretty far just not specifying a return type on your query/mutation hooks at all and letting the inferred return type of `useQuery` or `useMutation` come through.

For example, on the main branch right now if I look at `useDeleteTagTypeMutation` and just remove the `ITagTypeMutateState` return type from [this line](https://github.com/konveyor/tackle2-ui/blob/main/pkg/client/src/app/queries/tags.ts#L92), the hook now has a naturally inferred return type that includes `UseMutateFunction` with the right type arguments:

Screen Shot 2022-08-10 at 4 08 22 PM

That alone caused the stuff you're fixing here to become errors:

Screen Shot 2022-08-10 at 4 09 11 PM

Food for thought 🍻

PR #349, opened 2022-08-10
TACKLE-634: Mention unknown records in Import rejected title

aufi:
Adding more descriptive tooltip title to CSV import table mentioning
that rejected records might include also records with unwknown record
type (not application or dependency).

Fixes https://issues.redhat.com/browse/TACKLE-634


aufi:
image

ibolton336 (approved):

PR #355, opened 2022-08-16
TACKLE-694: Fix items per page bug

ibolton336:
https://issues.redhat.com/browse/TACKLE-694


mturley:
Oh dang good catch. I should include that fix in https://github.com/konveyor/lib-ui/pull/73 if I ever get back around to it

mturley (approved):

PR #356, opened 2022-08-16
Tackle-689: Tag /Tag type name filter fix

ibolton336:
Allow search across tag types and tag names for name multisearch option

https://issues.redhat.com/browse/TACKLE-689?filter=-1


pkg/client/src/app/pages/controls/tags/tags.tsx
```diff
@@ -140,6 +145,16 @@ export const Tags: React.FC = () => {
...
           .map((tag) => ({ key: tag?.name, value: tag?.name }))
+          .concat(
+            tagTypes?.map((tagType) => ({
+              key: tagType?.name,
+              value: tagType?.name,
+            }))
+          )
+          .sort((a, b) => {
+            if (a.value && b.value) return a?.value.localeCompare(b?.value);
+            else return 0;
```
mturley, line 156: Huh, I didn't realize single-line if could be used with else like this.

mturley (approved):
LGTM

pkg/client/src/app/pages/controls/tags/tags.tsx
ibolton336: I think you are right to point this out as odd syntax. I moved to curly brackets for else statements. Looks to be more maintainable.

PR #364, opened 2022-08-19
TACKLE-695: Change application form validation 

ibolton336:
Application requires business service but does not in 1.x.

https://issues.redhat.com/browse/TACKLE-695

### DIFF
--- a/pkg/client/src/app/pages/applications/components/application-form/application-form.tsx
+++ b/pkg/client/src/app/pages/applications/components/application-form/application-form.tsx
@@ -225,15 +225,16 @@ export const ApplicationForm: React.FC<ApplicationFormProps> = ({
       description: string()
         .trim()
         .max(250, t("validation.maxLength", { length: 250 })),
-      businessService: object().shape({
-        id: string()
-          .trim()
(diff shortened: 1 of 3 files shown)


pkg/client/src/app/pages/applications/components/application-form/application-form.tsx
```diff
@@ -342,7 +345,7 @@ export const ApplicationForm: React.FC<ApplicationFormProps> = ({
       name: formValues.name.trim(),
       description: formValues.description.trim(),
       comments: formValues.comments.trim(),
-      businessService: formValues.businessService
+      businessService: formValues?.businessService?.id
```
mturley, line 348: This will result in `businessService` being `0` in the payload when none is selected, not sure if that is desired. I wonder if it's okay to switch the initial value back to null, and have this be
```
businessService: formValues?.businessService?.id || null
```

mturley: Ah, makes sense ok.

mturley (approved):
LGTM!

PR #368, opened 2022-08-22
TACKLE-xxxxxx: Add filter selectors for QE

ibolton336:


mturley (commented):
`aria-label` should be a plain human-readable phrase with spaces since it's used for accessibility tools. We should either use `id` for these identifiers or use full phrases and see if QE is okay keying off those full strings.

mturley (approved):
LGTM

PR #382, opened 2022-08-31
Convert repo to use NPM workspaces

ibolton336:
Re-implements https://github.com/konveyor/tackle2-ui/pull/8/files now that we have bandwidth to take this on. Pending review from @rayfordj to see if this will work with cachito upstream.


rayfordj:
Thanks, @ibolton336!  I'll kick the tires on this shortly

rayfordj (approved):
/lgtm

Approved; in so much as Cachito no longer fails to deliver source(s) downstream

PR #387, opened 2022-09-02
TACKLE-759: Filter out current app from copy to list when selecting all

ibolton336:
When selecting ALL applications in the bulk copy assessment/review form, the source application was being selected. This led to the api not receiving the correct list of apps & caused some errors with overwriting the source application data. 

This PR also prevents applications without an assessmend & review from being copied from. [See here](https://github.com/konveyor/tackle2-ui/pull/387/files#diff-38b35274202426ae47eb724b71f4d5dc739f74351444c5bb38bfad24d9d1c636R394)

### DIFF
--- a/pkg/client/src/app/pages/applications/applications-table-assessment/applications-table-assessment.tsx
+++ b/pkg/client/src/app/pages/applications/applications-table-assessment/applications-table-assessment.tsx
@@ -391,6 +391,7 @@ export const ApplicationsTable: React.FC = () => {
     }
     if (
       row.review &&
+      applicationAssessment?.status === "COMPLETE" &&
       checkAccessAll(userScopes, ["assessments:patch", "reviews:post"])
     ) {
       actions.push({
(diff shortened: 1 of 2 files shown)


mturley (commented):
This works great, but when you click "Select all" it does still show a selected checkbox on the current app (even though it's disabled, it looks kinda funky).

I wonder if we can have the checkbox only appear checked if it's also enabled, or prevent the current app from ending up in selection state in the first place. Hmm, the latter would probably require an update to useSelectionState...

mturley (approved):
I'm happy with this if you want to merge it now, we can follow up to address the checked-but-disabled-and-unused checkbox issue separately.

One thing I'm considering for that is adding an `isItemSelectable` option to `useSelectionState` similar to what I did in the selectable table examples in PatternFly (`isRepoSelected` there): https://www.patternfly.org/v4/components/table#composable-selectable-with-checkbox

PR #390, opened 2022-09-06
TACKLE-759: Fix fetch risks not being called after assessment data change

ibolton336:
- Adds assessment to dependency array for fetch risks useEffect function.


mturley (approved):
Good stuff!

PR #392, opened 2022-09-06
Upgrade lib-ui, use `isItemSelectable` on copy assessment and review form

mturley:
In the Copy Assessment and Review modal, this PR uses the new `isItemSelectable` option added to `useSelectionState` in https://github.com/migtools/lib-ui/pull/112 to prevent the "select all" action from checking the disabled checkbox next to the current application. This also removes the `.filter()` added in https://github.com/konveyor/tackle2-ui/pull/387 since it is no longer necessary (the current application can never become part of the selection state, so there's no reason to filter it out).

This PR also renames some variables that used the `item` and `row` naming in a slightly confusing way, since we are mapping over `Application` objects to create `IRow` objects. I figured it would be nice to be clear what is an `app` and what is a `row` type-wise via the names.

### DIFF
--- a/pkg/client/package-lock.json
+++ b/pkg/client/package-lock.json
@@ -11,7 +11,7 @@
       "dependencies": {
         "@hookform/resolvers": "^2.8.0",
         "@hot-loader/react-dom": "^17.0.2",
-        "@migtools/lib-ui": "^8.4.1",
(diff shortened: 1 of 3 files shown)


gildub (approved):
LGTM

Issue #393, opened 2022-09-07
Session handling improvements


ibolton336:
Currently , the cookie is being re-hydrated by react query every 5 seconds or so. This can be improved by relying on keycloaks internal token update functionality that gets triggered when any api request is fired via an axios interceptor. We can use this hook to update the cookie and handle any token expiry logic. 

Also, the current token expires after 5 minutes. This can be lengthened to improve the experience when using 3rd party served html pages that rely on the cookie for authentication. These pages also need to redirect the user to /login once the token is expired given that the cookie is stale and not being refreshed due to a closed "main tackle tab".

PR #397, opened 2022-09-08
Cookie management improvements

ibolton336:
Addresses #393 

- Change token expiration to 1 hour to reduce auto logout token expiry redirects on 3rd party api served external html pages
- Remove useFetchCookie query hook in favor of leaning on keycloak-js's internal token initialization logic. Since we run the keycloak token validation check on every outgoing Axios request, we can refresh the cookie at this time if it is stale. 
- Automatically redirect to the main apps home page if an authentication error occurs on a 3rd party api served external html page. This will trigger the keycloak authentication init logic upon routing here. 

Depends on: https://github.com/konveyor/tackle2-ui/pull/401

There is an issue with navigating to the /login route. A blank page is rendered and requires the user to click a navigation item to continue using the app. This is fixed by [this PR](https://github.com/konveyor/tackle2-ui/pull/401)

### DIFF
--- a/pkg/client/src/app/common/KeycloakProvider.tsx
+++ b/pkg/client/src/app/common/KeycloakProvider.tsx
@@ -58,7 +58,6 @@ export const KeycloakProvider: React.FunctionComponent<
                 return new Promise<string>((resolve, reject) => {
                   if (keycloak.token) {
                     keycloak
-                      //TODO lengthen refresh time
                       .updateToken(60)
                       .then((refreshed) => {
                         if (refreshed) {
--- a/pkg/client/src/app/common/KeycloakProvider.tsx
+++ b/pkg/client/src/app/common/KeycloakProvider.tsx
@@ -2,6 +2,7 @@ import { initInterceptors } from "@app/axios-config";
 import { isAuthRequired } from "@app/Constants";
 import i18n from "@app/i18n";
 import keycloak from "@app/keycloak";
+import { deleteCookie, getCookie, setCookie } from "@app/queries/cookies";
 import { AppPlaceholder } from "@app/shared/components";
 import { Flex, FlexItem, Spinner } from "@patternfly/react-core";
 import { ReactKeycloakProvider } from "@react-keycloak/web";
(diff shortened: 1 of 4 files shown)


pkg/client/src/app/common/KeycloakProvider.tsx
```diff
@@ -14,6 +15,17 @@ interface IKeycloakProviderProps {
 export const KeycloakProvider: React.FunctionComponent<
   IKeycloakProviderProps
 > = ({ children }) => {
+  const checkAuthCookie = () => {
+    let currentCookie = getCookie("keycloak_cookie");
+    if (currentCookie !== "" && currentCookie !== null) {
+    } else {
+      currentCookie = keycloak?.token || "";
+
+      if (currentCookie != "" && currentCookie != null) {
```
mturley, line 24: Nit: can we avoid `==` and `!=` and replace them with `===` and `!==`? It doesn't particularly matter here but it's good to be consistent.

Also, this line could also become `if (currentCookie) {`. It feels a little weird to nest them though right?

It seems like this whole block of code is: `Check whether the cookie is set, and if not, set it to the keycloak token if present`. Since `currentCookie` is never used outside this function anyway, I think you could actually just simplify the whole function to this:

```ts
  const checkAuthCookie = () => {
    if (!getCookie("keycloak_cookie") && keycloak?.token) {
      setCookie("keycloak_cookie", keycloak.token, 365);
    }
  };
```

pkg/client/src/app/common/KeycloakProvider.tsx
```diff
@@ -14,6 +15,17 @@ interface IKeycloakProviderProps {
 export const KeycloakProvider: React.FunctionComponent<
   IKeycloakProviderProps
 > = ({ children }) => {
+  const checkAuthCookie = () => {
+    let currentCookie = getCookie("keycloak_cookie");
+    if (currentCookie !== "" && currentCookie !== null) {
```
mturley, line 20: You can simplify this `if`-`else` since the `if` case is empty here, and just use
```ts
if (currentCookie === "" || currentCookie === null) {
  currentCookie = keycloak?.token || "";
  // ...
```
Or honestly since empty-string and null are both falsy this could just be:
```ts
if (!currentCookie) {
  currentCookie = keycloak?.token || "";
  // ...
```

pkg/client/src/app/queries/cookies.ts
```diff
@@ -1,52 +1,27 @@
...
-    {
-      enabled: !!keycloak.token,
-      refetchInterval: 5000,
-      refetchIntervalInBackground: true,
-      refetchOnWindowFocus: false,
+export const getCookie = (name: CookieNames) => {
+  let cookieArr = document.cookie.split(";");
+  for (let i = 0; i < cookieArr.length; i++) {
+    let cookiePair = cookieArr[i].split("=");
+    if (name == cookiePair[0].trim()) {
```
mturley, line 22: Same nit about using `===` over `==` here.

mturley (commented):
Some small nitpicks, otherwise LGTM!

mturley (approved):
LGTM!

PR #402, opened 2022-09-12
Persist app inventory page filters to `sessionStorage`

mturley:
Adds an optional `storageKey` argument to `useFilterState` and uses it on the application inventory tables. When a `storageKey` is passed to `useFilterState` like this, the filter values will be persisted to sessionStorage using the new `useSessionStorage` hook added to lib-ui in https://github.com/migtools/lib-ui/pull/114. If `storageKey` is not passed, regular React state will be used instead.

Filters set on either app inventory page will now be persisted for the lifetime of the current browser tab (reloading the page or navigating away/back won't clear them, but closing and reopening the tab will). Separate tabs can have separate filter values because they don't share session storage.

### DIFF
--- a/pkg/client/package-lock.json
+++ b/pkg/client/package-lock.json
@@ -11,7 +11,7 @@
       "dependencies": {
         "@hookform/resolvers": "^2.8.0",
         "@hot-loader/react-dom": "^17.0.2",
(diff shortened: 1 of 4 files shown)


ibolton336 (approved):
Works really well! Nice job getting this upstream, too.

PR #403, opened 2022-09-13
Tech Debt: Turn on no implicit any rule in tsconfig

ibolton336:


gildub:
Helps with https://github.com/konveyor/tackle2-ui/issues/408

gildub (approved):
LGTM

mturley:
Nice!

mturley:
@ibolton336 I think yes, but in doing so we may identify places where it's not as easy as simply replacing the `any` with another simple type. We might find things unfolding from that that we want to open separate issues for, etc. But yeah, maybe a PR for any obvious types we can substitute in would be good. There might also be places where we can rely on type inference instead of an explicit type, though. hovering things to see what VS Code thinks the types are might give hints there. In some places we might want to introduce generics.

If you guys are starting on that, if you don't mind can you exclude proxy-form.tsx and identity-form.tsx to avoid conflicts with https://github.com/konveyor/tackle2-ui/pull/404? There may be a couple other small places where I've messed with types as well (`client/src/app/utils/utils.ts` for example).

PR #406, opened 2022-09-14
Add Azure target with 3rd party filtering

gildub:
https://issues.redhat.com/browse/TACKLE-794

- Regroups `targets` and `transformationTargets` data into "targets.ts". This will be easier in case we delegate that information to the hub (more likely with the coming feature of custom cards).
- Feed values from open and proprietary targets. 
- Handle edge case when card has only one target but we still need to display a select. Such card might have 3rd party targets and we need to have a consistent display along with the card's text.
- Add Azure targets, filtered by PROFILE envar.

image


jortel (dismissed):
Small change.
Franco suggested another we use $PROFILE=(konveyor|mta) instead of $PRODUCT.  I think this is better.  Please update your PR.

gildub:
Updated and ready for final review.

client/src/app/pages/applications/analysis-wizard/targets.ts
```diff
@@ -0,0 +1,198 @@
...
+    description:
+      "A comprehensive set of rules for migration from Apache Camel 2 to Apache Camel 3.",
+    options: ["camel"],
+    iconSrc: multiplyIcon,
+    forceSelect: false,
+  },
+  {
+    label: "Azure",
+    description:
+      "Upgrade your Java application so it can be deployed in different flavors of Azure.",
```
ibolton336, line 191: @gildub Do we have any approved copy for this description? I just spotted one small grammar tweak. 

``` Upgrade your Java application so that it can be deployed in different flavors of Azure.```

**Review by ibolton336 — CHANGES_REQUESTED** [review]

File: client/src/app/pages/applications/analysis-wizard/targets.ts (hunk 1 above)
Comment (ibolton336) [reply client/src/app/pages/applications/analysis-wizard/targets.ts:191]: Just saw that this is how it is in the mockups. Disregard.

File: client/src/app/Constants.ts [hunk 2]
```diff
@@ -20,10 +20,13 @@ export interface IEnvVars {
   KEYCLOAK_CLIENT_ID: string;
   KEYCLOAK_SERVER_URL: string;
   UI_INGRESS_PROXY_BODY_SIZE: string;
+  PROFILE: string;
 }
 export const ENV: IEnvVars = (window as { [key: string]: any })["_env"] || {};
 export const isAuthRequired = ENV.AUTH_REQUIRED !== "false";
 export const uploadLimit = ENV.UI_INGRESS_PROXY_BODY_SIZE || "500m";
+export const isPropietaryAllowed =
```
ibolton336, line 28: Can we think of a better name for this one since it is likely to be re-used for the branding also?

ibolton336 (changes requested):

client/src/app/Constants.ts
ibolton336: In MTC we set the env var like so:

```
export enum BrandType {
  Crane = 'Crane',
  RedHat = 'RedHat',
}

export const APP_BRAND: BrandType = (process.env.BRAND_TYPE as BrandType) || BrandType.Crane;
```

So that we could render the brand image like:

```
 <img
                src={APP_BRAND === BrandType.RedHat ? logoRedHat : logoCrane}
                alt="Logo"
                className={
                  APP_BRAND === BrandType.RedHat ? styles.redhatLogoStyle : styles.craneLogoStyle
                }
              />

``` 

Not saying we should do this exactly, but just suggesting something slightly clearer.

ibolton336:
Looking great @gildub - Just one comment and I think we can merge this. We can also address my comment in the branding ENV var related PR. Up to you.

PR #410, opened 2022-09-15
Adjust server node_modules

gildub:


Dockerfile
```diff
@@ -3,9 +3,7 @@ FROM registry.access.redhat.com/ubi8/nodejs-16 as builder
 USER 0
 COPY . .
 WORKDIR "/opt/app-root/src" 
-RUN npm install && npm run build -w client
-WORKDIR "/opt/app-root/src/server" 
-RUN npm install
+RUN npm install -w client && npm run build -w client && rm -rf node_modules && npm install -w server
```
mturley, line 6: Any reason not to just make these separate `RUN` lines?

```
RUN npm install -w client
RUN npm run build -w client
RUN rm -rf node_modules
RUN npm install -w server
```

mturley (approved):

Dockerfile
mturley: Ok, makes sense.

PR #412, opened 2022-09-15
Lockfile changes on npm install after workspaces refactor

mturley:
FYI @gildub @ibolton336 when I run `npm install` now (using npm 8.15.0) I get some changes to the package-lock.json that differ from what's currently tracked in git. It looks like just this `devOptional` flag got removed from a few places. Not sure what it means...


ibolton336:
Seeing the same thing.

ibolton336 (approved):

PR #400, opened 2022-09-08
npm workspaces

gildub:
This replaces pkg/client, pkg/server and pkg/qa-tests to respectively client, server and qa to work around cachito issue: https://issues.redhat.com/browse/CLOUDBLD-11019

Local and quay.io images have been tested: they are building and running.

PR #420, opened 2022-09-20
Downgrade css-loader to 5.2.7 to fix font loading issues

mturley:
As part of #400, `css-loader` was upgraded from `^5.2.6` to `^6.7.1`. Something about this upgrade seems to have caused our PatternFly Fonts to fail to load.

image (27)

Troubleshooting this with @ibolton336, I tried to `curl` one of the font files exposed by webpack-dev-server by copying its path from the browser dev tools network tab, and I found that its contents were not binary font data, but JS:
```
$ curl http://localhost:8080/101eb68a37edfa8a2f9a.woff
export default __webpack_public_path__ + "fonts/overpass-bold.woff";
```

This indicated that it was something about a loader change, and `css-loader` was the only webpack loader with a recent major version bump. Indeed, downgrading it seems to have solved the problem. I'm not certain I understand why, though... here are the breaking changes listed in the [css-loader 6.0.0 release notes](https://github.com/webpack-contrib/css-loader/releases/tag/v6.0.0).

### DIFF
--- a/client/package.json
+++ b/client/package.json
@@ -84,7 +84,7 @@
     "camelcase": "^6.3.0",
     "case-sensitive-paths-webpack-plugin": "^2.4.0",
     "copy-webpack-plugin": "^11.0.0",
-    "css-loader": "^6.7.1",
+    "css-loader": "^5.2.7",
     "css-minimizer-webpack-plugin": "^3.4.1",
     "cypress": "8.4.0",
     "cypress-file-upload": "^5.0.8",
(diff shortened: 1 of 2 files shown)


mturley:
cc @gildub : do you think it's a problem to downgrade css-loader here? was there some specific reason to upgrade it as part of #400? I'd love to understand how we can fix the font loading with the newer css-loader but maybe we should downgrade first, not sure

gildub (approved):

gildub:
@mturley, for the record following our brief discussion, the issue css-loader during PR#400 was more likely related to dependencies issue but I can't remember the details.

We effectively need to go to the bottom of this in the meantime of a downgrade.

Thanks for investigating.

PR #421, opened 2022-09-21
TACKLE-795: Downstream branding integration

ibolton336:
- Adds Favicon, branding text, and brand image for downstream image based on the newly added [PROFILE env var](https://github.com/konveyor/tackle2-operator/pull/113) 

image

https://issues.redhat.com/browse/TACKLE-795


client/config/webpack.dev.js
```diff
@@ -25,9 +26,9 @@ module.exports = merge(common("development"), {
       // In dev mode, populate window._env at build time
       filename: "index.html",
       template: path.resolve(__dirname, "../public/index.html.ejs"),
-      favicon: path.resolve(__dirname, "../public/favicon.ico"),
```
mturley, line 28: It works, so take this or leave it, but I wonder if instead of having `<link rel="icon">` in the EJS file at all we should just have the logic here?
```ts
favicon: path.resolve(__dirname, `../public/${brandType || 'konveyor'}-favicon.ico`
```
that would result in the right `<link>` getting added to the HTML I think. Currently webpack duplicates the `<link rel="icon" href="favicon.ico" />` twice anyway. I think that would mean we don't have to bother with copying the favicon file in the CopyPlugin config above?

We could move this to `client/config/webpack.common.js` so we don't have to repeat it for dev and prod, as well.

client/public/index.html.ejs
```diff
@@ -2,14 +2,18 @@
...
-    <title>Tackle</title>
-    <link rel="icon" href="favicon.ico" />
+    <% if(brandType == 'mta'){ %>
+      <title>Migration Toolkit for Applications</title>
+      <link rel="icon" href="mta-favicon.ico" />
+      <% } else{ %>
+      <title>Tackle</title>
+      <link rel="icon" href="konveyor-favicon.ico" />
+      <% } %>
     <meta id="appName" name="application-name" content="Konveyor Tackle UI" />
```
mturley, line 12: We probably want to have these `<meta>` tags for `application-name` and `description` also use the right brand name I think? I'm not sure how those get used.

client/public/manifest.json
```diff
@@ -1,13 +1,6 @@
 {
   "short_name": "tackle-ui",
   "name": "Tackle UI",
-  "icons": [
```
mturley, line 4: Do you know what this manifest.json is for? I imagine having the favicon in here is important for some reason. If we use the webpack config approach above I think we can restore these lines since I think it will just use `favicon.ico` as the compiled filename regardless of the source file.

client/src/app/Constants.ts
```diff
@@ -14,13 +14,25 @@ import {
   ProposedAction,
   Risk,
 } from "@app/api/models";
+export interface IEnvVars {
```
mturley, line 17: I think you're regressing the changes from https://github.com/konveyor/tackle2-ui/pull/440 here? You should be able to remove `IEnvVars` from here, restore the `import { ENV } from "./env";` instead of having `export const ENV` here, and then below do:

```ts
export const APP_BRAND = ENV.PROFILE as BrandType
```

I don't think you need the `|| BrandType.Konveyor` if you do that, since `"konveyor"` is already the default for `PROFILE` provided by `ENV`.

client/src/app/pages/applications/analysis-wizard/targets.ts
```diff
@@ -193,9 +194,10 @@ export const transformationTargets: ITransformationTargets[] = [
     label: "Azure",
     description:
       "Upgrade your Java application so it can be deployed in different flavors of Azure.",
-    options: isPropietaryAllowed
-      ? ["azure-appservice", "azure-aks"]
-      : ["azure-appservice"],
+    options:
+      APP_BRAND === BrandType.Konveyor
+        ? ["azure-appservice", "azure-aks"]
```
mturley, line 199: This isn't really a comment about this change, but what's the deal here? users aren't allowed to analyze AKS apps if they're using the Red Hat branded version of MTA?

mturley (changes requested):
A couple of small things here. Looking great!

client/src/app/pages/applications/analysis-wizard/targets.ts
ibolton336: Yeah @gildub Can probably speak to this more but this is a recent change that adds "proprietary" targets for upstream images only.

gildub: @mturley, let's chat about that so I can explain what's happening.

client/src/app/Constants.ts
```diff
@@ -14,13 +14,16 @@ import {
...
 
+export enum BrandType {
+  Konveyor = "konveyor",
+  MTA = "mta",
+}
+
 export const isAuthRequired = ENV.AUTH_REQUIRED !== "false";
 export const uploadLimit = ENV.UI_INGRESS_PROXY_BODY_SIZE || "500m";
-export const isPropietaryAllowed =
-  ENV.PROFILE === "konveyor" || ENV.PROFILE === "";
```
gildub, line 23: By default the app is in upstream context with PROFILE being either "konveyor" or not defined ("").
By not allowing not defined, the logic is changing a bit here.
Maybe that's because initially the values for PROFILE envar weren't fully known.
But from an open source project viewpoint I think it's important we keep that way, so the brand could be anything.

gildub (commented):
Please see inline.
This might be an important detail.

mturley (approved):
LGTM!

client/src/app/pages/applications/analysis-wizard/targets.ts
mturley: Sounds good, I'll resolve this thread and we can discuss on slack

PR #437, opened 2022-09-22
Switch from webpack's watch mode to webpack-dev-server, fix local prod mode

mturley:
This is something I started tinkering with to see how it would work, and I think it might be a good approach if you guys are ok with it. If not that's totally cool. Personally I think it improves the developer experience a bit. Definitely open to feedback!

I made a few other tweaks as well here. If we decide against using webpack-dev-server, I'll follow up with a smaller PR to take some of these other minor improvements anyway.

This PR:
* Renames some package.json scripts in the client workspace to be a little more clear what they're for:
  * `server` and `build:dev` are now `start:dev:server` and `start:dev:client`
  * `proxy`, `proxy:hub` and `proxy:keycloak` are now `port-forward`, `port-forward:hub` and `port-forward:keycloak`
* Changes the `start:dev:client` script to use `webpack serve` instead of `webpack`
* Adds a `port-forward` script to the root package.json so we can run `npm run port-forward` and `npm run start` from the repo root (also applies Prettier styles to that file)
* Adds a `clean` script to the client workspace for removing an existing dist directory, and adds it to the `prebuild` script (installs the `rimraf` dev dependency to do this in a cross-platform manner)
* Replaces the `watch` and `watchOptions` options in client/config/webpack.dev.js with `devServer` and `optimization` based on [the documentation here](https://webpack.js.org/guides/development/#using-webpack-dev-server). webpack-dev-server is set up to use port 9000.
* Moves the `concurrently` dependency from the repo root package.json into the client workspace package.json
* Adds the `app.set("views", path.join(__dirname, "../client/dist"));` line to the Express server so we no longer need to copy the `index.html.ejs` file from `client/dist` to `server/views` in the Dockerfile. This also fixes local usage of `npm run start` after a build (previously it would have errors due to the missing `server/views/index.html.ejs` file, now it just finds the one in `client/dist` just like everything else it's serving).
* In the `app.get("*", (_, res) => {` route of the Express server, which should now only be reached if you are trying to view the UI in production mode, replaces the dev-mode behavior of serving of the `index.html` file with a message directing you to port 9000 and explaining how to run the UI in local production mode.
* Updates the explanation of the local dev environment in the README (and applies Prettier styles to that file).

With these changes, we can now run `npm run start:dev` and then view the UI at localhost:9000 with faster start times and browser auto-reload on file changes. We can also run the UI in a simulated production mode by running `npm run build`, then separately running `npm run proxy` and `npm run start`, and viewing the UI at localhost:8080.

The `devServer` configuration includes a simple proxy configuration:
```ts
  devServer: {
    port: 9000,
    proxy: {
      // NOTE: Any future non-UI paths handled by ../../server/index.js should be added here.
      "/auth": "http://localhost:8080",
      "/hub": "http://localhost:8080",
    },
    historyApiFallback: true,
  },
```

This just simply forwards any requests to `/auth` and `/hub` to port 8080, where Express will then proxy them in the same manner it does in production (including cookie management). The goal here is to avoid replicating any of the Express server logic and truly only have UI-related configuration in webpack. I'll admit it is a little bit of a bummer we have to maintain this list of routes (if we add a third proxy route to Express in the future we'll need to add it here so it works in dev mode). Perhaps in the future we could rename the routes to `/api/auth` and `/api/hub` and then this configuration could simply proxy anything under `/api`.

The `historyApiFallback` option tells webpack-dev-server to serve up the index.html file it generates for any URL, since we are using client-side routing (history API routing). This prevents errors when doing a hard page reload.

If you are running `npm run start:dev`, since Express is still running, if you go to port 8080 in your browser you'll get this message:

Screen Shot 2022-09-23 at 4 41 59 PM

This is to help us adjust our muscle memory / bookmarks and prevent any other developers who may have this set up from being confused when they come back after this change. It may be a bit overkill though, I'll remove it if you guys don't think we need it :)


ibolton336 (approved):
Awesome - this is everything I hoped it would be and more.

ibolton336 (changes requested):
Seeing an issue when doing a hard refresh on some of the pages:

http://localhost:9000/applications/assessment-tab

gildub:
@mturley, that sounds good. 

We need to completely remove pathfinder from the proxy service, since its API is now served through the hub.
Checkout https://github.com/konveyor/tackle2-ui/pull/422.

mturley:
@ibolton336 ah, looks like we just need the [`historyApiFallback` option.](https://webpack.js.org/configuration/dev-server/#devserverhistoryapifallback)

mturley:
This should be ready for re-review @ibolton336 @gildub. I fixed Ian's issue and also renamed some package.json scripts and updated the PR description.

ibolton336 (approved):

PR #440, opened 2022-09-23
Define defaults for missing ENV variables

mturley:
Instead of defining `IEnvVars` as an interface, this defines an object `ENV_VAR_DEFAULTS` as a constant, defines the `EnvVarKey` type based on its keys, and defines the `EnvVars` type as `Record<EnvVarKey, string>`. That way, we can map over `Object.keys(ENV_VAR_DEFAULTS)` at run-time to patch in any vars missing from `window._env` based on a set of defaults, and preserve the type safety (the only valid env var keys are the ones defined in `ENV_VAR_DEFAULTS`).

Moves all this to a separate `env.ts` file since it's a bit more verbose now.

Screen Shot 2022-09-23 at 2 11 41 PM

This also uses `{ ...window._env }` to define `ENV` before we patch in the defaults for missing vars, so that we can still inspect `window._env` in the console if we want to see exactly what was passed by the server (without defaults added).

### DIFF
--- a/client/src/app/env.ts
+++ b/client/src/app/env.ts
@@ -16,7 +16,7 @@ declare global {
     _env: EnvVars;
   }
 }
-export const ENV: EnvVars = window._env || {};
+export const ENV: EnvVars = { ...window._env } || {};
 
 ENV_VAR_KEYS.forEach((key) => {
   if (!ENV[key]) ENV[key] = ENV_VAR_DEFAULTS[key];
--- a/client/src/app/env.ts
+++ b/client/src/app/env.ts
@@ -1,0 +1,30 @@
+const ENV_VAR_KEYS = [
+  "AUTH_REQUIRED",
+  "KEYCLOAK_REALM",
+  "KEYCLOAK_CLIENT_ID",
+  "KEYCLOAK_SERVER_URL",
+  "UI_INGRESS_PROXY_BODY_SIZE",
+  "PROFILE",
+] as const;
+
+export type EnvVars = Record<typeof ENV_VAR_KEYS[number], string>;
+
+const ENV_DEFAULTS: EnvVars = {
+  AUTH_REQUIRED: "true",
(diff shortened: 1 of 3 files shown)


client/src/app/env.ts
```diff
@@ -0,0 +1,30 @@
...
+  "PROFILE",
+] as const;
+
+export type EnvVars = Record<typeof ENV_VAR_KEYS[number], string>;
+
+const ENV_DEFAULTS: EnvVars = {
+  AUTH_REQUIRED: "true",
+  KEYCLOAK_REALM: "",
+  KEYCLOAK_CLIENT_ID: "",
+  KEYCLOAK_SERVER_URL: "",
```
gildub, line 16: We could also add those defaults : 
KEYCLOAK_REALM="tackle"
KEYCLOAK_CLIENT_ID="tackle-ui"

gildub (dismissed):
That's neat !

client/src/app/env.ts
```diff
@@ -0,0 +1,30 @@
...
+  KEYCLOAK_REALM: "tackle",
+  KEYCLOAK_CLIENT_ID: "tackle-ui",
+  KEYCLOAK_SERVER_URL: "",
+  UI_INGRESS_PROXY_BODY_SIZE: "",
+  PROFILE: "konveyor",
+};
+
+declare global {
+  interface Window {
+    _env: EnvVars;
```
ibolton336, line 23: 🤟

client/src/app/env.ts
```diff
@@ -0,0 +1,30 @@
...
+  }
+}
+export const ENV: EnvVars = window._env || {};
+
+ENV_VAR_KEYS.forEach((key) => {
+  if (!ENV[key]) ENV[key] = ENV_DEFAULTS[key];
```
ibolton336, line 29: 🤟

ibolton336 (approved):

PR #450, opened 2022-10-03
Use custom notifications

gildub:
Replaces `@redhat-cloud-services/frontend-components-notifications` with custom notifications based on React context provider.
Also allows to share the same approach as `konveyor/forklift-ui` (which code was used for this one) making it a candidate for defining the notifications as a common component of `@migtools/lib-ui`.


client/src/app/shared/components/Notifications.tsx
```diff
@@ -0,0 +1,55 @@
...
+import { NotificationsContext } from "../notifications-context";
+
+export type notificationType =
+  | "success"
+  | "info"
+  | "warning"
+  | "danger"
+  | "default";
+
+export interface INotification {
```
mturley, line 16: It seems like maybe these types should be defined in the notifications-context file rather than the Notifications component file, since the context provider is where notifications are stored and passed down? not sure if that matters.

client/src/app/shared/components/Notifications.tsx
```diff
@@ -0,0 +1,55 @@
...
+  title: string;
+  variant: notificationType;
+  key?: string;
+  message?: string | JSX.Element;
+  actionClose?: boolean;
```
mturley, line 21: It seems like (almost?) every use of `pushNotification()` has `actionClose: true`, I wonder if that should be a default? we could swap it to a `hideCloseButton` boolean that is usually excluded, maybe.

It also seems strange to call this `actionClose` as a boolean since the `actionClose` prop of Alert itself is a ReactNode.

client/src/app/shared/components/Notifications.tsx
```diff
@@ -0,0 +1,55 @@
+import React from "react";
+import {
+  Alert,
+  AlertActionCloseButton,
+  AlertGroup,
+} from "@patternfly/react-core";
+import { NotificationsContext } from "../notifications-context";
+
+export type notificationType =
```
mturley, line 9: Instead of duplicating the string union here, we can remove the `notificationType` type and instead import `AlertProps` from PatternFly, and use `AlertProps['variant']` as the type below.

Either that, or we could use `export interface INotification extends Pick<AlertProps, 'variant'> {`, both have the same effect.

mturley (commented):
Very minor nitpicks. This looks great!

client/src/app/shared/components/Notifications.tsx
mturley: Oh, also, for `message` you can use `React.ReactNode` as the type instead of `string | JSX.Element`. ReactNode will allow anything that can be a JSX child.

mturley:
Hmm, looks like `bulk-copy-notifications-container` is still using the old package and that's causing CI to fail

client/src/app/shared/components/Notifications.tsx
gildub: @mturley, yeah I was wondering about a default `actionClose` too. It's now wrapped through a default notification.

I remember you and I having that discussion about JSX.Element vs ReactNode in my early days with React ;) 
A `React.ReactNode` is better for `message` and `actionClose`.

client/src/app/shared/components/Notifications.tsx
gildub: Nice !
I went with first approach. 
Thanks

client/src/app/shared/notifications-context.tsx
```diff
@@ -1,5 +1,21 @@
...
+  | "danger"
+  | "default";
+
+export type INotification = {
+  title: string;
+  variant: notificationType;
+  key?: string;
+  message?: React.ReactNode;
+  hideCloseButton?: boolean;
+  actionClose?: React.ReactNode;
```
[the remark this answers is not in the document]
gildub: @mturley, right, done.

mturley (approved):
Nice work!

PR #325, opened 2022-07-27
TACKLE-635: Follow up to page redirect on last item delete

ibolton336:
This PR addresses the issue described in [635](https://issues.redhat.com/browse/TACKLE-635) across all areas reported in the app. Instead of redirecting to the first page, the app will now show the next chronological page in the list. For example, If on page 3 and the last item is removed, navigate the user to page 2.

PR #451, opened 2022-10-03
TACKLE-772: On deleting a table item, remain on current page unless we deleted the last item on the page

mturley:
Fixes [TACKLE-772](https://issues.redhat.com/browse/TACKLE-772)

This was a regression caused by https://github.com/konveyor/tackle2-ui/pull/325, which fixed related issues [TACKLE-635](https://issues.redhat.com/browse/TACKLE-635) and [TACKLE-712](https://issues.redhat.com/browse/TACKLE-712). The original bug was that deleting an item never caused a page change, which left an unexpected empty state if the last item on a page was deleted. https://github.com/konveyor/tackle2-ui/pull/325 changed the behavior such that when an item is deleted, if it was the last item on the page the table will go back 1 page, else the table will go to the first page.

That `else` case was causing this regression. If the deleted item was not the last one on the page, we should stay on the page we were on rather than going back to the first page. Initially I fixed this by removing the else case from each of those delete callbacks, but I then realized we could just replace that per-table behavior entirely by enforcing the "don't allow us to be on a page greater than the last page" logic within `usePaginationState` itself in a `useEffect`.

As part of testing this fix, I found another bug that this PR also fixes: If you deleted the very last item in the table, the above logic would decrement the page number to 0 (page numbers start at 1). If you then added a new item, the pagination state would be screwed up and display negative numbers:

Screen Shot 2022-10-03 at 1 51 53 PM

To fix this second issue, this PR modifies `usePaginationState` so the returned `setPageNumber` function can never set a page number lower than 1.

### DIFF
--- a/client/src/app/pages/applications/applications-table-analyze/applications-table-analyze.tsx
+++ b/client/src/app/pages/applications/applications-table-analyze/applications-table-analyze.tsx
@@ -418,9 +418,6 @@ export const ApplicationsTableAnalyze: React.FC = () => {
         onConfirm: () => {
           if (row.id) {
             deleteApplication({ id: row.id });
-            if (currentPageItems.length === 1 && paginationProps.page) {
-              setPageNumber(paginationProps.page - 1);
-            }
           }
         },
       })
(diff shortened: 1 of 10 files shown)


mturley:
I have a better idea for how to fix this, marking as draft for a moment.

ibolton336 (approved):
Tested and LGTM!

PR #459, opened 2022-10-04
[Backport v2.1.2] TACKLE-795: Manually cherry-pick branding pr 421

ibolton336:
Backports #421 manually


pkg/client/src/app/Constants.ts
```diff
@@ -22,11 +22,15 @@ export interface IEnvVars {
...
+export enum BrandType {
+  Konveyor = "konveyor",
+  MTA = "mta",
+}
 export const ENV: IEnvVars = window["_env"] || {};
 export const isAuthRequired = ENV.AUTH_REQUIRED !== "false";
 export const uploadLimit = ENV.UI_INGRESS_PROXY_BODY_SIZE || "500m";
-export const isPropietaryAllowed =
-  ENV.PROFILE === "konveyor" || ENV.PROFILE === "" ? true : false;
+export const APP_BRAND = ENV.PROFILE as BrandType;
```
mturley, line 33: @ibolton336 maybe on 2.1.2 we should have this be:

```ts
export const APP_BRAND = (ENV.PROFILE || BrandType.Konveyor) as BrandType;
```

Just because below you have places where if PROFILE isn't set you'll end up with Red Hat branding, which is probably not ideal for someone building the upstream app

ibolton336: Good call - I'll make the change.

mturley (approved):
LGTM! Good stuff

PR #460, opened 2022-10-06
Use ConfirmDialog where needed without dispatch

gildub:
Replaces Redux approach (using the store to trigger confirmation and maintain its state) with `ConfirmDialog` boxes and local state.

Covers the following pages:
- application-assessment
- application-table-analyze
- application-table-assessment
- controls/stakeholders
- controls/stakeholder-groups
- controls/business services
- controls/job functions
- controls/tags and controls/tag-types
- manage-imports
- identity


mturley (commented):
I'm not sure if we want all the `ConfirmDialogProps` as an object in state. In other apps we've had a separate `<ConfirmDialog>` (ConfirmModal in forklift and crane) for each use case, with props inline, and the only state is the `isOpen` (your `openConfirmDialog` here). If you need multiple confirm dialogs on a page you can render more than one, and just have each shown from a separate state like `isConfirmDeleteDialogOpen`, `isConfirmSomethingElseDialogOpen` etc. That way the props are in the JSX as normal and state is kept minimal.

Also `openConfirmDialog` as the name of a boolean feels weird since it's a verb, it makes it seem like a function. I'd use something like `isConfirmDialogOpen` or `confirmDialogOpen`.

These are just my opinion, if you disagree I'm open to that! The solution here does work fine.

gildub:
@mturley, well I ended up with `ConfirmDialogProps` as an object while evolving from legacy code and trying to be not too disruptive. And it felt kinda new ! But you're right we don't need to hold the props as a state especially if we need several dialog boxes. 

+1 for `isConfirmDialogOpen`, it's a better practice.

client/public/locales/en/translation.json
```diff
@@ -178,6 +178,7 @@
         "binaryVersion": "Version",
         "businessCriticality": "Business criticality",
         "businessService": "Business service",
+        "businessServiceDeleted": "Business service deleted",
```
mturley, line 181: Are these locale file changes intended as part of this PR?

PR #461, opened 2022-10-10
TACKLE-814: Add error parsing check for cyclic deps error

ibolton336:
Signed-off-by: ibolton336 <ibolton@redhat.com>

- This PR accounts for the "error" field in the api error response object. This is needed to correctly parse the "cyclic dependencies are not allowed" error. 
Notes: We should add i18n for these special case errors that are passed from the api. The current string is:

image

I didn't want to change too much here is this is a regression bug. Something to consider for future releases, though.


gildub (approved):
LGTM

PR #469, opened 2022-10-12
Use React.FC vs React.FunctionComponent

gildub:
React.FC is actually an alias for React.FunctionComponent.
So either one can be used meanwhile let's be consistent.
This patch sets FC everywhere.

Signed-off-by: Gilles Dubreuil <gilles@redhat.com>

### DIFF
--- a/client/src/app/common/KeyDisplayToggle.tsx
+++ b/client/src/app/common/KeyDisplayToggle.tsx
@@ -1,7 +1,7 @@
-import * as React from 'react';
-import { Button } from '@patternfly/react-core';
-import EyeSlashIcon from '@patternfly/react-icons/dist/js/icons/eye-slash-icon';
-import EyeIcon from '@patternfly/react-icons/dist/js/icons/eye-icon';
+import * as React from "react";
+import { Button } from "@patternfly/react-core";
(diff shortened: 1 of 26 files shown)


mturley (approved):
Sounds good to me. @ibolton336 ?

ibolton336:
Sounds good

ibolton336:
> Sounds good to me. @ibolton336 ?

Thoughts on https://medium.com/raccoons-group/why-you-probably-shouldnt-use-react-fc-to-type-your-react-components-37ca1243dd13 ?

mturley:
That's interesting. It sounds like that article is advocating for using neither `React.FC` or `React.FunctionComponent` and instead just having functions with the prop types annotated on the argument only?

I get the point about children I suppose. Seems weird to have an unused prop (I usually put `children: React.ReactNode` in the prop types if I'm going to use children anyway). The point about compound components makes sense if you plan to use compound components like that, but that pattern has always felt weird to me.

Seems like not a huge deal either way. If you want to switch to inline prop types and ditch `React.FC` I'd be fine with that, but I don't think it gains us much.

mturley:
Oh, ha, look in the comments:
> As of react 18.0 there are no children in React.FC. Thus implict children is not relevant anymore

PR #479, opened 2022-10-13
Move getBusinessServiceByID to react query and cleanup dead code

ibolton336:
- Adds a new function in the businessservices.ts query file for getting business services by ID
- Remove type defs for rq inside bus-svc query in favor of inferred types
- removes dead code that was referenced by the old applications toolbar and removes the old applications toolbar.


mturley (approved):
LGTM

PR #481, opened 2022-10-13
[Tech debt] Move common useFetch to react query

ibolton336:
- Replaces dated Tackle 1.2 instances of useFetch- a custom fetch hook that implements common fetching utility functions. This logic can be easily replaced by react-query. 

- Should be merged before [this related PR. ](https://github.com/konveyor/tackle2-ui/pull/485 ) which replaces the  polling logic for bulk notification status with a react-query polling approach we have agreed on in the past.


gildub (approved):
LGTM

PR #484, opened 2022-10-14
Only consider windup tasks on the Analysis table

mturley:
Adds a `filters` argument to the `useFetchTasks` hook so that we can query only tasks that matter for a particular view. For now the only filter is `addon`, and it is only used in the Analysis table to filter on `{ addon: "windup" }`. These filters are applied BEFORE the selection logic in the query hook (finding the latest unique task per application).

cc @djzager

Signed-off-by: Mike Turley <mike.turley@alum.cs.umass.edu>


djzager (approved):

ibolton336 (approved):

ibolton336:
Nice! Also as a side effect looks like this filtered out addons with type : admin also. Probably never resulted in a bug because we pull the latest task for status in the analysis wizard, but makes me feel better. 
image

mturley:
Ah, fair point! Yeah, that's probably a good thing

PR #485, opened 2022-10-17
[Tech Debt / Bug fix] - Convert to query polling for bulk notifications

ibolton336:
Bug fix:
- Currently on main, the app crashes when you are attempting a bulk copy. This is because the BulkNotificationsContainer references the NotificationsContextProvider while being at the same nested depth. This PR removes the BulkNotificationsContainer, so this will be fixed. 

Tech Debt: 
- The BulkNotificationsProvider can be safely replaced with the polling mechanism introduced here https://github.com/konveyor/tackle2-ui/pull/217/files which uses ReactQuery without the need for the extra redux dependencies. 

NOTE:  Currently, there is no way to poll for bulk review copy completion due to an api limitation. We do not receive any info in the POST response to the reviews/bulk copy route on the _hub_ API as we do with the bulk copy assessment POST response using the _pathfinder_ api . This leaves us reliant on the assessment copy completion from pathfinder for displaying notifications relating to both bulk copy assessments and bulk copy assessments and reviews.


client/src/app/pages/applications/applications-table-assessment/applications-table-assessment.tsx
```diff
@@ -257,6 +259,27 @@ export const ApplicationsTable: React.FC = () => {
...
+  };
+  const onHandleCopyError = (error: AxiosError) => {
+    setIsSubmittingBulkCopy(false);
+    pushNotification({
+      title: getAxiosErrorMessage(error),
+      variant: "danger",
+    });
+  };
+
+  const { mutate: createCopy, isCopying } = useCreateBulkCopyMutation({
```
ibolton336, line 279: This is the new bulk copy mutation that houses the new query polling logic which replaces the [submit logic here](https://github.com/konveyor/tackle2-ui/pull/485/files#diff-0631a29b250364a7dd66110d8c5567ed72ef22f5091b180a8be69fa818799bbdL295)
as well as the polling logic in useFetch

gildub (approved):
LGTM

PR #486, opened 2022-10-21
Bug fix: Prevent duplicate notifications & fix dismiss bug

ibolton336:
- Currently, you cannot dismiss a push notification that was created without a 'key' prop. Rather than adding this prop to all areas of the app, I changed the logic to use title for the key. 
- Adds a check for duplicate notifications before pushing


gildub (approved):
LGTM

PR #491, opened 2022-10-27
[Bug Fix] Update notification type for analysis submission

ibolton336:
- Currently we are showing the user an error notification when submitting an application for analysis. This should be an info notification.


gildub (approved):
LGTM

PR #497, opened 2022-10-31
Fix Husky desire for some reorder translation files

gildub:


ibolton336:
@gildub Are we still using the pre-commit hooks that run the reorder for us?

gildub:
@ibolton336, yes it's still kicking in.

ibolton336 (approved):

PR #510, opened 2022-11-04
Refactor Analysis Wizard to use step-level yup schemas and HookFormPF* abstractions, clean up styles, add StringListField component

mturley:
The intention here is to create a clean-as-possible example of how we can best manage form state in a wizard. In addition to consolidating schemas, general cleanup and use of the new form components, this PR adds some validation that was missing and improves the UX of the string-list fields (include/exclude packages, exclude rules tags) which are now abstracted into a StringListField component.


gildub (approved):
LGTM

gildub (approved):
LGTM

PR #534, opened 2022-11-29
Refactor taskGroup handling

gildub:
Taskgroup deletion and recreation is now handled by `onModeChange()`  which is passed down to set-mode" component in order to avoid using `useEffect()`.

The initialization of the first taskgroup is handled only first time by initial useEffect().


mturley (approved):
Nice! LGTM

PR #542, opened 2022-12-12
Not require Accept header in API Proxy

aufi:
Proxy response handler expected Accept header to check if it is "application/json", Maayan reported failure in this code when the Accept header was ommited (and the request was fulfilled by backend hub API). Adding safe navigation operator to Accept headers lookup.

Signed-off-by: Marek Aufart <maufart@redhat.com>


ibolton336:
Nice find @aufi

ibolton336 (approved):

PR #544, opened 2022-12-14
:sparkles: add "addons" tab for running arbitrary tackle addons

djzager:
This PR introduces a new tab called "Addons" that allows users to run arbitrary addons (assuming extra addons are installed). The "Addons" tab is only visible if extra addons are visible to the hub.


client/src/app/pages/applications/components/application-list-expanded-area/application-list-expanded-area-extras.tsx
```diff
@@ -0,0 +1,75 @@
...
+      <Tr key="Name">
+        <Td dataLabel="Addon">{task.addon}</Td>
+        <Td dataLabel="Status">
+          <ApplicationAnalysisStatus state={task.state? task.state : "No task"} />
+        </Td>
+      </Tr>
+    )
+  })
+
+  var tasksGroup;
```
ibolton336, line 38: Using 'let' is preferable here - It is just an updated block scoped form of var. We typically use either let or const for defining variables. Minor nitpick for consistency.

client/src/app/pages/applications/components/application-list-expanded-area/application-list-expanded-area-extras.tsx
```diff
@@ -0,0 +1,75 @@
...
+import { Application, Task } from "@app/api/models";
+import { ApplicationTags } from "../application-tags";
+import { ApplicationAnalysisStatus } from "../../components/application-analysis-status";
+
+export interface IApplicationListExpandedAreaProps {
+  application: Application;
+  tasks: Task[];
+}
+
+export const ApplicationListExpandedAreaExtras: React.FC<
```
ibolton336, line 21: Maybe we can change the file/component name(s) to "Addons" instead of "Extras"? WDYT?

client/src/app/pages/applications/components/application-list-expanded-area/application-list-expanded-area-extras.tsx
```diff
@@ -0,0 +1,75 @@
...
+      </Tr>
+    )
+  })
+
+  var tasksGroup;
+  if (tasks.length > 0) {
+    tasksGroup = (
+      <DescriptionListGroup>
+        <DescriptionListTerm>{t("terms.tasks")}</DescriptionListTerm>
+        <TableComposable aria-label="Simple table">
```
ibolton336, line 43: We could wrap the table in a description list description so that the addons table takes up the entire width of the container. 
image

client/src/app/api/models.tsx
```diff
@@ -400,7 +405,7 @@ export interface Taskgroup {
   id?: number;
   name: string;
   addon: string;
-  data: TaskData;
+  data: any; // TODO(djzager) - may want to fix this
```
ibolton336, line 408: Any reason not to use the TaskData type here?

client/src/app/pages/applications/components/application-list-expanded-area/application-list-expanded-area-extras.tsx
```diff
@@ -0,0 +1,75 @@
...
+          <Thead>
+            <Tr>
+              <Th>Addon</Th>
```
ibolton336, line 46: Will need internationalization for these.

ibolton336 (changes requested):
Just a couple of minor things - looking good overall!

client/src/app/pages/applications/components/application-list-expanded-area/application-list-expanded-area-extras.tsx
mturley: You may also want to use `variant="compact"` on the inner table, and maybe also `borders={false}`.

client/src/app/api/models.tsx
djzager: This point still remains but the other issues have been addressed.

ibolton336 (approved):
LGTM - There is one outstanding bug with the code-editor component that prevents right clicking inside the window to paste contents. Seems to be an issue with the component itself rather than with the implementation here.

Issue #198, opened 2022-05-12
Sometimes selectors for DOM objects are not unique


ibragins:
Sometimes selectors for objects are not unique and UI automated tests can't figure out which object is needed.
Examples:, 
1.  When configuring proxy - credentials selection both for HTTP and HTTPS are the same
2. Perspective view droplist has all the same selectors as, for example, droplist when creating tags
If I will find anything else - I will add them here as comments

ibolton336:
```
id="https-proxy-credentials-select-toggle"
id="http-proxy-credentials-select-toggle"
```

Looking like this is resolved now @ibragins - Please reopen if you are seeing anything that doesn't look right!

Issue #443, opened 2022-09-28
[QE testing] form field selectors: align on using `id` where possible to simplify QE tests


mturley:
Related to https://github.com/konveyor/tackle2-ui/issues/198. Ideally we can fix both together.

For rationale, see the "Accessibility and QE selectors" section at the bottom of the description in https://github.com/konveyor/tackle2-ui/pull/404. The main conclusions there are:
* We need a unique `id` attribute on each form input in order to provide accessible labels without redundant `aria-label` attributes
* We should NOT use `aria-label` as a selector for QE tests since it is user-facing text that is subject to change or be removed (it is used by screen-readers for accessibility)
* QE tests should use `id` selectors where possible and work with dev on what selector to use when `id` cannot be provided on a case-by-case basis. When this comes up it is likely a bug in PatternFly and we can provide a real `id` once that bug is addressed.

https://github.com/konveyor/tackle2-ui/pull/404 included changes to support the above for the Proxy settings form and the Identity form (the Create/Edit Credential modal in the Admin view). That PR includes heavy refactoring that should not be backported to 2.1.x, but we should create a separate PR to backport matching `id` / `aria-label` changes to 2.1.2 once that release branch exists. We should then make sure we follow up to help transition relevant QE tests to use the new `id` selectors.

As more forms are refactored to use the new react-hook-form components, we will also be aligning their field attributes in a similar way, and any of those that are merged before 2.1.2 code freeze should also have their selector changes backported as part of this issue.

mturley:
PRs containing changes that will break QE tests, which should have custom backport PRs to align the 2.1.2 branch to use their new field selectors:

* https://github.com/konveyor/tackle2-ui/pull/404

Please edit this comment to add more if you identify them.

gildub:
QE Requests (7 Sept 2023 - initially from https://github.com/konveyor/tackle2-ui/issues/1336) :

- [x] Cancel button ID to be 'cancel' across the UI.
- [x] Submit button ID to be 'submit' across the UI.
- [ ] Top table level Kebab ID (for example id="toolbar-kebab") across the UI so it can be identified consistently (same ID) on every page where there is a top level dropdown (kebab).
  - [x] appliactions-table-assessment.tsx: Toolbar and row IDs
  - [ ] applications-table-analysis.tsx: Toolbar and row IDs
  - [ ] migration-waves.tsx : Toolbar ID
[...]

PR #553, opened 2023-01-10
QE: Add missing select IDs

ibolton336:
Closes https://github.com/konveyor/tackle2-ui/issues/198

This fixes the case where an explicit unique ID was not present for a select component "toggle" button. This should allow QE to reliably find a unique ID for select components rather than rely on aria-label or data-testID. 

This PR also aims to convert all aria-labels to human readible format as suggested https://github.com/konveyor/tackle2-ui/issues/443


gildub (approved):
LGTM

PR #554, opened 2023-01-11
Add RWX_SUPPORTED envar

gildub:
I took the liberty to change the variant of the "clear repository" button from link to default to be able to clearly show when the button is disabled. I believe this is also allow to better follow the application standard.

Screenshot from 2023-01-13 11-43-03

A tooltip is provided when the button is disabled which is the case when RWX is not supported (RWX_SUPPORTED = "false").

Screenshot from 2023-01-13 11-41-53

https://github.com/konveyor/tackle2-operator/issues/140


client/public/locales/en/translation.json
```diff
@@ -14,6 +14,7 @@
         "cancel": "Cancel",
         "checkDocumentation": "Check documentation",
         "clearAllFilters": "Clear all filters",
+        "clearRepositoryNotSupported": "This action is disabled when RWX option is not supported",
```
ibolton336, line 17: We should get this text vetted by @jortel  / @rromannissen

ibolton336 (changes requested):

client/src/app/pages/repositories/Mvn.tsx
```diff
@@ -157,26 +159,24 @@ export const RepositoriesMvn: React.FC = () => {
...
                   size={15}
                   width={10}
                 />
                 {"  "}
-                <Button
-                  variant="link"
-                  isInline
-                  isDisabled={isCleaning}
-                  onClick={() => setIsConfirmDialogOpen(true)}
+                <ConditionalTooltip
```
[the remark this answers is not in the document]
ibolton336: Maybe we should confirm that there is no issue changing the button type since this is how it was mocked up.

https://raw.githubusercontent.com/konveyor/enhancements/master/enhancements/tackle/tackle-git-maven-integration/images/maven-config.png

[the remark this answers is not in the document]
ibolton336: https://github.com/konveyor/enhancements/tree/master/enhancements/tackle/tackle-git-maven-integration - Looks like we may have forgotten to add this tooltip within the same file also. Would be nice to include it as a part of this update.

client/public/locales/en/translation.json
gildub: @ibolton336, let me run ask them.

client/src/app/pages/repositories/Mvn.tsx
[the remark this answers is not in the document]
gildub: @ibolton336, yeah let's get @rromannissen to validate the style change

client/public/locales/en/translation.json
jortel: Don't have much of a preference but would omit the word `option`.

jortel: Also, you'll need to gray/disable the size field as well when RWX not supported.

client/src/app/pages/repositories/Mvn.tsx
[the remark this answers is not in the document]
ibolton336: @gildub Do you mind rolling that back to the original style and addressing it in a future PR? Just want to keep this rolling so we have time to test.

client/public/locales/en/translation.json
rromannissen: What about "This action is disabled when RWX volumes are not available"?

rromannissen:
LGTM! Nice idea with the tooltip!

ibolton336 (approved):

client/public/locales/en/translation.json
gildub: @rromannissen, we're going to follow-up in separate PR as this one got merged in order to have the feature checked-in ASAP. Thanks

PR #559, opened 2023-01-13
feat: Adjust button space and tooltip when RWX is disabled 

gildub:
Follow up from https://github.com/konveyor/tackle2-ui/pull/554

### DIFF
--- a/client/public/locales/en/translation.json
+++ b/client/public/locales/en/translation.json
@@ -14,7 +14,7 @@
         "cancel": "Cancel",
         "checkDocumentation": "Check documentation",
         "clearAllFilters": "Clear all filters",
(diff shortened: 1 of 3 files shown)


ibolton336 (approved):
LGTM

ibolton336:
Do you have any suggestions on PR naming conventions /labels @gildub ?  I was thinking we could stick to a few buckets for naming as PF does https://github.com/patternfly/patternfly-react/blob/main/CONTRIBUTING.md#pr-labels
Something like: 

```
feat:
bug: 
qe: 
chore:
doc:
css:
```

gildub:
@ibolton336, why not, although I don't have lots of experience with such approach.

PR #561, opened 2023-01-16
bug: Fix message and use i18n

gildub:
https://issues.redhat.com/browse/TACKLE-948

### DIFF
--- a/client/public/locales/en/translation.json
+++ b/client/public/locales/en/translation.json
@@ -122,6 +122,7 @@
         "copyAssessmentQuestion": "Copy assessment?",
         "couldNotFetchBody": "Resource doesn't exists or you don't have access to it",
         "couldNotFetchTitle": "Not available",
+        "duplicateTag": "A tag with this name already exists. Use a different name.",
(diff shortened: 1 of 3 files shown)


client/public/locales/en/translation.json
```diff
@@ -122,6 +122,7 @@
         "copyAssessmentQuestion": "Copy assessment?",
         "couldNotFetchBody": "Resource doesn't exists or you don't have access to it",
         "couldNotFetchTitle": "Not available",
+        "duplicateTag": "A tag with this name already exists. Use a different name.",
```
ibolton336, line 125: Noticed in some PRs we are using punctuation at the end of the sentence & some we are not. Not sure what is the best way but lets be consistent.

ibolton336 (changes requested):

client/public/locales/en/translation.json
gildub: I agree about being consistent.
Here is a special case of having 2 sentences in the same message. 
It would be weird to finish the first sentence with a punctuation and not the second sentences. That won't be consistent.
Let's address the other single messages with/without a punctuation in a separate PR.

ibolton336 (approved):

PR #577, opened 2023-01-18
bug: Add missing filename when uploading a local binary

ibolton336:
- Currently, an object is assigned to the filename string. This should be the binary name.


jortel (approved):
LGTM

PR #579, opened 2023-01-23
:sparkles: Bump nodejs to version 18.x 

gildub:
From nodejs version 18.x, [LTS support started with 18.12.1](https://nodejs.org/en/download/releases/).

This makes the build based on ubi8/nodejs-18 (building image) and ubi8/nodejs18-minimal (Runtime pod) - Both are currently based on nodejs 18.12.1) 

The same is also supported on UBI9 allowing to benefit from latest features and performances from Red Hat Entreprise Linux 9.

### DIFF
--- a/.github/workflows/ci-actions.yml
+++ b/.github/workflows/ci-actions.yml
@@ -15,7 +15,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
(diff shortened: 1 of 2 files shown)


gildub:
Putting this one on hold until we have feedback from the other sub projects about the UBI8 to UBI9 migration constraints.

PR #616, opened 2023-02-08
:bug: Husky prepare fix and prettier hook

ibolton336:
- Adds a fallback for prettier formatting when not using the format-on-save feature provided by the vscode prettier plugin
- Fixes our previous husky hook configuration to accommodate our change to npm workspaces


gildub (approved):
LGTM

PR #618, opened 2023-02-08
:bug: Implement clear notifications 

ibolton336:
https://issues.redhat.com/browse/MTA-28

### DIFF
--- a/client/src/app/shared/notifications-context.tsx
+++ b/client/src/app/shared/notifications-context.tsx
@@ -34,14 +34,15 @@ export const NotificationsProvider: React.FunctionComponent<
 > = ({ children }: INotificationsProvider) => {
   const [notifications, setNotifications] = React.useState<INotification[]>([]);
 
-  const pushNotification = (notification: INotification) => {
-    const hasDupe = notifications.find((n) => n.title === notification.title);
-    if (!hasDupe) {
-      setNotifications([
-        ...notifications,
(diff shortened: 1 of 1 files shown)


client/src/app/shared/notifications-context.tsx
```diff
@@ -34,14 +34,15 @@ export const NotificationsProvider: React.FunctionComponent<
...
-    }
+  const pushNotification = (
+    notification: INotification,
+    clearNotificationDelay?: number
+  ) => {
+    setNotifications([
+      ...notifications,
+      { ...notificationDefault, ...notification },
+    ]);
+    setTimeout(() => setNotifications([]), clearNotificationDelay || 10000);
```
gildub, line 45: Can we replace hard coded value with a const store at higher level ?

gildub (changes requested):

client/src/app/shared/notifications-context.tsx
ibolton336: Any particular reason?

gildub: For me a hard coded value deep in the application is no go.
But ultimately is if we want the user to have control over it through an option then we'll have to lift it up.

gildub: Never mind my comment, I totally misread it :(

gildub (approved):
LGTM

PR #625, opened 2023-02-10
:sparkles: DnD grab handle implementation

ibolton336:
Implement "grab handler "approach for drag and drop to allow other UI components to operate freely without falling into the draggable zone.

Previously, we were seeing dropdowns and kebab menus fail to register click events. This solves the issue by allowing the user to use a patternfly "grip" icon to drag and drop the cards rather than allowing the user to click & drag anywhere on the card.


gildub (approved):
LGTM

PR #633, opened 2023-02-13
:bug: Try using new ubuntu release

jmontleon:
Signed-off-by: Jason Montleon <jmontleo@redhat.com>


jmontleon:
The failure looks like a kernel / overlay fs driver issue that docker and podman have experienced. I'd like to try an updated ubuntu release to see if it gets us anything.

djzager (approved):

Issue #630, opened 2023-02-13
Integrate custom rules file upload to custom targets form


ibolton336:
- Add ability for multiple custom rules files to be uploaded to a custom target
- potentially parse custom rule files for source/target for association to windup. May be handled by Hub API 
- XML validation against multiple schema files
- Other validation for file name & type
- Associate the uploaded files to the custom target once they have been uploaded to the api.

PR #634, opened 2023-02-14
:sparkles: Custom target api integration

ibolton336:
#630


client/src/app/api/rest.tsx
```diff
@@ -449,10 +452,15 @@ export const deleteIdentity = (id: number): AxiosPromise => {
...
-export const getSettingById = (id: number | string): AxiosPromise<boolean> => {
+export const getSettingById = (id: number | string): AxiosPromise<any> => {
   return APIClient.get(`${SETTINGS}/${id}`, jsonHeaders);
 };
 
+export const updateBundleOrderSetting = (
+  obj: BundleOrderSetting
+): AxiosPromise<Setting> => {
+  return APIClient.put(`${SETTINGS}/${obj.key}`, obj.value, jsonHeaders);
+};
```
gildub, line 463: We better use axios as APIClient is to be deprecated (as discussed).

client/src/app/api/rest.tsx
```diff
@@ -560,11 +564,27 @@ export const removeFileTaskgroup = ({
...
 
-export const updateMigrationTarget = (obj: MigrationTarget) =>
-  axios.put(`${RULESETS}/${obj.name}`, obj);
+export const updateRuleBundle = (obj: RuleBundle) =>
+  axios.put(`${RULEBUNDLES}/${obj.name}`, obj);
+
+export const createRuleBundle = (obj: RuleBundle) =>
+  axios.post<RuleBundle>(RULEBUNDLES, obj).then((response) => response.data);
+
+export const deleteRuleBundle = (id: number): AxiosPromise =>
```
gildub, line 573: AxiosPromise type can actually be inferred.

client/src/app/queries/rulebundles.ts
```diff
@@ -0,0 +1,116 @@
...
+) => {
+  const queryClient = useQueryClient();
+  const { isLoading, mutate, error } = useMutation(updateRuleBundle, {
+    onSuccess: (res) => {
+      onSuccess(res);
+      queryClient.invalidateQueries([RuleBundlesQueryKey]);
+    },
+    onError: (err: AxiosError) => {
+      onError(err);
+    },
```
gildub, line 42: Can be replaced with `onError: (err) => console.log(err),`

client/src/app/queries/rulebundles.ts
```diff
@@ -0,0 +1,116 @@
...
+) => {
+  const queryClient = useQueryClient();
+  const { isLoading, mutate, error } = useMutation(createRuleBundle, {
+    onSuccess: (res) => {
+      onSuccess(res);
+      queryClient.invalidateQueries([RuleBundlesQueryKey]);
+    },
+    onError: (err: AxiosError) => {
+      onError(err);
+    },
```
gildub, line 63: Can be replaced with `onError: (err) => console.log(err),`

client/src/app/queries/rulebundles.ts
```diff
@@ -0,0 +1,116 @@
...
+) => {
+  const queryClient = useQueryClient();
+  const { isLoading, mutate, error } = useMutation(createImageFile, {
+    onSuccess: (res) => {
+      onSuccess(res);
+      queryClient.invalidateQueries([]);
+    },
+    onError: (err: AxiosError) => {
+      onError(err);
+    },
```
gildub, line 84: Can be replaced with `onError: (err) => console.log(err),`

client/src/app/queries/rulebundles.ts
```diff
@@ -0,0 +1,116 @@
...
+  return {
+    bundleOrderSetting: data || {
+      key: BundleOrderSettingKey,
+      value: ruleBundles.map((ruleBundle) => ruleBundle?.id),
+    },
+    isFetching: isLoading,
+    fetchError: error,
+    refetch,
+  };
+};
```
gildub, line 116: We can replace with new `settings` query from https://github.com/konveyor/tackle2-ui/pull/641

client/src/app/pages/applications/analysis-wizard/__tests__/analysis-wizard.test.tsx
```diff
@@ -140,7 +140,8 @@ describe("<AnalysisWizard />", () => {
     await waitFor(() => expect(nextButton).toHaveAttribute("disabled", ""));
   });
 
-  it("can run analysis on applications with a binary definition using defaults", async () => {
+  //TODO
+  it.skip("can run analysis on applications with a binary definition using defaults", async () => {
```
gildub, line 144: What's the issue here ?

gildub (commented):
This looks good.
Please see inline for details.

client/src/app/pages/migration-targets/migration-targets.tsx
```diff
@@ -28,37 +28,74 @@ import { useTranslation } from "react-i18next";
...
+      promise
+        .then((response) => {
+          refreshBundleOrderSetting();
+        })
+        .catch((error) => {});
     }
   }
+  useEffect(() => {
+    refreshBundleOrderSetting();
+  }, [refreshBundleOrderSetting]);
```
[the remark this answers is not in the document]
ibolton336: Don't think the settings query will be able to be reused in its current state. Your settings query uses the  updateSetting 
axios call which converts the value of the request body to a string. That won't work for bundleOrder since we are storing it as an array of numbers.
``` 
export const updateSetting = (obj: Setting) =>
  axios.put<Setting>(`${SETTINGS}/${obj.key}`, obj.value?.toString());
```

client/src/app/queries/rulebundles.ts
ibolton336: See above comment

client/src/app/pages/applications/analysis-wizard/__tests__/analysis-wizard.test.tsx
ibolton336: This test relies on mock data that we are now pulling from the API. IMO we should leave the end to end stuff to the QE tests and use RTL for testing that individual hooks/components are working as expected.

client/src/app/api/models.tsx
```diff
@@ -419,19 +418,50 @@ export interface ITypeOptions {
...
+
+export interface IReadFile {
+  fileName: string;
+  loadError?: DOMException;
+  loadPercentage?: number;
+  loadResult?: "danger" | "success";
+  data?: string;
+  fullFile: File;
+  responseID?: number;
 }
```
gildub, line 467: Finally the IReadFile made it all the way up ;)

gildub (approved):
LGTM

client/src/app/common/CustomRules/add-custom-rules.tsx
```diff
@@ -29,11 +33,37 @@ export const AddCustomRules: React.FC<IAddCustomRules> = ({
...
+
+  const { mutate: uploadFile } = useUploadFileMutation(() => {}, onUploadError);
+
+  const onCreateRuleFileSuccess = (
+    response: any,
+    formData: FormData,
+    file: IReadFile
+  ) => {
+    //Set file ID for use in form submit
+    // setValue("imageID", response?.id);
```
mturley, line 53: Is this commented-out code important? We probably want to remove it and open an issue if necessary

client/src/app/common/CustomRules/add-custom-rules.tsx
```diff
@@ -29,11 +33,37 @@ export const AddCustomRules: React.FC<IAddCustomRules> = ({
...
+    const fileWithID: IReadFile = { ...file, ...{ responseID: response?.id } };
+    setReadFileData((prevReadFiles) => [...prevReadFiles, fileWithID]);
+  };
+
+  const onCreateRuleFileFailure = (error: AxiosError) => {
+    // resetField("imageID");
```
mturley, line 59: Same here, this function does nothing. Is it a TODO or unnecessary code?

client/src/app/pages/migration-targets/custom-target-form.tsx
```diff
@@ -56,142 +92,232 @@ export const CustomTargetForm: React.FC<CustomTargetFormProps> = ({
...
     defaultValues: {
-      name: target?.name || "",
-      description: target?.description || "",
-      id: target?.id || 0,
+      id: ruleBundle?.id || 0,
+      name: ruleBundle?.name || "",
+      description: ruleBundle?.description || "",
+      imageID: ruleBundle?.image.id || 0,
+      customRulesFiles: [],
+      // customRulesFiles: ruleBundle?.rulesets || [],
```
mturley, line 124: Commented-out code here too

client/src/app/pages/migration-targets/custom-target-form.tsx
```diff
@@ -56,142 +92,232 @@ export const CustomTargetForm: React.FC<CustomTargetFormProps> = ({
...
+  const onUpdateRuleBundleFailure = (error: AxiosError) => {};
+
+  const { mutate: updateRuleBundle } = useUpdateRuleBundleMutation(
+    onUpdateRuleBundleSuccess,
+    onUpdateRuleBundleFailure
+  );
+
+  //
+  const values = getValues();
+  console.log("values", values);
```
mturley, line 223: Console.log here that we may want to remove

client/src/app/pages/migration-targets/migration-targets.tsx
```diff
@@ -28,37 +28,158 @@ import { useTranslation } from "react-i18next";
...
+    bundleOrderSetting,
+    isFetching,
+    refetch: refreshBundleOrderSetting,
+  } = useFetchBundleOrder(ruleBundles);
 
-  const { migrationTargets } = useFetchMigrationTargets();
+  useEffect(() => {
+    refetchRuleBundles();
+    refreshBundleOrderSetting();
+  }, [isFetching]);
```
mturley, line 65: Hmm, @ibolton336 is it possible for this `useEffect` to cause an infinite loop? If I'm reading it correctly it'll run the effect whenever `isFetching` changes, which is triggered by `refreshBundleOrderSetting` in the effect. If it's not a loop it may be running more times than intended (since that refresh call will switch `isFetching` to true and then to false, so the effect will run twice). If this is not causing a loop I don't actually understand why 😅 some magic in react-query?

What's the intended trigger for this effect? If you simply want to fetch something else when the `refreshBundleOrderSetting` is done, you may be better off passing in an `onSuccess` to the query within `useFetchBundleOrder`.

client/src/app/queries/rulebundles.ts
```diff
@@ -0,0 +1,140 @@
...
+  const { data, isLoading, error, refetch } = useQuery<RuleBundle[]>(
+    [RuleBundlesQueryKey],
+    async () => await getRuleBundles(),
+    {
+      onError: (err) => console.log(err),
+    }
+  );
+  return {
+    ruleBundles: data || [],
+    isFetching: isLoading,
```
mturley, line 25: Ahh.. the reason the effect I mentioned above isn't looping is because we're really using the `isLoading` property from `useQuery` and not the `isFetching` property. `isLoading` will be true only during the first fetch, so that's why we don't get a loop. It does mean the effect is triggered twice though, is that intentional?

We probably want to be more explicit (fetch something only on mount with `[]` as the useEffect deps, then handle any cascading fetches as `onSuccess` or something like that. or if these things always need to be fetched in sequence, we could combine them into one `useQuery` and await one before fetching the other?) 

If not, we probably want to use the name `isLoading` instead of `isFetching` to avoid confusion with the `isFetching` from useQuery. Although if that's a pattern duplicated elsewhere, cleaning that up is probably out of scope here.

mturley (commented):
@ibolton336 I did a once-over here and found a few things that are odd, but it mostly looks good (didn't test, though)

ibolton336:
@mturley @gildub Fixed the issues you guys called out - Thanks for the reviews. Going to get this merged and follow up with smaller PRs for outstanding issues to prevent further bloat of this one.

PR #636, opened 2023-02-14
:bug: Update to UBI9 to fix arm64 builds

jmontleon:
Signed-off-by: Jason Montleon <jmontleo@redhat.com>

I know we weren't in a hurry to do this, however multi arch builds are experiencing odd failures on arm64 using UBI8.
https://github.com/konveyor/tackle2-ui/actions/runs/4168992831/jobs/7216425953#step:2:1067
https://github.com/konveyor/tackle2-ui/actions/runs/4174604421/jobs/7228394767#step:2:1077

I can reproduce much of this locally, and was able to workaround it using UBI9 node16 and node18 images without a problem both locally and within github actions:
https://github.com/jmontleon/tackle2-ui/actions/runs/4175222062/jobs/7229841318
https://quay.io/repository/jmontleon/tackle2-ui?tab=tags

I've opened an issue with qemu-user-static to see if there are any known issues we might be hitting, or other debugging  steps we can try to resolve the issue, but this is probably our best bet to move on at the moment.
https://github.com/multiarch/qemu-user-static/issues/185

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,12 +1,12 @@
 # Builder image
-FROM registry.access.redhat.com/ubi8/nodejs-16 as builder
+FROM registry.access.redhat.com/ubi9/nodejs-16 as builder
 USER 0
 COPY . .
 WORKDIR "/opt/app-root/src" 
 RUN npm install -w client && npm run build -w client && rm -rf node_modules && npm install -w server
 
 # Runner image
-FROM registry.access.redhat.com/ubi8/nodejs-16-minimal
+FROM registry.access.redhat.com/ubi9/nodejs-16-minimal
 
 # Add ps package to allow liveness probe for k8s cluster
 # Add tar package to allow copying files with kubectl scp


djzager (approved):

gildub:
Upgrade to UBI9 was first suggested [recently](https://github.com/konveyor/tackle2-ui/pull/579) and put on hold because the entire tackle2 stack needs to be aligned.

See https://github.com/konveyor/tackle2-ui/issues/638 for follow up.

gildub (dismissed):
@jmontleon, that's interesting issue which gives an opportunity to bump to latest UBI. 
Meanwhile we have to wait for the whole project to bump UBI along.

jmontleon:
> @jmontleon, that's interesting issue which gives an opportunity to bump to latest UBI. Meanwhile we have to wait for the whole project to bump UBI along.

- I don't agree with this. tackle2-addon-test-generator has been UBI 9 all along.
- tackle2-addon-windup had to be bumped in order to get a publicly available s390x svn package (CentOS 8 did not have an s390x release)
- in general for node and golang builds, all other things being equal, I don't view changing the distribution version as high risk. There's some, but not much.
- Downstream does not have to change with us. I get that the desire is to keep drift minimal, but I know for a fact it won't be the first time ever we've been on different releases upstream and down (ansible-runner upstream moved WAY faster than downstream, and then I/we missed a product rename that had us sitting longer on EL7 down while containers upstream were well and away on 8)

gildub (approved):

PR #637, opened 2023-02-14
:seedling: Add options for developing against a local hub instance

ibolton336:
Allow configuration for developing against a local hub instance

### DIFF
--- a/client/package.json
+++ b/client/package.json
@@ -12,9 +12,11 @@
     "port-forward:hub": "kubectl port-forward svc/tackle-hub -n konveyor-tackle 9002:8080",
     "port-forward:keycloak": "kubectl port-forward svc/tackle-keycloak-sso -n konveyor-tackle 9001:8080",
     "port-forward": "concurrently -n keycloak,hub \"npm:port-forward:keycloak\" \"npm:port-forward:hub\"",
+    "port-forward-local": "concurrently -n keycloak \"npm:port-forward:keycloak\"",
     "start:dev:server": "NODE_ENV=development node ../server/index.js",
     "start:dev:client": "GENERATE_SOURCEMAP=true webpack serve --config config/webpack.dev.ts",
     "start:dev": "concurrently -n port-forward,server,client -c \"white.bold.inverse,green.bold.inverse,blue.bold.inverse\" \"npm:port-forward\" \"npm:start:dev:server\" \"npm:start:dev:client\"",
+    "start:local": "concurrently -n port-forward-local,server,client -c \"white.bold.inverse,green.bold.inverse,blue.bold.inverse\" \"npm:port-forward-local\" \"npm:start:dev:server\" \"npm:start:dev:client\"",
     "test": "$(npm bin)/jest --rootDir=. --config=config/jest.config.js",
     "tsc": "../node_modules/typescript/bin/tsc -p ./tsconfig.json",
     "prepare": "cd .. && husky install client/.husky"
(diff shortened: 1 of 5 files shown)


gildub:
Why not just use the local 8080 port by stopping whatever is running there ?
BTW what's the :seedling: for ? It's nice but is not described by konveyor/release-tools.

ibolton336:
> Why not just use the local 8080 port by stopping whatever is running there ?

The point here is to allow the port # to be configurable rather than hard-coding it to be 8080. In my case, I was running a local hub instance at port 8080 (the default). I don't see any harm in allowing the port to be set via env var on the UI side. Any objection to that?

gildub:
@ibolton336, the objection I see is that we are making a special case that is creating unnecessary complexity.
The rationale for the tackle environment is that we run against a cluster or a local environment. In either case the hub runs on 8080. This is the standard of the cluster. Anyone can run the hub locally that way.

ibolton336:
@gildub The hub runs locally on 8080. So our default setup results in the port for our express server `8080` and the port for the hub `8080`.  Adding the ability to set the port that our app runs on should be standard practice IMO. Sure, we can change the port that the hub runs on locally.

gildub:
@ibolton336, I understand the idea of having the option to change the port for the hub.
What bothers me is that it makes more complicated to run the environment locally especially knowing that the UI is part of a cluster and that we have standards for ports and all is needed is to have those services running where expected. So we don't have to bend in all directions.

ibolton336:
@gildub You don't have to change anything if you are using the hub service which is port forwarded from the cluster onto your local machine. The default is still 8080.

gildub:
By change I mean we need to add those targets to npm scripts:
- "port-forward-local"
- "start:local"

client/package.json
```diff
@@ -12,9 +12,11 @@
     "port-forward:hub": "kubectl port-forward svc/tackle-hub -n konveyor-tackle 9002:8080",
     "port-forward:keycloak": "kubectl port-forward svc/tackle-keycloak-sso -n konveyor-tackle 9001:8080",
     "port-forward": "concurrently -n keycloak,hub \"npm:port-forward:keycloak\" \"npm:port-forward:hub\"",
+    "port-forward-local": "concurrently -n keycloak \"npm:port-forward:keycloak\"",
```
gildub, line 15: I don't think we need this one because we already have `port-forward:keycloak`.

client/package.json
```diff
@@ -12,9 +12,11 @@
...
     "start:dev:server": "NODE_ENV=development node ../server/index.js",
     "start:dev:client": "GENERATE_SOURCEMAP=true webpack serve --config config/webpack.dev.ts",
     "start:dev": "concurrently -n port-forward,server,client -c \"white.bold.inverse,green.bold.inverse,blue.bold.inverse\" \"npm:port-forward\" \"npm:start:dev:server\" \"npm:start:dev:client\"",
+    "start:local": "concurrently -n port-forward-local,server,client -c \"white.bold.inverse,green.bold.inverse,blue.bold.inverse\" \"npm:port-forward-local\" \"npm:start:dev:server\" \"npm:start:dev:client\"",
```
gildub, line 19: We only need to add that line:
`"start:dev:local": "concurrently -n port-forward,server,client -c "white.bold.inverse,green.bold.inverse,blue.bold.inverse\" \"npm:port-forward:keycloak\" "npm:start:dev:server\" \"npm:start:dev:client\"",`

gildub (changes requested):
This could be simplified.

PR #641, opened 2023-02-16
:sparkle: General page

gildub:


client/src/app/api/rest.tsx
```diff
@@ -568,3 +553,12 @@ export const createMigrationTarget = (obj: MigrationTarget) =>
 
 export const deleteMigrationTarget = (id: number): AxiosPromise =>
   axios.delete(`${RULESETS}/${id}`);
+
+export const getSettingById = (key: SettingKey) =>
+  axios.get<boolean>(`${SETTINGS}/${key}`).then((response) => response.data);
+
+export const updateSetting = (obj: Setting) =>
+  axios.put<Setting>(`${SETTINGS}/${obj.key}`, obj.value?.toString());
```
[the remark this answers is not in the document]
gildub: Right, `updateSetting` can now take any value.

ibolton336:
@gildub Is the api side implemented for this? Seeing:

``` 
PUT http://localhost:9000/hub/settings/download.html.enabled 400 (Bad Request)
```

ibolton336 (changes requested):
This is currently broken on main 

```
{error: "EOF"}
error: "EOF"
``` 
response for review.assessment.required & download.csv.enabled PUT requests

gildub:
@ibolton336, I'm testing against konveyor/tackle2-hub main branch : 

```
$   curl -s http://localhost:9000/hub/settings/download.html.enabled | jq
true
```

gildub:
@ibolton336 , the PUT request requires a data field.

There was an issue with the payload.
Fixed now.

client/src/app/api/rest.ts
```diff
@@ -592,3 +574,12 @@ export const createFile = ({
...
       return response.data;
     });
+
+export const getSettingById = <T>(key: string): Promise<T> =>
+  axios.get(`${SETTINGS}/${key}`).then((response) => response.data);
+
+export const updateSetting = (obj: Setting): Promise<Setting> =>
+  axios.put(
+    `${SETTINGS}/${obj.key}`,
+    typeof obj.value == "boolean" ? obj.value.toString() : obj.value
```
mturley, line 584: We should use `===` here for consistency

client/src/app/pages/applications/analysis-wizard/set-targets.tsx
```diff
@@ -27,11 +25,7 @@ export const SetTargets: React.FC = () => {
     refetch: refetchRuleBundles,
   } = useFetchRuleBundles();
 
-  const {
-    bundleOrderSetting,
-    isFetching,
-    refetch: refreshBundleOrderSetting,
-  } = useFetchBundleOrder(ruleBundles);
+  const bundleOrderSetting = useSetting<number[]>("ui.bundle.order");
```
mturley, line 28: I wonder if there is a way to have the `number[]` here automatically inferred by the `"ui.bundle.order"` you're passing in, since we already know the mapping of setting keys to setting value types. Hmm...

client/src/app/queries/settings.ts
```diff
@@ -0,0 +1,41 @@
...
+      queryFn: () => getSettingById<T>(key),
+      onError: (error) => console.log(error),
+    });
+  }
+  return useQuery({
+    queryKey: [SettingQueryKey, key],
+    queryFn: () => getSettingById<T>(key),
+    onError: (error) => console.log(error),
+  });
+};
```
mturley, line 27: Both sides of this `if` are the same, and the `queryClient` is never used?

mturley (commented):
Ok @gildub , bear with me here. I think I have a solution to infer the type of a setting value from the setting key all the way down to the `useSetting` call without having to explicitly pass a type param like `useSetting<boolean>(...)` at all, and it doesn't use union types either. Maybe give this a try:

```ts
export type SettingTypesByKey = {
  "download.csv.enabled": boolean;
  "download.html.enabled": boolean;
  "git.insecure.enabled": boolean;
  "mvn.dependencies.update.forced": boolean;
  "mvn.insecure.enabled": boolean;
  "review.assessment.required": boolean;
  "svn.insecure.enabled": boolean;
  "ui.bundle.order": number[];
};

export type Setting<T extends keyof SettingTypesByKey> = {
  key: T;
  value: SettingTypesByKey[T];
};
```

Testing out the types here, check this out:
```ts
const goodSetting1: Setting<"download.csv.enabled"> = {
  key: "download.csv.enabled",
  value: true, // No error here!
};

const badSetting1: Setting<"download.csv.enabled"> = {
  key: "download.csv.enabled",
  value: [1], // TypeError: Type 'number[]' is not assignable to type 'boolean' !
};

const goodSetting2: Setting<"ui.bundle.order"> = {
  key: "ui.bundle.order",
  value: [1], // No error here!
};

const badSetting2: Setting<"ui.bundle.order"> = {
  key: "ui.bundle.order",
  value: true, // TypeError: Type 'boolean' is not assignable to type 'number[]' !
};
```

Okay, so we know it can check the `value` type if we explicitly give it a key as a type param. Can we infer the other way? Let's try this:

```ts
export const getSettingById = <T extends keyof SettingTypesByKey>(id: T) => {
  return axios
    .get<Setting<T>>(`${SETTINGS}/${id}`, jsonHeaders)
    .then((res) => res.data);
};
```

Now what if we call that without passing a type param:
```ts
const csvEnabledSetting = getSettingById("download.csv.enabled"); // This is inferred as Promise<Setting<"download.csv.enabled">>
const csvEnabled = (await csvEnabledSetting).value; // This is inferred as boolean!

const uiBundleOrderSetting = getSettingById("ui.bundle.order"); // This is inferred as Promise<Setting<"ui.bundle.order">>
const uiBundleOrder = (await uiBundleOrderSetting).value; // This is inferred as number[]!
```

Perfect! What about `useSetting`?

```ts
export const useSetting = <T extends keyof SettingTypesByKey>(key: T) => {
  return useQuery({
    queryKey: [SettingQueryKey, key],
    queryFn: () => getSettingById<T>(key),
    onError: (error) => console.log(error),
  });
};
```

Check it out:

```ts
const csvEnabledSettingQuery = useSetting("download.csv.enabled"); // This is inferred as UseQueryResult<Setting<"download.csv.enabled">!
const csvEnabled = csvEnabledSettingQuery.data?.value; // This is inferred as boolean | undefined!

const uiBundleOrderQuery = useSetting("ui.bundle.order"); // This is inferred as UseQueryResult<Setting<"ui.bundle.order">!
const uiBundleOrder = uiBundleOrderQuery.data?.value; // This is inferred as number[] | undefined!
```

Everything can be inferred from the key without any explicit repetition of the value's type!

client/src/app/queries/settings.ts
```diff
@@ -0,0 +1,41 @@
...
+  return useMutation({
+    mutationFn: (obj: Setting) => updateSetting(obj),
+    onSuccess: () => {
+      queryClient.invalidateQueries([SettingQueryKey]);
+    },
+    onError: () => {
+      queryClient.invalidateQueries([SettingQueryKey]);
+    },
+  });
+};
```
mturley, line 41: Looking a little further, now I'm wondering how would this affect `useSettingMutation`?

client/src/app/queries/settings.ts
gildub: @mturley, not the same since we have a return statement either way.
Nice catch re `queryClient`, we don't need it here.

mturley (commented):
I think in places where you just have the bare `Setting` type now like `updateSetting`, you could copy the same generics I describe above or you could always just use `Setting<any>` in a pinch. But to get the full value out of it I think we would need to then enhance `updateSetting` and `useSettingMutation` with generics as well:

```ts
export const updateSetting = <K extends keyof SettingTypesByKey>(obj: Setting<K>): AxiosPromise<Setting<K>> => {
  return APIClient.put(
    `${SETTINGS}/${obj.key}`,
    obj.value?.toString(),
    jsonHeaders
  );
};

export const useSettingMutation = () => {

  return useMutation({
    mutationFn: (obj: Setting<keyof SettingTypesByKey>) => updateSetting(obj),
    onSuccess: () => {
      queryClient.invalidateQueries([SettingQueryKey]);
    },
    onError: () => {
      queryClient.invalidateQueries([SettingQueryKey]);
    },
  });
};
```

There's a problem here though: because you are currently using one `useSettingMutation` instance and calling `.mutate` with different settings, we have to use the widened type `Setting<keyof SettingTypesByKey>` which I think will allow any key and value and not check them against each other. This is super weird:

```ts
const badSettingButNoError: Setting<keyof SettingTypesByKey> = {
  key: "download.csv.enabled",
  value: [1], // Hmm... no error here
};
```

I'm not sure if/how we can create a `AnySetting` type that will still check the value without having to lift the generic key type out.

One way to avoid this problem would be to split up your `useSettingMutation`s so that you have one per setting key. You could write it like this:

```ts
export const useSettingMutation = <K extends keyof SettingTypesByKey>(
  key: K
) => {

  return useMutation({
    mutationFn: (value: SettingTypesByKey[K]) => updateSetting({ key, value }),
    onSuccess: () => {
      queryClient.invalidateQueries([SettingQueryKey]);
    },
    onError: () => {
      queryClient.invalidateQueries([SettingQueryKey]);
    },
  });
};
```

But then you'd have to use it like this, separately per setting:

```ts
// Instead of this:
const settingMutationQuery = useSettingMutation();

// You'd do this:
const reviewAssessmentSettingMutation = useSettingMutation("review.assessment.required");
const downloadHTMLSettingMutation = useSettingMutation("download.html.enabled");

// Instead of this:
  const onChangeReviewAssessmentSetting = () => {
    if (reviewAssessmentSetting.isSuccess) {
      const setting: FlagSetting = {
        key: "review.assessment.required",
        value: !reviewAssessmentSetting.data,
      };
      settingMutationQuery.mutate(setting);
    }
  };
  
// You'd do this:
  const onChangeReviewAssessmentSetting = () => {
    if (reviewAssessmentSetting.isSuccess) {
      reviewAssessmentSettingMutation.mutate(!reviewAssessmentSetting.data);
    }
  };

// Etc, same for the other settings
```

There would probably be other ripple effects to resolve too. But this should mean you always have a strict type inferred for every setting value (including when you call those `mutate` functions).

client/src/app/queries/settings.ts
mturley: > not the same since we have a return statement either way.
I don't understand what you mean. How would this code behave any differently if you removed the `if (key === "ui.bundle.order")` and just left one copy of the `return useQuery({` block? I don't see any differences between these two `useQuery` calls.

gildub:
Thank you @mturley, this is really generic now!

ibolton336 (changes requested):
For general settings, I think you'll need to update the csv setting to account for this https://github.com/konveyor/tackle2-addon-windup/issues/80#issue-1584447341
Also, can we get rid of the loading screen flicker for the general settings selections & have them behave  like the other settings sliders?

gildub:
I added `<Alert/>` when there is an error. 
Meanwhile we're loosing information by removing the `isLoading` case because when the network is slow or broken the page keeps loading with a blank page. I think it makes sense to keep it and the other setting sliders could benefit from it too. What do you think ?

Adding CSV setting for windup configuration into the analysis wizard in follow-up PR.

ibolton336:
> I added `<Alert/>` when there is an error. 
> Meanwhile we're loosing information by removing the `isLoading` case because when the network is slow or broken the page keeps loading with a blank page. I think it makes sense to keep it and the other setting sliders could benefit from it too. What do you think ?
> 
> Adding CSV setting for windup configuration into the analysis wizard in follow-up PR.

Loading state sounds good. But let's think of a less jarring way to express it than having the entire component state flicker with a loading screen that usually lasts less than a second.

gildub:
@ibolton336, I agree, I removed the loading part for now so we can move forward.
Let's find a more granular way to add "loading" for the settings components.

ibolton336 (approved):

PR #643, opened 2023-02-16
:sparkles: [WIP] Applications: SetQueryData for optimization.

gildub:
setQueryData allows to immediately update a query's cached data after mutation

### DIFF
--- a/client/src/app/queries/applications.ts
+++ b/client/src/app/queries/applications.ts
@@ -53,6 +53,7 @@ export const useUpdateApplicationMutation = (
   const { isLoading, mutate, error } = useMutation(updateApplication, {
     onSuccess: (res) => {
       onSuccess(res);
+      queryClient.setQueryData([ApplicationsQueryKey, res.data.id], res);
       queryClient.invalidateQueries([ApplicationsQueryKey]);
     },
     onError: (err: AxiosError) => {


client/src/app/queries/applications.ts
```diff
@@ -53,6 +53,7 @@ export const useUpdateApplicationMutation = (
   const { isLoading, mutate, error } = useMutation(updateApplication, {
     onSuccess: (res) => {
       onSuccess(res);
+      queryClient.setQueryData([ApplicationsQueryKey, res.data.id], res);
```
ibolton336, line 56: What does this do? Seems like this is something that would be better handled by our existing patterns (or by react query internally) rather than adding a one off implementation here.

ibolton336 (changes requested):

client/src/app/queries/applications.ts
gildub: setQueryData updates the applications list we are currently on instantly.
This allows for avoiding waiting for a refresh after creating/editing.
As for patterns, we use react query individually per API resource which makes it difficult for testing reaons to add such optimization across all queries but why not.

ibolton336: For this one, aren't we already invalidating the applications fetch query so we are going to pull in new data asynchronously anyway?

The response from an application update PUT request doesn't actually have the ID defined here
image

ibolton336 (changes requested):

Issue #607, opened 2023-02-03
API requests need to be consistent


gildub:
In react query we have a mix of following approaches.

```
export const updateReview = (obj: Review): AxiosPromise<Review> => {
  return APIClient.put(`${REVIEWS}/${obj.id}`, obj);
};
```

```
export const updateApplicationQuery = (obj: Application) =>
  axios.put<Application>(`${APPLICATIONS}/${obj.id}`, obj);
```

We are using both approaches because the legacy application was making use of the APIClient class but we since we introduce React Query we introduced a direct Axios requests.

This is a tech debt.

We need to be consistent across and use the Axios direct approach everywhere.

Note: To avoid confusion between the two approaches the second type of queries names are suffixed with "Query". 
Can be removed once reduced.

gildub:
I have a PR in progress (not pushed yet).

PR #644, opened 2023-02-17
:sparkles: Applications Queries to use Axios

gildub:
https://github.com/konveyor/tackle2-ui/issues/607

### DIFF
--- a/client/src/app/pages/applications/components/application-form/application-form.tsx
+++ b/client/src/app/pages/applications/components/application-form/application-form.tsx
@@ -18,8 +18,6 @@ import {
 } from "@patternfly/react-core";
 import QuestionCircleIcon from "@patternfly/react-icons/dist/js/icons/question-circle-icon";
 
-import spacing from "@patternfly/react-styles/css/utilities/Spacing/spacing";
-
 import {
   SingleSelectFetchOptionValueFormikField,
   MultiSelectFetchOptionValueFormikField,
@@ -52,6 +50,7 @@ import {
 import "./application-form.css";
 import { useFetchBusinessServices } from "@app/queries/businessservices";
 import { useFetchTagCategories } from "@app/queries/tags";
+
 export interface FormValues {
   name: string;
   description: string;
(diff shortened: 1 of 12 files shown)


ibolton336:
No need to assert the AxiosError type. We are using Axios so the error will always be returned as an AxiosError type from axios.

client/src/app/pages/applications/components/application-form/application-form.tsx
```diff
@@ -316,8 +316,8 @@ export const ApplicationForm: React.FC<ApplicationFormProps> = ({
     onSaved(response);
   };
 
-  const onCreateUpdateApplicationError = (error: AxiosError) => {
-    setAxiosError(error);
+  const onCreateUpdateApplicationError = (error: unknown) => {
+    if (error instanceof Error) setError(error);
```
ibolton336, line 320: What is going on here?

client/src/app/api/rest.tsx
```diff
@@ -478,7 +478,17 @@ export const deleteBulkApplications = (ids: number[]) =>
...
-  axios.get<Application[]>(APPLICATIONS).then((response) => response.data);
+  axios
+    .get<Application[]>(APPLICATIONS)
+    .then((response) => response.data)
+    .catch((error: Error | AxiosError) => {
+      if (axios.isAxiosError(error)) {
+        return error;
+      } else {
+        console.log(`${error}`);
+        return;
```
ibolton336, line 489: We should still be throwing the error so that it can be picked up by our errorBoundary component. Also - We should be implementing this global error handling for non axios errors at the global axios interceptor level rather than as a one off implementation for applications.

ibolton336 (changes requested):

client/src/app/pages/applications/components/application-form/application-form.tsx
gildub: I restored AxiosError for now until we decide cleary on how we're going to handle/filter fetch errors.

gildub:
Closing as files were not syncing with my branch.

Issue #651, opened 2023-02-20
CSV download


ibolton336:
https://issues.redhat.com/browse/MTA-116

PR #653, opened 2023-02-21
:sparkles: Add download buttons for report

### LINKED ISSUE #613: Download CSV
https://issues.redhat.com/browse/MTA-116

ibolton336:
Closes #613 #651

### DIFF
--- a/client/src/app/pages/applications/analysis-wizard/analysis-wizard.tsx
+++ b/client/src/app/pages/applications/analysis-wizard/analysis-wizard.tsx
@@ -36,6 +36,7 @@ import {
 } from "./schema";
 import { useAsyncYupValidation } from "@app/shared/hooks/useAsyncYupValidation";
 import { CustomRules } from "./custom-rules";
+import { useSetting } from "@app/queries/settings";
 
 interface IAnalysisWizard {
   applications: Application[];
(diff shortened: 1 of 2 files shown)


client/src/app/pages/applications/components/application-list-expanded-area/application-list-expanded-area-analysis.tsx
```diff
@@ -118,6 +118,21 @@ export const ApplicationListExpandedAreaAnalysis: React.FC<
             "Not available"
           )}
         </DescriptionListDescription>
+        {task?.state === "Succeeded" && (
+          <DescriptionListDescription>
+            <Tooltip content="Click to download Analysis report">
+              <Button variant="link" isInline>
+                <Link
+                  to={`/hub/applications/${application.id}/bucket${task?.data?.output}`}
```
gildub, line 126: We probably want to use the defined path APPLICATIONS here.

gildub (commented):
Don't we need to test for global settings allowing CSV download ?

client/src/app/pages/applications/components/application-list-expanded-area/application-list-expanded-area-analysis.tsx
ibolton336: Good idea :+1:

gildub (commented):
It's hard to visually distinct the html link from the csv one. 
Not sure what are the requirements meanwhile it might worth adding a comma or equivalent to clearly separate them ?

CSV link return html file.

I18n is missing.

gildub (approved):
I'm still facing a html file instead of tar file for csv but that's no blocker so we can address it later.

PR #657, opened 2023-02-22
:bug: Fix empty assessments query error in console

ibolton336:
- Assessments query needs a default return value when no assessments exist for an app.


gildub (approved):
LGTM

PR #658, opened 2023-02-22
:bug: Fix regression with form validation messages not being shown 

ibolton336:
- Adds onBlur method to  hook-form-pf-group-controller component to fix missing validation
- Adds missing xmlValidation prop for edit identity form


gildub (approved):
Nice !

PR #659, opened 2023-02-23
:bug: Aggregate target sources

ibolton336:
- Aggregates all sources and targets from selected rule bundles and sets the source/target field values accordingly. 
- This required the cardSelected check to change from the target values to the ruleBundle values to allow multiple targets to be assigned from a single rulebundle (in the case of a custom target). 
- Fixes a bug when the formTargets were not being updated when a selected card's target dropdown value changes. Before the targets were only updated on card click.

### DIFF
--- a/client/src/app/pages/applications/analysis-wizard/set-targets.tsx
+++ b/client/src/app/pages/applications/analysis-wizard/set-targets.tsx
@@ -36,12 +36,42 @@ export const SetTargets: React.FC = () => {
   const { watch, setValue } = useFormContext<AnalysisWizardFormValues>();
   const formTargets = watch("formTargets");
   const formRuleBundles = watch("formRuleBundles");
+  const formSources = watch("formSources");
+
+  const handleOnSelectedCardTargetChange = (
+    selectedRuleTarget: string,
+    selectedRuleBundle: RuleBundle
+  ) => {
+    const otherSelectedRuleTargets = formTargets.filter(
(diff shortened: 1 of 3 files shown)


gildub (commented):
Why do we need to filter out source cards from withing set-targets.tsx ?
Can't we passed already filtered targets to the component ?

client/src/app/pages/applications/analysis-wizard/set-targets.tsx
```diff
@@ -36,12 +36,42 @@ export const SetTargets: React.FC = () => {
...
     isSelecting: boolean,
     selectedRuleTarget: string,
     selectedRuleBundle: RuleBundle
   ) => {
+    const otherSelectedRuleSources = formSources.filter(
+      (formSource) =>
+        !selectedRuleBundle.rulesets
+          .map((rule) => rule.metadata.source)
+          .includes(formSource)
+    );
```
gildub, line 74: We might not have much choices, but I wonder if we could avoid dealing with sources here.

gildub: My concern is from a component viewpoint.
Set-targets need to know only about targets.
It would make sense to fetch ruleBundles in the wizard component and pass only the subset of targets to set-targets.tsx.

gildub (approved):
Regarding my concern, we can follow-up in a separate PR.

PR #662, opened 2023-02-27
:sparkles: Scroll to bottom notification link

ibolton336:
- Scroll to the bottom of the custom targets page when "Take me there" link is clicked on customTarget create

### DIFF
--- a/client/src/app/pages/migration-targets/migration-targets.tsx
+++ b/client/src/app/pages/migration-targets/migration-targets.tsx
@@ -1,4 +1,4 @@
-import React, { useEffect, useState } from "react";
+import React, { useEffect, useRef, useState } from "react";
 import {
   DndContext,
   closestCenter,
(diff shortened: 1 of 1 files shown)


client/src/app/pages/migration-targets/migration-targets.tsx
```diff
@@ -59,6 +59,14 @@ export const MigrationTargets: React.FC = () => {
     refetch: refreshBundleOrderSetting,
   } = useFetchBundleOrder(ruleBundles);
 
+  // Sroll to bottom on new custom target
```
gildub, line 62: This comment seems superfluous

client/src/app/pages/migration-targets/migration-targets.tsx
```diff
@@ -59,6 +59,14 @@ export const MigrationTargets: React.FC = () => {
...
+  const targetsEndRef = useRef<null | HTMLDivElement>(null);
+
+  const scrollToBottom = () => {
+    targetsEndRef.current?.scrollIntoView({ behavior: "smooth" });
+  };
+  //
```
gildub, line 68: Not needed

gildub (changes requested):
Please see inline

client/src/app/pages/migration-targets/migration-targets.tsx
```diff
@@ -108,6 +114,15 @@ export const MigrationTargets: React.FC = () => {
           type: "custom target",
         }),
         variant: "success",
+        message: (
+          <Button
+            variant="link"
+            onClick={scrollToBottom}
+            className={spacing.pl_0}
+          >
+            Take me there
```
gildub, line 123: What about i18n ?

gildub (approved):
I think i18n should be maintained until fully dropped, let's hope the feature won't be reactivated.

PR #668, opened 2023-02-27
:bug: change isLoading->isInitialLoading

### LINKED ISSUE #667: v4 react-query breaking change: isLoading -> isInitialLoading
https://tanstack.com/query/v4/docs/react/guides/migrating-to-react-query-4#disabled-queries

djzager:
v4 react-query had a breaking change where the idle state has been removed making it so that `isLoading` can be set to true when there is nothing to load.

https://tanstack.com/query/v4/docs/react/guides/migrating-to-react-query-4#disabled-queries

We can use `isInitialLoading` to know whether or not we should display the loading spinner.

Fixes #667


ibolton336 (approved):
Nice catch! :+1:

Issue #671, opened 2023-02-27
Maven configuration: Warning to resolve


gildub:
We get following warning when mounting http://localhost:9000/repositories/maven
```
Warning: A component is changing an uncontrolled input to be controlled. This is likely caused by the value changing from undefined to a defined value, which should not happen. Decide between using a controlled or uncontrolled input element for the lifetime of the component. More info: https://reactjs.org/link/controlled-components
```

PR #678, opened 2023-03-02
:bug: mvn: Refactor without useEffect

gildub:
Resolves https://github.com/konveyor/tackle2-ui/issues/671

### DIFF
--- a/client/src/app/pages/repositories/Mvn.tsx
+++ b/client/src/app/pages/repositories/Mvn.tsx
@@ -49,7 +49,7 @@ export const RepositoriesMvn: React.FC = () => {
   // };
 
   let storageValue: string = "";
-  const { data: cache, isLoading, isSuccess, refetch } = useFetchCache();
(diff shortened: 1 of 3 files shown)


client/src/app/pages/repositories/Mvn.tsx
```diff
@@ -49,20 +49,18 @@ export const RepositoriesMvn: React.FC = () => {
   //     mvnForcedSettingMutation.mutate(!mvnForcedSetting.data);
   // };
 
-  const [storageValue, setStorageValue] = useState<string>();
+  let storageValue: string = "";
+  const cache = useFetchCache();
```
ibolton336, line 53: What are your thoughts on destructuring the isSuccess & refetch values/methods here? 

Looking good!

gildub: It's about "Type Narrowing", by keeping the context it help TS.
And it makes consuming requests easier since we don't have to rename anything.

ibolton336 (approved):

Issue #612, opened 2023-02-07
Review without assessment


ibolton336:
https://issues.redhat.com/browse/MTA-115
https://github.com/konveyor/enhancements/issues/88

PR #683, opened 2023-03-03
:sparkles: Allow reviews without assessment when flagged

gildub:
Resolves https://github.com/konveyor/tackle2-ui/issues/612

### DIFF
--- a/client/src/app/pages/applications/applications-table-assessment/applications-table-assessment.tsx
+++ b/client/src/app/pages/applications/applications-table-assessment/applications-table-assessment.tsx
@@ -569,7 +569,10 @@ export const ApplicationsTable: React.FC = () => {
     );
   };
 
-  const reviewAssessmentSetting = useSetting("review.assessment.required");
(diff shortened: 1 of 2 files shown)


client/src/app/pages/applications/applications-table-assessment/applications-table-assessment.tsx
```diff
@@ -580,7 +582,7 @@ export const ApplicationsTable: React.FC = () => {
 
     const row = selectedRows[0];
     const assessment = getApplicationAssessment(row.id!);
-    if (!assessment) {
+    if (!assessment && !reviewAssessmentSetting.data) {
```
ibolton336, line 585: Could we destructure here and alias the setting to a name that would be self documenting?

Something like:

``` 
  const {data: requireAssessBeforeReview} = useSetting("review.assessment.required");
```
or something to that effect.

ibolton336 (changes requested):

ibolton336 (approved):

PR #689, opened 2023-03-06
:bug: Update tags model to include source for tags

ibolton336:
- Removes unused "category" field in applications form tag dropdown 
- Add 'source' field to tagRef so that it will not be overwritten on application edit 
- Fixes the issues we were seeing around duplicate tags being created with automated tags turned on during subsequent analysis attempts


gildub (approved):
LGTM

ibolton336:
Closes #688

PR #691, opened 2023-03-06
:bug: Fixes preliminary bugs identified by QE

ibolton336:
- Strongly type custom rule files value in analysis wizard form. This allows us to use the form values as the table state rather than creating new state to temporarily hold the table values. This simplifies the custom rule file table data in wizard.

Fixes preliminary bugs identified by QE: 
- In the Create wizard: there is no warning about custom target's name if that name is already exists -> the Create button is enabled but does nothing

- In the Create wizard: Title is: "New custom Target" - custom starts with lowercase letter

- Analysis wizard: Add manual rule -> press Next -> press Back => "No custom rules available" and when clicking Add rules there is an error about file is already uploaded


gildub (approved):
LGTM

PR #694, opened 2023-03-07
:bug: Fix firefox html/csv report download

ibolton336:
Currently, the report download feature does not work in firefox. This fixes the issue by ensuring the accept headers are cleared when requesting a file download. The 'download' html attribute handles this in chrome, but fails to work in firefox.


gildub (approved):
LGTM

PR #700, opened 2023-03-09
:bug: Mirror rule bundle target metadata selection in advanced options

ibolton336:
- Currently, advanced options does not reflect the targets selected in the set targets step within the wizard. This fixes that and allows the selection of rule bundle associated targets and non-associated targets alike from the same select dropdown. This is necessary because there are some targets available that are not tied to rulebundles. This functionality was inherited from the windup web console. 
- Fixes a bug when a user selects a target from a "category: "kind" rulebundle in the set-target dropdown, navigates through the wizard, and then navigates back to the set-targets page to find their selection did not persist. This PR adds a check for previously selected targets. 

- Adds typescript support for the RuleBundle form data.

### DIFF
--- a/client/src/app/common/CustomRules/rules-utils.tsx
+++ b/client/src/app/common/CustomRules/rules-utils.tsx
@@ -47,7 +47,6 @@ export const parseRules = (file: IReadFile): ParsedRule => {
 
 export const getruleBundleTargetList = (ruleBundle: RuleBundle) => {
   return ruleBundle.rulesets.reduce((acc: string[], ruleset) => {
-    acc.push(ruleset?.metadata?.target);
-    return acc;
+    return [...acc, ruleset?.metadata?.target];
   }, []);
 };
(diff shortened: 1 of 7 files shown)


client/src/app/common/CustomRules/rules-utils.tsx
```diff
@@ -44,3 +44,10 @@ export const parseRules = (file: IReadFile): ParsedRule => {
     };
   }
 };
+
+export const getruleBundleTargetList = (ruleBundle: RuleBundle) => {
+  return ruleBundle.rulesets.reduce((acc: string[], ruleset) => {
+    acc.push(ruleset?.metadata?.target);
+    return acc;
```
mturley, line 51: I think this is probably okay, but technically I don't think it's guaranteed to be safe to mutate the accumulator in-place in a reduce like this. You could do it immutably with a spread operator:
```ts
return [...acc, ruleset?.metadata?.target];
```

mturley (approved):
LGTM! One minor nitpick, you can take it or leave it.

mturley (approved):

PR #704, opened 2023-03-13
:bug: Fix initial loading state for existing rule file placeholders

ibolton336:
Fixes https://issues.redhat.com/browse/MTA-300

### DIFF
--- a/client/src/app/api/models.ts
+++ b/client/src/app/api/models.ts
@@ -463,10 +463,17 @@ export interface ParsedRule {
   fileID?: number;
 }
 
+export type FileLoadError = {
+  name?: string;
+  message?: string;
+  stack?: string;
+  cause?: {};
+};
+
 export interface IReadFile {
   fileName: string;
   fullFile?: File;
-  loadError?: DOMException;
+  loadError?: FileLoadError;
   loadPercentage?: number;
   loadResult?: "danger" | "success";
   data?: string;
--- a/client/src/app/common/CustomRules/useRuleFiles.tsx
+++ b/client/src/app/common/CustomRules/useRuleFiles.tsx
@@ -1,7 +1,7 @@
 import { useContext, useState } from "react";
 import XSDSchema from "./windup-jboss-ruleset.xsd";
 import { XMLValidator } from "fast-xml-parser";
-import { IReadFile } from "@app/api/models";
+import { FileLoadError, IReadFile } from "@app/api/models";
 import { NotificationsContext } from "@app/shared/notifications-context";
 import { AxiosError } from "axios";
 import { useUploadFileMutation } from "@app/queries/taskgroups";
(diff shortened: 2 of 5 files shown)


client/src/app/api/models.ts
```diff
@@ -464,7 +464,7 @@ export interface ParsedRule {
 export interface IReadFile {
   fileName: string;
   fullFile?: File;
-  loadError?: DOMException;
```
mturley, line 467: @ibolton336 what was the type error here solved by changing this to `any`? Could we maybe find a type that is more permissive like `Error` or `Exception` that TS would be happy with?

mturley (commented):
Minor nitpick, LGTM otherwise

client/src/app/api/models.ts
ibolton336: Mapping the yup validation form value types to the interfaces we have has been slightly painful. I am trying to map the schema to the readfile like so:


```
export const customRulesFilesSchema: yup.SchemaOf<IReadFile> = yup.object({
  fileName: yup.string().required(),
  fullFile: yup.mixed<File>(),
  loadError: yup.mixed<Error>(),
  loadPercentage: yup.number(),
  loadResult: yup.mixed<"danger" | "success" | undefined>(),
  data: yup.string(),
  responseID: yup.number(),
});
```
```
export interface IReadFile {
  fileName: string;
  fullFile?: File;
  loadError?: Error;
  loadPercentage?: number;
  loadResult?: "danger" | "success";
  data?: string;
  responseID?: number;
}
```

But I am seeing some TS errors in the defaultValue props passed to the useForm 

```(property) customRulesFiles?: ({
    fileName?: string | undefined;
    fullFile?: File | undefined;
    loadError?: {
        name?: string | undefined;
        message?: string | undefined;
        stack?: string | undefined;
        cause?: {} | undefined;
    } | undefined;
    loadPercentage?: number | undefined;
    loadResult?: "danger" | ... 1 more ... | undefined;
    data?: string | undefined;
    responseID?: number | undefined;
} | undefined)[] | undefined
Type 'IReadFile[]' is not assignable to type '({ fileName?: string | undefined; fullFile?: File | undefined; loadError?: { name?: string | undefined; message?: string | undefined; stack?: string | undefined; cause?: {} | undefined; } | undefined; loadPercentage?: number | undefined; loadResult?: "danger" | ... 1 more ... | undefined; data?: string | undefined; ...'.
  Type 'IReadFile' is not assignable to type '{ fileName?: string | undefined; fullFile?: File | undefined; loadError?: { name?: string | undefined; message?: string | undefined; stack?: string | undefined; cause?: {} | undefined; } | undefined; loadPercentage?: number | undefined; loadResult?: "danger" | ... 1 more ... | undefined; data?: string | undefined; r...'.
    Types of property 'loadError' are incompatible.
      Type 'Error | undefined' is not assignable to type '{ name?: string | undefined; message?: string | undefined; stack?: string | undefined; cause?: {} | undefined; } | undefined'.
        Type 'Error' is not assignable to type '{ name?: string | undefined; message?: string | undefined; stack?: string | undefined; cause?: {} | undefined; }'.
          Types of property 'cause' are incompatible.
            Type 'unknown' is not assignable to type '{} | undefined'.ts(2322)
```

ibolton336: @mturley

mturley: @ibolton336 I wonder if we should just create our own `type FileLoadError = { name?: string; message?: string; stack?: string; cause?: {}; }` and use it both in the yup schema and the `IReadFile` type?

ibolton336: Good call. I went down this route and decided against it because I was attempting duplicate types just for the form validation. Sounds like I just didn't take it far enough to replace the actual global type interface with our new custom types for specific object fields within. Makes sense. 

I found myself wanting to use yup to validate only part of our formData that exists within stongly typed objects. After trying the yup test, partial schema, & other custom methods applied to the yup instance itself, I kinda gave up and just threw an any type on there.

mturley: If it's more than a simple change to get rid of that `any` I'm fine with merging this and coming back to it.

ibolton336: I should have updated the PR status - I just realized that the bug is actually stating that the user cant upload 2 files simultaneously. Trying to fix this now.

ibolton336: @mturley FInished the updates if you have time for a review.

client/src/app/common/CustomRules/useRuleFiles.tsx
```diff
@@ -1,7 +1,7 @@
-import { useContext, useState } from "react";
+import { useContext, useRef, useState } from "react";
```
mturley, line 1: You've got the `useRef` as an unused import now

mturley (approved):
LGTM! 🎉 good stuff

PR #705, opened 2023-03-13
:bug: Don't allow large images to be uploaded

ibolton336:
Fixes https://issues.redhat.com/browse/MTA-299


client/src/app/pages/migration-targets/custom-target-form.tsx
```diff
@@ -383,29 +383,34 @@ export const CustomTargetForm: React.FC<CustomTargetFormProps> = ({
             dropzoneProps={{
               accept: ".png, .jpeg",
               maxSize: 1000000,
-              onDropRejected: () => setIsImageFileRejected(true),
+              onDropRejected: (event) => {
+                const currentFile = event[0];
+                if (currentFile.size > 1000000) {
+                  methods.setError("imageID", {
+                    type: "custom",
+                    message: "Max image file size of 1mb exceeded.",
```
mturley, line 391: Are we sure of the exact limit? Size units are weird. The text in Jira says "should be up to 1Mb of size", but I am guessing that is intended to mean `1 MB` since [Megabits are a thing](https://eastmanreference.com/mb-does-not-equal-mb), and I'm also not sure if we want 1 Megabyte (`1 MB`, 1000000 bytes) or 1 Mebibyte (`1 MiB`, 1049000 bytes). Whatever we go with we should use the exact units.

mturley (commented):
Code LGTM, just a small question/nit about the limit itself.

client/src/app/pages/migration-targets/custom-target-form.tsx
mturley: If you want we can change the `1mb` to `1 MB` in this text and call it done, and if someone needs it to mean `1 MiB` we can address in a followup 😆

ibolton336: Perfect - Just updated the text @mturley

mturley (approved):
LGTM, thanks!

PR #706, opened 2023-03-14
:bug: bulk copy: Fix assessment state evaluation

gildub:
If any of the application copied to has either an existing assessment or a review then the message should be displayed and confirmation is required:
Screenshot from 2023-03-14 15-33-18
Screenshot from 2023-03-14 15-33-26

Conversely no message and confirmation box are required when all of the applications copied to has no assessment or review:
Screenshot from 2023-03-14 15-33-40

Uses a function instead for useEffect().
resolves https://issues.redhat.com/browse/MTA-261


ibolton336:
Working for me - Nice job

PR #712, opened 2023-03-14
:bug: Don't auto select parsed sources from custom rule files

ibolton336:
According to Ramon, we do not need to parse source names from the custom-rules-files.


gildub (approved):
LGTM

PR #725, opened 2023-03-15
:bug: Fix error validation for image id

ibolton336:
Remove redundant onChange handler causing error to flash temporarily when the image is uploading


ibolton336:
https://issues.redhat.com/browse/MTA-345

PR #731, opened 2023-03-16
:sparkles: Add conditional validation for no-targets case

ibolton336:
Add custom validation for:
IF no targets are selected, THEN require a customRuleFile / Repo for a rule file(s)

As agreed upon by Ramon & Jeff in ForumTackle. This allows a user to run analysis with:
- only uploaded rules (which contain no targets)
- only a git repo of rules files.


gildub (approved):
LGTM

PR #745, opened 2023-03-20
:bug: Display tags color upon their category

gildub:
PF Label support the following 8 colors "gold" | "red" | "blue" | "cyan" | "green" | "orange" | "purple" | "grey" .
Tag colors depends upon their TagCategory colors which are defined as either keyword names (blue, etc) or hexadecimal values (for the recent ones V3).

Therefore depending on the color of the tag the Label can be displayed with a supported color or will fall back to "grey" (Label default) when not supported.

LabelCustomColor from migtools/lib-ui wraps Label to make tag color supported through CSS properties.

image

Resolves https://issues.redhat.com/browse/MTA-332

### DIFF
--- a/client/src/app/pages/applications/components/application-tags/application-tags.tsx
+++ b/client/src/app/pages/applications/components/application-tags/application-tags.tsx
@@ -13,10 +13,11 @@ import {
   ToolbarItem,
   ButtonVariant,
 } from "@patternfly/react-core";
+import { LabelCustomColor } from "@migtools/lib-ui";
+
 import spacing from "@patternfly/react-styles/css/utilities/Spacing/spacing";
 import textStyles from "@patternfly/react-styles/css/utilities/Text/text";
 import FilterIcon from "@patternfly/react-icons/dist/esm/icons/filter-icon";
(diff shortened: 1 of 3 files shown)


ibolton336 (changes requested):
Need to rethink this one together. Probably will need input from UXD and Ramon here.

mturley:
This might help :) https://github.com/migtools/lib-ui/pull/125

gildub:
@mturley, thanks and nice work with the migtools/lib-ui wrapper.
Good luck with the backend change in that regard.

client/src/app/pages/applications/components/application-tags/application-tags.tsx
```diff
@@ -222,22 +223,19 @@ export const ApplicationTags: React.FC<ApplicationTagsProps> = ({
                       </Text>
                     </TextContent>
                     <Flex>
-                      {tagsInThisCategoryInThisSource
-                        ?.sort((a, b) => a.name.localeCompare(b.name))
```
mturley, line 226: @gildub why remove the sorting here?

mturley (approved):
LGTM, left a question we can follow up on later.

PR #763, opened 2023-03-23
:bug: Update targets and sources to reflect rule file removal

ibolton336:


client/src/app/pages/applications/analysis-wizard/custom-rules.tsx
```diff
@@ -195,6 +195,22 @@ export const CustomRules: React.FC<CustomRulesProps> = (props) => {
                 type="button"
                 variant="plain"
                 onClick={() => {
+                  customRulesFiles.forEach((file) => {
+                    const { source, target } = parseRules(file);
+                    if (source && formSources.includes(source)) {
+                      const updatedSources = formSources.filter(
+                        (source) => source !== source
```
gildub, line 202: Need to change the iterator var so one can understand what is what.

client/src/app/pages/applications/analysis-wizard/custom-rules.tsx
```diff
@@ -195,6 +195,22 @@ export const CustomRules: React.FC<CustomRulesProps> = (props) => {
...
+                      );
+                      setValue("formSources", [...updatedSources]);
+                    }
+                    if (target && formTargets.includes(target)) {
+                      const updatedTargets = formSources.filter(
+                        (target) => target !== target
```
gildub, line 208: Need to change the iterator var so one can understand what is what.

gildub (changes requested):

client/src/app/pages/applications/analysis-wizard/custom-rules.tsx
ibolton336: Good catch.

gildub (approved):
Looks good to me.

PR #771, opened 2023-03-27
:bug: Do not refresh token when feature disabled

gildub:
When the "Use Refresh Token" (OpenID Connect Compatibility Modes) is off the refresh_token is not generated.

Resolves https://issues.redhat.com/browse/MTA-41


ibolton336 (approved):
LGTM

PR #772, opened 2023-03-27
:bug: Use app name as download file name

ibolton336:
Closes [MTA-393](https://issues.redhat.com/projects/MTA/issues/MTA-393)


gildub (changes requested):
Doesn't change anything on both Firefox and Chrome.
Actually the download filed is just ignored even with hard coded value.

PR #787, opened 2023-04-03
:seedling: Convert formik forms to react-hook-form

ibolton336:
- Formik removal: Refactor the application assessment wizard to use react-hook-form
- Breaks up application-assessment page level component into more manageable components. Moves the wizard into its own form component ApplicationAssessmentWizard
- Refactors the multi-input-selection & questionaire-form components to use HookFormPFGroup & HookFormPFText patterns respectively rather than the formik field array component.
- Converts the stakeholders form to use react-hook-form instead of formik
- Remove stakeholder & stakeholder group select components in favor of the RHF pattern. Another step toward removing the redundant formik/fetch select dropdown abstractions
- Renames duplicate ApplicationAssessment component to ApplicationAssessmentStatus for showing assessment status in the table
- Refactors Review form to use react-hook-form


gildub (changes requested):
Looks good meanwhile I found the following: 
- Editing an existing review doesn't work. The review wizard comes up as a new review
- The "Addtional notes or comments" field (at bottom of each wizard step) has a extra "Name" label which is not supposed to be there:
image

client/src/app/pages/applications/application-assessment/application-assessment.tsx
```diff
@@ -342,7 +65,7 @@ export const ApplicationAssessment: React.FC = () => {
       </ApplicationAssessmentPage>
     );
   }
-
+  console.log("assessment", assessment);
```
gildub, line 68: Not needed

client/src/app/pages/applications/application-assessment/components/application-assessment-wizard/application-assessment-wizard.tsx
```diff
@@ -0,0 +1,351 @@
...
+
+export interface ApplicationAssessmentWizardProps {
+  assessment?: Assessment;
+  isOpen: boolean;
+}
+
+export const ApplicationAssessmentWizard: React.FC<
+  ApplicationAssessmentWizardProps
+> = ({ assessment: initialAssessment, isOpen }) => {
+  console.log("initial assessment");
```
gildub, line 58: Do we need to keep it ?

client/src/app/pages/applications/application-assessment/components/application-assessment-wizard/application-assessment-wizard.tsx
```diff
@@ -0,0 +1,351 @@
...
+    });
+  const canJumpTo = maxCategoryWithData
+    ? sortedCategories.findIndex((f) => f.id === maxCategoryWithData.id) + 1
+    : 0;
+
+  const onSubmit = (formValues: ApplicationAssessmentWizardValues) => {
+    if (!initialAssessment) {
+      console.log("An assessment must exist in order to save the form");
+      return;
+    }
```
gildub, line 194: The form validation makes this block obsolete.

client/src/app/pages/applications/application-assessment/components/application-assessment-wizard/application-assessment-wizard.tsx
```diff
@@ -0,0 +1,351 @@
...
+        [SAVE_ACTION_KEY]: SAVE_ACTION_VALUE.SAVE_AS_DRAFT,
+      };
+    }, [initialAssessment]),
+    // defaultValues: {
+    //   stakeholders: initialAssessment?.stakeholders || [],
+    //   stakeholderGroups: initialAssessment?.stakeholderGroups || [],
+    //   comments: initialComments,
+    //   questions: initialQuestions,
+    //   [SAVE_ACTION_KEY]: SAVE_ACTION_VALUE.SAVE_AS_DRAFT,
+    // },
```
gildub, line 126: Cleanup ?

gildub (changes requested):
See inline

client/src/app/pages/applications/application-assessment/components/application-assessment-wizard/application-assessment-wizard.tsx
gildub: @ibolton336, can you confirm the form cannot be submitted when validation fails ?
In which case this block is not needed anymore.

ibolton336: @gildub This check is actually to make sure that the assessment exists so that we have an assessment to patch. Else we return with an error. The assessment is created before this final step in the wizard. 

There is an onInvalid callback to hook into that can be used to handle any validations we need. That is defined on line 180.

client/src/app/pages/applications/application-assessment/components/application-assessment-wizard/application-assessment-wizard.tsx
```diff
@@ -0,0 +1,343 @@
...
+  isOpen: boolean;
+}
+
+export const ApplicationAssessmentWizard: React.FC<
+  ApplicationAssessmentWizardProps
+> = ({ assessment: initialAssessment, isOpen }) => {
+  const { t } = useTranslation();
+
+  const [currentStep, setCurrentStep] = useState(0);
+  const [saveError, setSaveError] = useState<AxiosError>();
```
gildub, line 61: `saveError` is not read anywhere.

client/src/app/pages/applications/application-assessment/components/application-assessment-wizard/application-assessment-wizard.tsx
gildub: @ibolton336, I suppose my confusion comes from `initialAssessment` which is a rename of parameter `assessment`. Can't we just keep `assessment` ?

client/src/app/pages/applications/application-assessment/components/application-assessment-wizard/application-assessment-wizard.tsx
ibolton336: +1 Looks like this was never used. Will remove.

gildub (changes requested):
For the stakeholder form, the Job function and Group fields are marked as mandatory (with asterisk) while in fact there are optional.
Might have slipped through from recent change.

gildub (changes requested):
Looks good.
The only issue is when updating a tag category the assigned color is not displayed.

gildub (approved):

PR #598, opened 2023-01-31
:sparkles: Migration waves table architecture

ibolton336:
UPDATE: Now tracking a feature branch off of main for migration-waves
Closes #570 
Closes #572 
Closes #571

PR #792, opened 2023-04-04
:seedling: Refactor `ComposableTableWithControls` to use render-props pattern for more flexibility, apply changes to Migration Waves and Jira Trackers tables

mturley:
In preparation for the rest of the dynamic reports work, this PR takes the table abstraction introduced in https://github.com/konveyor/tackle2-ui/pull/598 for migration waves and refactors it to be even more composable. The goal is that all JSX used to render the Tr and Td elements is passed in from the consumer, with functions made available via the render-props pattern for abstracting repetitive props out into the component. This way the consumer can benefit from a simplified way to render tables without losing the flexibility to pass any props necessary on any row/cell.

There are some remaining questions to be answered about this abstraction, to be addressed in followup PRs. Some of these are covered in TODO comments here, in particular how to handle composable access to the props of the Thead/Tr elements without too much repetition, and how to handle non-compound-expandable rows and sorting controls. These will be addressed as part of adding the Issues table for dynamic reports.

### DIFF
--- a/client/src/app/pages/jira-trackers/jira-trackers.tsx
+++ b/client/src/app/pages/jira-trackers/jira-trackers.tsx
@@ -21,8 +21,8 @@ import {
   FilterType,
 } from "@app/shared/components/FilterToolbar";
 import { useFilterState } from "@app/shared/hooks/useFilterState";
-import { IComposableRow } from "@app/shared/components/composable-table-with-controls/composable-table-with-controls";
 import { useFetchJiraTrackers } from "@app/queries/jiratrackers";
+import { Tbody, Tr, Td } from "@patternfly/react-table";
 
 export const JiraTrackers: React.FC = () => {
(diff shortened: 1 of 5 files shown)


client/src/app/pages/jira-trackers/jira-trackers.tsx
```diff
@@ -148,7 +129,21 @@ export const JiraTrackers: React.FC = () => {
               />
             }
             toolbarClearAllFilters={handleOnClearAllFilters}
-          ></ComposableTableWithControls>
+            renderTableBody={() => (
+              <Tbody>
+                {currentPageItems?.map((jiraTracker) => (
```
ibolton336, line 134: This is awesome - thanks for getting this refactor in across the board.

ibolton336 (approved):

PR #793, opened 2023-04-04
:bug: Add loading spinner for active tasks

ibolton336:
https://issues.redhat.com/browse/MTA-438

### DIFF
--- a/client/src/app/pages/applications/components/application-detail-drawer/application-detail-drawer.tsx
+++ b/client/src/app/pages/applications/components/application-detail-drawer/application-detail-drawer.tsx
@@ -6,9 +6,11 @@ import {
   Tabs,
   Tab,
   TabTitleText,
+  Spinner,
+  Bullseye,
 } from "@patternfly/react-core";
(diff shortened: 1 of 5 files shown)


client/src/app/pages/applications/components/application-detail-drawer/application-detail-drawer.tsx
```diff
@@ -19,6 +21,7 @@ import { ApplicationTags } from "../application-tags";
 export interface IApplicationDetailDrawerProps
   extends Pick<IPageDrawerContentProps, "onCloseClick"> {
   application: Application | null;
+  task: Task | undefined | null;
```
mturley, line 24: I wonder if we should keep the `ApplicationDetailDrawer` logic a little more presentational, like maybe have `isTagsTabLoading` and `isReportsTabLoading` props here instead of passing the task all the way in? I'm not 100% sure if that's actually better, so I'm fine either way.

mturley (approved):
LGTM, but I left a suggestion you can take if you agree with it. Not a blocker.

PR #799, opened 2023-04-05
:bug: Disable add button when file with error is present

ibolton336:
- Fixes https://issues.redhat.com/browse/MTA-443
- Clears error alert when no files with error state exist after removal

### DIFF
--- a/client/src/app/common/CustomRules/useRuleFiles.tsx
+++ b/client/src/app/common/CustomRules/useRuleFiles.tsx
@@ -171,6 +171,9 @@ export default function useRuleFiles(
         )
     );
     setRuleFiles(updatedRuleFilesList);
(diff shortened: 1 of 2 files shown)


client/src/app/common/CustomRules/useRuleFiles.tsx
```diff
@@ -171,6 +171,9 @@ export default function useRuleFiles(
         )
     );
     setRuleFiles(updatedRuleFilesList);
+    if (!updatedRuleFilesList.some((file) => file.loadResult === "danger")) {
+      setUploadError("");
```
mturley, line 175: This is a case where `uploadError` should probably not be its own state since it could be derived from the contents of `ruleFiles` at render time. But that's a more complex refactor, so we could revisit it and merge this for now.

mturley (approved):
LGTM but there may be a better way to fix it.

PR #807, opened 2023-04-10
:sparkles: add tab for running arbitrary addons (#544)

djzager:
This PR introduces a new tab called "Addons" that allows users to run arbitrary addons (assuming extra addons are installed). The "Addons" tab is only visible if extra addons are visible to the hub.


ibolton336:
Closing this for now

PR #809, opened 2023-04-12
:seedling: Add `sortableColumns` param for `useTableControls` instead of `isSortable` on each `getThProps` call, move params that are unused in `useTableControlState` but required in `useTableControlProps` to the latter

mturley:
A followup to #808 after a quick discussion on Slack.

One nice future enhancement here would be to figure out how to take the `sortableColumns` and limit `getSortValues` and `initialSort` so they only require/allow columnKeys that are present in `sortableColumns`. Currently you must return a value for every columnKey from `getSortValues` and you can use any columnKey in `initialSort` even if those keys are not in `sortableColumns`. Tightening this up would require specifying the `sortableColumns` in the args of `useTableControlState` instead of `useTableControlProps`, and probably some conditional types wizardry that I still haven't learned about. This is good enough for now and still an improvement.


ibolton336 (approved):
LGTM!

PR #818, opened 2023-04-13
:bug: Fix category fallback color on tags table

ibolton336:
Adds additional color support to tag table


mturley (commented):
Copying notes from Slack: 

> I don't think we want to make that change actually, that means if the category on the backend has a color that isn't part of our curated list it will display a different color that isn't what's configured.
> I think we want the backend to be able to specify any color if it really wants to, and we can fall back to showing the hex if we don't have a name for it.
> The backend guys were going to change the automatic tags so they only use our curated colors.
> getCategoryFallbackColor was only intended for categories that have no color property on them at all for whatever reason.
> I think?
> if those assumptions are right it's working as intended (the ones that do match our color list in your screenshot are showing the name)

PR #819, opened 2023-04-14
:seedling: Improve separation of concerns in `useSortState` and `usePaginationState`, use column key union type in generics instead of entire columnNames object type

mturley:
As far as client-side table state goes, this completes the journey I started in #792 and continued in #808 and #809.

The `useSortState` and `usePaginationState` hooks we had included some presentational logic (they were coupled to returning PatternFly props), which made them hard to swap out with server-driven replacements without duplicating that logic. Also, `useSortState` based its logic on column indexes, and `useTableControls` attempts to expose only arguments concerned with columnKeys to insulate the consumer from that complexity, so `useTableControlState` had to wrap `useSortState` with extra logic to convert between the two.

This PR renames the existing `useSortState` and `usePaginationState` to `useLegacySortState` and `useLegacyPaginationState` respectively so that we don't have to refactor all the legacy tables in the UI to use the new implementations. It introduces replacement `useSortState` and `usePaginationState` hooks that address the above issues, and moves the presentational logic into some new `propHelpers` in `useTableControlProps`. The new versions of these two hooks now return only basic state that is similar to what we might use when working with server-side sorting and pagination, and they no longer depend on any types from PatternFly packages.

During this refactor, I noticed something strange about the way I had been using the `TColumnNames extends Record<string, string>` generic type. Apparently `keyof TColumnNames` was resolving to `string | number | symbol` in contexts where the specific column names are not in scope, which made it difficult to work with in useSortState (in reality the keys of the columnNames object will always be strings). To resolve this, I refactored all the generics of these new hooks and components so they take `TColumnKey extends string` as a type param, which can still be inferred from the `columnNames` object due to its type being `Record<TColumnKey, string>`.

This also presented the opportunity to have a third generic type `TSortableColumnKey extends TColumnKey`, so that the array passed in for `sortableColumns` is used as a narrower type to enforce what keys need to be returned from `getSortValues`. Now, you no longer need to return placeholder sort values for columns you don't intend to make sortable. Also, if you use an initialSort columnKey that is not one of your sortableColumns, this will cause a type error.

This should unblock the team from making further progress on tables like Migration Waves and Jira Trackers which have client-side pagination/sorting/filtering. Now that all the presentational logic is consolidated in `useTableControlProps`, `useTableControlState` is fairly trivial and we should be able to make a drop-in replacement that works with server-side state without having to refactor the existing tables further.

### DIFF
--- a/client/src/app/pages/waves/waves.tsx
+++ b/client/src/app/pages/waves/waves.tsx
@@ -107,15 +107,13 @@ export const Waves: React.FC = () => {
     numRenderedColumns,
     expansionState: { isCellExpanded },
     selectionState: { selectedItems },
-    paginationState: {
-      paginationProps, // TODO maybe paginationProps should be in propHelpers and not part of the responsibility of usePaginationState
-      currentPageItems,
-    },
+    paginationState: { currentPageItems },
     propHelpers: {
       toolbarProps,
       toolbarBulkSelectorProps,
       filterToolbarProps,
       paginationToolbarItemProps,
+      paginationProps,
       tableProps,
       getThProps,
       getTdProps,
--- a/client/src/app/pages/waves/waves.tsx
+++ b/client/src/app/pages/waves/waves.tsx
@@ -91,15 +91,12 @@ export const Waves: React.FC = () => {
         },
       },
     ],
+    sortableColumns: ["name", "startDate", "endDate"],
     getSortValues: (wave) => ({
       name: wave.name || "",
       startDate: wave.startDate || "",
       endDate: wave.endDate || "",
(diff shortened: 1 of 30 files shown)


ibolton336 (approved):

client/src/app/pages/waves/waves.tsx
```diff
@@ -91,31 +91,26 @@ export const Waves: React.FC = () => {
         },
       },
     ],
+    sortableColumns: ["name", "startDate", "endDate"],
```
ibolton336, line 94: Nice :)

Issue #821, opened 2023-04-17
Unable to login with auth enabled on Kubernetes with just port-forwarding tackle-ui


jwmatthews:
I am trying to login to Konveyor deployed on an Amazon EKS cluster.
I'm running:
`kubectl port-forward svc/tackle-ui 7080:8080 -n konveyor-tackle`

When I visit 127.0.0.1:7080 I see the login page, and when I login I see the same error as seen in:
https://github.com/konveyor/tackle2-operator/issues/167#issuecomment-1482627378

I'm redirected to `http://tackle-keycloak-sso.konveyor-tackle.svc:8080`

I chased this down over the weekend and documented notes here:
https://gist.github.com/jwmatthews/25ad0f2814d8bb6796649dcd22be4ed1

The fix for this ended up being we need to add a few more header parameters when we setup the reverse proxy for Keycloak:  
https://www.keycloak.org/server/reverseproxy

Note, the confusing piece is why does it work in OpenShift when we access the tackle-ui route?
I believe the answer is because in OpenShift we are accessing a Route which is implemented with haproxy reverse proxy that is setting these header parameters which then make it's way through...(my theory, not confirmed).

PR #822, opened 2023-04-17
:bug: Add forwarding headers to reverse proxy for keycloak

jwmatthews:
With this, I can run the below and login correctly with auth enabled

`kubectl port-forward svc/tackle-ui 7080:8080 -n konveyor-tackle`

Fixes https://github.com/konveyor/tackle2-ui/issues/821

For more info on debugging this, I captured some rough notes here: https://gist.github.com/jwmatthews/25ad0f2814d8bb6796649dcd22be4ed1

### DIFF
--- a/server/setupProxy.js
+++ b/server/setupProxy.js
@@ -7,6 +7,22 @@ module.exports = function (app) {
       target: process.env.KEYCLOAK_SERVER_URL || "http://localhost:9001",
       changeOrigin: true,
       logLevel: process.env.DEBUG ? "debug" : "info",
+      onProxyReq: (proxyReq, req, res) => {
+        // Keycloak needs these header set so we can function in Kubernetes (non-OpenShift)
+        // https://www.keycloak.org/server/reverseproxy
+        //
+        // Note, on OpenShift, this works as the haproxy implementation
+        // for the OpenShift route is setting these for us automatically
+        proxyReq.setHeader("X-Forwarded-For", req.socket.remoteAddress);
+        proxyReq.setHeader("X-Forwarded-Proto", req.protocol);
+        proxyReq.setHeader("X-Forwarded-Port", req.socket.localPort);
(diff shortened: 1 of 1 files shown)


jwmatthews:
@jmontleon would you help try this out on OpenShift and see if its working as expected?

ibolton336 (approved):

jmontleon:
This works for kubectl port-forward, but something isn't quite right with the OpenShift route:
I see this when logging in:
image

And if I proceed:
image

And if I send anyway:
image

jmontleon:
It looks like it's trying to connect to the route on port 8080.

jmontleon:
Removing these two lines, it worked for both kubectl port-forward and the route:
```
        proxyReq.setHeader("X-Forwarded-Proto", req.protocol);
        proxyReq.setHeader("X-Forwarded-Port", req.socket.localPort);
```

I'll try to narrow it down to one or the other (or both).

jmontleon (changes requested):
This doesn't quite work right with the OpenShift route.

jmontleon:
`This form is not secure` appears to come from using `proxyReq.setHeader("X-Forwarded-Proto", req.protocol);`. If I remove `proxyReq.setHeader("X-Forwarded-Port", req.socket.localPort);` I still get the warning but I can log in and don't get redirected to port 8080.

With just `proxyReq.setHeader("X-Forwarded-Port", req.socket.localPort);` no insecure warning, but then I get redirected to port 8080 on login.

jwmatthews:
@jmontleon would you take another look at this with the last commit?  

I tested this on an EKS cluster and OpenShift cluster, looking good for use with port-forward and behind the OpenShift Route

jmontleon (approved):
This works for me as it is now, both over the OpenShift route and kubectl port-forward

ibolton336 (approved):

PR #828, opened 2023-04-19
:seedling: Convert stakeholder table to new table abstraction

ibolton336:
- useTableControls implementation for Stakeholders expandable form 
@mturley Nice work! We don't have to necessarily merge this as I know it is still under construction a bit. Just wanted to explore the changes. Works flawlessly. Only thing I did notice - we might need a way to make individual rows expandable. If I have a stakeholder with no groups assigned, I wonder if there is a way to make only that row not have an expandable icon displayed without shifting the columns/table rows.

### DIFF
--- a/client/src/app/pages/controls/stakeholders/stakeholders.tsx
+++ b/client/src/app/pages/controls/stakeholders/stakeholders.tsx
@@ -1,7 +1,6 @@
 import * as React from "react";
 import { AxiosError, AxiosResponse } from "axios";
 import { useTranslation } from "react-i18next";
-import { useSelectionState } from "@migtools/lib-ui";
 import {
   Button,
   ButtonVariant,
@@ -9,52 +8,49 @@ import {
   DescriptionListDescription,
   DescriptionListGroup,
   DescriptionListTerm,
+  Toolbar,
+  ToolbarContent,
   ToolbarGroup,
   ToolbarItem,
 } from "@patternfly/react-core";
 import {
-  cellWidth,
-  expandable,
-  ICell,
-  IExtraData,
-  IRow,
-  IRowData,
-  sortable,
-  TableText,
+  ExpandableRowContent,
+  TableComposable,
+  Tbody,
+  Td,
+  Th,
+  Thead,
+  Tr,
 } from "@patternfly/react-table";
+
+import spacing from "@patternfly/react-styles/css/utilities/Spacing/spacing";
 
 import {
   AppPlaceholder,
   AppTableActionButtons,
-  AppTableWithControls,
   ConditionalRender,
   ConfirmDialog,
-  NoDataEmptyState,
 } from "@app/shared/components";
 import { getAxiosErrorMessage } from "@app/utils/utils";
 import { Stakeholder } from "@app/api/models";
 import { NewStakeholderModal } from "./components/new-stakeholder-modal";
 import { UpdateStakeholderModal } from "./components/update-stakeholder-modal";
-import { useLegacyPaginationState } from "@app/shared/hooks/useLegacyPaginationState";
 import {
-  FilterCategory,
   FilterToolbar,
   FilterType,
 } from "@app/shared/components/FilterToolbar";
-import { useFilterState } from "@app/shared/hooks/useFilterState";
-import { useLegacySortState } from "@app/shared/hooks/useLegacySortState";
-import { controlsWriteScopes, RBAC, RBAC_TYPE } from "@app/rbac";
 import {
   useDeleteStakeholderMutation,
   useFetchStakeholders,
 } from "@app/queries/stakeholders";
 import { NotificationsContext } from "@app/shared/notifications-context";
-
-const ENTITY_FIELD = "entity";
-
-const getRow = (rowData: IRowData): Stakeholder => {
-  return rowData[ENTITY_FIELD];
-};
+import { useTableControls } from "@app/shared/hooks/use-table-controls";
+import { SimplePagination } from "@app/shared/components/simple-pagination";
+import {
+  ConditionalTableBody,
+  TableHeaderContentWithControls,
+  TableRowContentWithControls,
+} from "@app/shared/components/table-controls";
 
 export const Stakeholders: React.FC = () => {
   const { t } = useTranslation();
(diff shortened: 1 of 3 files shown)


mturley:
Nice!

Interesting, yeah, I think we can probably take advantage of the composable nature there and have some conditional expression on that cell where you only spread the `getCompoundExpandTdProps(...)` if that row should have it, else spread `getTdProps(...)`. Either that or we can add an `isCellExpandable` boolean to `getCompoundExpandTdProps` and have it exclude the necessary props when that's false.

mturley (dismissed):
LGTM! I'm totally fine merging this if you want, I already keep the migration waves / jira tables up to date whenever I change these hooks so it's not too much effort to manage one more. Within reason, the more tables we have using it the more confident we can be in the abstraction

mturley:
Actually @ibolton336 I approved too quickly, some minor things:
* On main, it looks like there is padding/margin around the contents of the expanded row content, and here the text is right up against the borders there
* For this single-expand use case I actually wanted to add support for automatically rendering that expand toggle cell as part of `<TableRowContentWithControls>`, but we can do that as a separate refactor after this if you want. Would just mean a little less code in the table.

ibolton336:
> Actually @ibolton336 I approved too quickly, some minor things:
> 
> * On main, it looks like there is padding/margin around the contents of the expanded row content, and here the text is right up against the borders there

For some reason I am seeing an extra <td> in the expandable row that seems to be implicitly created by the legacy table.  

Screenshot 2023-04-20 at 9 26 13 AM

client/src/app/pages/controls/stakeholders/stakeholders.tsx
```diff
@@ -284,16 +285,7 @@ export const Stakeholders: React.FC = () => {
...
-                            rowIndex,
-                            isExpanded: isCellExpanded(stakeholder),
-                            onToggle: () =>
-                              setCellExpanded({
-                                item: stakeholder,
-                                isExpanding: !isCellExpanded(stakeholder),
-                              }),
-                            expandId: "composable-expandable-example",
-                          }}
+                          {...getExpandTdProps({ item: stakeholder, rowIndex })}
```
mturley, line 288: Can we rename it to `getSingleExpandTdProps` to distinguish it from the compound expand version?

client/src/app/pages/applications/components/application-form/application-form.tsx
```diff
@@ -666,18 +666,20 @@ export const ApplicationForm: React.FC<ApplicationFormProps> = ({
...
-                aria-label="binary packaging details"
-                bodyContent={t("message.binaryPackaging")}
-                className="popover"
-              >
-                <span className="pf-c-icon pf-m-info">
-                  <QuestionCircleIcon />
-                </span>
-              </Popover>
-            }
+            formGroupProps={{
```
mturley, line 669: Looks like the diff contains some stuff from another change?

client/src/app/pages/controls/stakeholders/stakeholders.tsx
```diff
@@ -306,68 +116,244 @@ export const Stakeholders: React.FC = () => {
...
+                          <AppTableActionButtons
+                            onEdit={() => editRow(stakeholder)}
+                            onDelete={() => deleteRow(stakeholder)}
+                          />
+                        </Td>
+                      </TableRowContentWithControls>
+                    </Tr>
+                    {isCellExpanded(stakeholder) ? (
+                      <Tr isExpanded>
+                        <Td />
```
mturley, line 322: Oh I see, this is to space it out so it is indented past the toggle. I guess that makes sense... hm, it does kinda suck that we have to remember to put this extra Td in though. I wonder if it's worth adding a new component for `<ExpandedContentRowWithControls>` or something that renders the Tr, the two Tds with the right props, and then passes children in.

Probably not worth it for now, I'll make a note to maybe mess with that later.

client/src/app/shared/hooks/use-table-controls/useTableControlProps.ts
```diff
@@ -173,6 +173,25 @@ export const useTableControlProps = <
...
+  }): Omit<TdProps, "ref"> => ({
+    expand: {
+      rowIndex,
+      isExpanded: isCellExpanded(item),
+      onToggle: () =>
+        setCellExpanded({
+          item,
+          isExpanding: !isCellExpanded(item),
+        }),
+      expandId: "expandable-example",
```
mturley, line 191: These ids need to be unique for QE tests. We probably want to be able to pass in a custom id here, or maybe have it be something built-in like:
```ts
expandId: `expandable-row-${item.name}`
```

mturley (changes requested):
Looking good! A couple minor things

ibolton336:
> Actually @ibolton336 I approved too quickly, some minor things:
> 
> * On main, it looks like there is padding/margin around the contents of the expanded row content, and here the text is right up against the borders there
> * For this single-expand use case I actually wanted to add support for automatically rendering that expand toggle cell as part of `<TableRowContentWithControls>`, but we can do that as a separate refactor after this if you want. Would just mean a little less code in the table.

Thanks @mturley For all the reviews. I think I hit all of the requested updates. Whenever you have time.

mturley (approved):

sshveta:
HI @mturley @ibolton336 .
In this PR the stakeholder table ID is changed to stakeholders-table .

All the other tables (like stakeholder Group , Business Service , applications page ) in MTA are using "table[aria-label='main-table']";.

Is it possible to keep the ID same and generic for all the tables ? 
That was enabling us to write common methods to delete all entities in page . 

Please let me know what you think .

mturley:
Hi @sshveta. 

I think we failed to properly communicate here, sorry about that. As established by #443 (which perhaps should not be closed), we do want to align on using `id` for QE tests where possible (or more specifically, avoiding having tests depend on `aria-label` which is user-facing text that is subject to change).

Perhaps we should use `id="main-table"` for each of the primary tables on the page, that is a decent idea. I'm curious about @ibolton336 's thoughts.

ibolton336:
@mturley @sshveta Sorry for the oversight here. We can certainly move to a common id for these tables if that would be helpful.

sshveta:
Thanks @ibolton336 . Please let me know if I need to open a bug/issue for this .

Issue #827, opened 2023-04-19
Dependencies Table & server side pagination/filter/sort 


ibolton336:
- Add skeleton for the dependencies table. Using this issue to cover the initial server side filter/sort/pagination
- Add drawer support for dependencies table. This includes the filter toolbar and the query for the list of affected applications. 
![image](https://user-images.githubusercontent.com/11218376/233121700-f80f3d09-9040-4f33-b32d-cd0a558eaa5b.png)
![image](https://user-images.githubusercontent.com/11218376/233121753-9caa64af-0a0c-49ad-ba29-d6e0a64f5d04.png)

mturley:
Nearly closed by #839, I just need to add the drawer to this page and figure out what columns we can display.

mturley:
Not sure how important it is right now to make sure these issues are up to date, but it might be reasonable to move the drawer contents part of this to another issue and mark this one resolved. The drawer contents stuff will need to wait until after Issues stuff is done

PR #829, opened 2023-04-19
:sparkles: Dynamic Reports: Add Dependencies table with server-side pagination and sorting (no filters yet)

mturley:
Adds the Dependencies page, a new query for `/hub/analyses/dependencies/`, and implements server-side sorting and pagination. Also includes some helpers for an initial attempt at server-side filtering, to be finished in a followup PR. (This change introduces some filter-related TODO comments that I'll remove in the next PR).

* Centralizes the TypeScript types for arguments and return values of table-related hooks in one file at `@app/shared/hooks/use-table-controls/types.ts`, with more explicit inheritance and comments explaining the flow of arguments from one hook to the next.
* Adjusts types for some of the table-related helper components to be more explicit and readable.
* Separates `useSortState` into `useSortState` and `useSortDerivedState`, with the former including only arguments and return values necessary for the storage of sort-related state and the latter including logic used to derive the sorted items (perform client-side sorting).
* Separates `usePaginationState` into `usePaginationState` and `usePaginationDerivedState` following the same pattern.
* Adds a new hook `useUrlParams`, a generic hook which acts like `useState` (returns the same `[value, setValue]` tuple) but takes some extra arguments to facilitate using URL search parameters as the state storage.
  * Some extra magic here: The value stored here can be of any generic type as long as it can be serialized to a string, because the hook requires `serialize` and `deserialize` functions which convert that type to/from a plain object with string values. From there, the hook internally converts that to and from `URLSearchParams` objects and handles the react-router logic.
  * Another neat thing about `useUrlParams`: When the setter function is called, the operation it performs to update URL params is atomic. It references the current global `document.location.search` when updating and retains all other params present there, so (a) multiple instances of it can be used for multiple sets of URL params and (b) calling the setter to update those can be done in sequence without waiting for another render, and values will not clobber each other.
* Adds a new hook `useSortUrlParams` which has the same arguments and return values (sharing interface definitions) of `useSortState`, but persists the state using `useUrlParams` instead of `useState`. This is a drop-in replacement for `useSortState` and we can easily choose whether to use URL params for state or not.
* Adds a new hook `usePaginationUrlParams` which is a similar drop-in replacement for `usePaginationState`.
* Adds a new hook `useTableControlUrlParams` which can be used in place of `useTableControlState`. It's not a full drop-in replacement because it doesn't concern itself with args needed for derived state (client-side sorting/pagination). It returns a subset of the args needed for `useTableControlProps`, the remaining args to be derived by the consumer based on query data from the API (server-side sorting/pagination).
* Adds types to `@app/api/models.ts`:
  * `HubFilter`: A structured representation of the filters we can express in the hub's URL params (temporary -- will need to adjust this in the next PR).
  * `HubRequestParams`: A structured representation of the URL params we need to pass to the hub for server-side filtering/sorting/pagination.
  * `HubPaginatedResult`: A wrapper for results of a server-paginated request with metadata for the total and a copy of the params that were used for the request.
* Adds a file `@app/utils/hub-request-utils.ts` including:
  * `serializeRequestParamsForHub`: Converts a `HubRequestParams` object into the `URLSearchParams` objects to be passed into Axios requests
* `getHubRequestParams`: Takes `sortState` and `paginationState` objects returned by the swappable `useSortState`/`useSortUrlParams` and `usePaginationState`/`usePaginationUrlParams`, along with some additional args, and returns the `HubRequestParams` object
* In `@app/api/rest.ts`, adds a generic function `getHubPaginatedResult` which takes any URL and `HubRequestParams` object and sends the Axios request, then parses the headers of the response to return the `HubPaginatedResult` object. Also adds a `getDependencies` function here which uses that function to fetch the dependency objects.
* Adds a `useFetchDependencies` hook which uses the above helpers with react-query to facilitate fetching the server-paginated/sorted dependencies array.
  * Note that the query key here includes the serialized params from `serializeRequestParamsForHub`, which is a unique string that will change any time the sort/pagination controls are interacted with. This means we get a cached copy of the results for each specific paginated/sorted list, and if you go back to a previously loaded page or sort, it will appear instantly while being refetched in the background.
* Adds the Dependencies page (`@app/pages/dependencies`), route and nav item, with a new Dependencies table wired up with all of the above! Server-side pagination and sorting are working end to end. Filtering is still not included. Usage of the above hooks on this page is as follows:
  * We call `useTableControlUrlParams` with as many of the args as we can define without state or query data in scope and get `tableControlState` returned, which is an object including all of the arguments passed in with `sortState` and `paginationState` added. (these objects include the setters/getters for each set of state).
  * We pull out the `sortState` and `paginationState` and pass them to `getHubRequestParams`, which we call as part of `useFetchDependencies`. We get API data back including the current page of items and the total item count.
  * We call `useTableControlProps`, which is also used internally by `useTableControls` on the existing client-paginated tables, and we pass in `tableControlState` (all the args and state collected so far) along with the `currentPageItems`, `totalItemCount`, `isLoading`, and additional arguments for the state we haven't abstracted out to the URL params and the server (`filterState` which is currently a stub, `expansionState` and `selectionState` which currently call the same hooks we use internally in `useTableControlState` when we use `useTableControls` on the existing tables).
  * We get all the same propHelpers and such returned that we have on the existing tables. Rendering the table is exactly the same as the client-paginated tables!

This PR does make one very minor change to the existing tables that use `useTableControls`: the `currentPageItems` variable is no longer part of the `paginationState` object returned there, it is now at the root level of the returned object. @ibolton336 this should be the only thing you need to adjust in #828 assuming we merge this first.


ibolton336 (approved):
🚀 🌔 LGTM

mturley:
Commenting here to reference that this is part of #827 since editing the description didn't auto-tag that over there

Issue #589, opened 2023-01-25
Add JIRA credential type


ibolton336:
![Image](https://user-images.githubusercontent.com/11218376/214674942-422c3c2f-109b-4d8d-b729-e8a5147fafb4.png)

PR #830, opened 2023-04-19
:sparkles: Adds Jira creds

gildub:
Resolves https://github.com/konveyor/tackle2-ui/issues/589

### DIFF
--- a/client/src/app/FeatureFlags.ts
+++ b/client/src/app/FeatureFlags.ts
@@ -1,3 +1,3 @@
 export const FEATURES_ENABLED = {
-  migrationWaves: true,
+  migrationWaves: false,
 };
--- a/client/src/app/FeatureFlags.ts
+++ b/client/src/app/FeatureFlags.ts
@@ -1,3 +1,3 @@
 export const FEATURES_ENABLED = {
-  migrationWaves: false,
+  migrationWaves: true,
 };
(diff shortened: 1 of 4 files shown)


client/src/app/pages/identities/components/identity-form/identity-form.tsx
```diff
@@ -637,6 +639,35 @@ export const IdentityForm: React.FC<IdentityFormProps> = ({
         </>
       )}
 
+      {values?.kind === "jira" && (
```
ibolton336, line 642: Since we are using the same form validation between jira & proxy... Wondering if we can use the same form fields also?

client/src/app/FeatureFlags.ts
```diff
@@ -1,3 +1,4 @@
 export const FEATURES_ENABLED = {
   migrationWaves: false,
+  jira: false,
```
ibolton336, line 3: I think Jira can use the migration waves flag as it already does in the routes file for show/hiding the side nav options.

ibolton336:
Screenshot 2023-04-21 at 9 58 45 AM

Looks like something is up with the validation.

ibolton336 (changes requested):
Validation is currently not working for jira creds

ibolton336 (approved):

Issue #836, opened 2023-04-25
:bug: Editing application tags overrides automated tag source


gildub:
When updating an application by adding a tag "manually" makes all the tags that were automatically added have their source changed from "Analysis" to "" (Anonymous which is displayed to user as "Manual").

PR #837, opened 2023-04-25
:bug: Fix tag source override when updating application

gildub:
The `ApplicationForm` now handles directly tags (TagRef[]) allowing to preserve value of existing tags "source" field and to avoid automated tags (source = "Analysis") to be overridden with source = "". 

Resolves https://github.com/konveyor/tackle2-ui/issues/836


ibolton336 (approved):
Good job on this fix & nice catch.

PR #842, opened 2023-04-28
:bug: Fix persona selection to reflect current page

ibolton336:
- Convert sidenav persona select to simpleselect
- Fixes a bug where the select was not reflecting the current page on load


gildub (approved):
LGTM

PR #843, opened 2023-04-29
:sparkles: Add expandable issues table content & affected applications page

ibolton336:
- Adds initial issues table expandable area content
- Adds numsearch filter type for handling filtering by number if the need arises. I have commented out the effort filter until more info surfaces on the needs.
- Adds affected applications view with apps viewed by associated issue ruleid. Uses react-router route params approach. 
- https://stackoverflow.com/a/38576357 - disable dot rule to allow routing to rule urls with dot notation.

TODO: 
- convert search filters to select where needed for applications
- Improve grouped categories implementation to dynamically hide empty groups
- Add validation description to composite issue expanded area
- Use new hub api label changes

### DIFF
--- a/client/src/app/pages/issues/affected-applications/affected-applications.tsx
+++ b/client/src/app/pages/issues/affected-applications/affected-applications.tsx
@@ -1,0 +1,217 @@
+import * as React from "react";
+import {
+  Breadcrumb,
+  BreadcrumbItem,
+  PageSection,
+  PageSectionVariants,
+  Text,
+  TextContent,
+  Toolbar,
+  ToolbarContent,
+  ToolbarItem,
+} from "@patternfly/react-core";
+import { useTranslation } from "react-i18next";
+import { AppPlaceholder, ConditionalRender } from "@app/shared/components";
+import {
+  TableComposable,
+  Tbody,
+  Td,
+  Th,
+  Thead,
+  Tr,
+} from "@patternfly/react-table";
+import {
+  useTableControlUrlParams,
+  getHubRequestParams,
+} from "@app/shared/hooks/table-controls";
+import { SimplePagination } from "@app/shared/components/simple-pagination";
+import {
+  ConditionalTableBody,
+  TableHeaderContentWithControls,
+  TableRowContentWithControls,
+} from "@app/shared/components/table-controls";
+import { useFetchIssues } from "@app/queries/issues";
+import { useLocalTableControls } from "@app/shared/hooks/table-controls";
+import { useFetchApplications } from "@app/queries/applications";
+import { Application } from "@app/api/models";
+import { Link, useParams } from "react-router-dom";
+interface IAffectedApplicationsRouteParams {
+  ruleidparam: string;
+}
+
+export const AffectedApplications: React.FC = () => {
+  const { t } = useTranslation();
+
+  const { ruleidparam } = useParams<IAffectedApplicationsRouteParams>();
+
+  const tableControlState = useTableControlUrlParams({
+    columnNames: {
+      name: "Name",
+      description: "Description",
+      businessService: "Business serice",
+      tags: "Tags",
+    },
+    sortableColumns: ["name"],
+    initialSort: {
+      columnKey: "name",
+      direction: "asc",
+    },
+    initialItemsPerPage: 10,
+  });
+
+  const {
+    result: { data: issues, total: totalItemCount },
+    isFetching,
+    fetchError,
+  } = useFetchIssues(
+    getHubRequestParams({
+      filterState: {
+        filterValues: {
+          ruleid: [ruleidparam || ""],
+        },
+        setFilterValues: tableControlState.filterState.setFilterValues,
+      },
+    })
+  );
+
+  const { data: applications } = useFetchApplications();
+  const issueAppIds = issues.map((issue) => issue.application);
+  const affectedApplications: Application[] = applications.filter(
+    (app) => app.id && issueAppIds.includes(app.id)
+  );
+
+  const tableControls = useLocalTableControls({
+    idProperty: "name",
+    items: applications,
+    columnNames: {
+      name: "Name",
+      description: "Description",
+      businessService: "Business serice",
+      tags: "Tags",
+    },
+    sortableColumns: ["name"],
+    getSortValues: (item) => ({
+      name: item?.name || "",
+    }),
+    initialSort: { columnKey: "name", direction: "asc" },
+    hasPagination: true,
+    isLoading: isFetching,
+  });
+
+  const {
+    currentPageItems,
+    numRenderedColumns,
+    propHelpers: {
+      toolbarProps,
+      paginationToolbarItemProps,
+      paginationProps,
+      tableProps,
+      getThProps,
+      getTdProps,
+    },
+  } = tableControls;
+  console.log({ currentPageItems, totalItemCount });
+
+  return (
+    <>
+      <PageSection variant={PageSectionVariants.light}>
+        <TextContent>
+          <Text component="h1">{t("terms.issues")}</Text>
+        </TextContent>
+        <Breadcrumb>
+          <BreadcrumbItem>
+            <Link to="/composite/issues">Issues</Link>
+          </BreadcrumbItem>
+          <BreadcrumbItem to="#" isActive>
+            {ruleidparam || "Active rule"}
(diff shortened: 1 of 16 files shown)


client/src/app/shared/components/FilterToolbar/FilterToolbar.tsx
```diff
@@ -113,6 +117,34 @@ export const FilterToolbar = <TItem, TFilterCategoryKey extends string>({
...
+        .filter((filterCategory) => filterCategory.filterGroup === filterGroup)
+        .map((filterCategory) => {
+          return (
+            <DropdownItem
+              id={`filter-category-${filterCategory.key}`}
+              key={filterCategory.key}
+              onClick={() => onCategorySelect(filterCategory)}
+            >
+              {filterCategory.title}
+            </DropdownItem>
```
mturley, line 132: You could probably factor out this DropdownItem piece that repeats below as a `renderDropdownItem` function or its own component, if that's worth it. Maybe not worth it.

client/src/app/shared/components/FilterToolbar/FilterToolbar.tsx
```diff
@@ -30,6 +31,7 @@ export interface IBasicFilterCategory<
   key: TFilterCategoryKey;
   title: string;
   type: FilterType; // If we want to support arbitrary filter types, this could be a React node that consumes context instead of an enum
+  filterGroup: string;
```
mturley, line 34: I think we'll probably need this to be optional, so it can be omitted for filters that don't have groups.

client/src/app/shared/components/FilterToolbar/FilterToolbar.tsx
```diff
@@ -75,6 +77,7 @@ export const getFilterLogicOperator = <
 
 export interface IFilterToolbarProps<TItem, TFilterCategoryKey extends string> {
   filterCategories: FilterCategory<TItem, TFilterCategoryKey>[];
+  filterGroups: string[];
```
mturley, line 80: You might be able to totally get away with not having a `filterGroups` param here at all. All the filterCategories have their `filterGroup`, so the `filterGroups` could be derived from those below.

client/src/app/shared/components/FilterToolbar/FilterToolbar.tsx
```diff
@@ -113,6 +117,34 @@ export const FilterToolbar = <TItem, TFilterCategoryKey extends string>({
     (category) => category.key === currenTFilterCategoryKey
   );
 
+  const groupedDropdownItems = filterGroups?.map((filterGroup) => (
```
mturley, line 120: You could do something just before this line, like:

```ts
const filterGroups = filterCategories.reduce(
  (groups, category) =>
    !category.filterGroup || groups.includes(category.filterGroup)
      ? groups
      : [...groups, category.filterGroup],
  [] as string[]
);
```

and then you don't need to pass in `filterGroups`, they'll just be ordered according to the order they appear in `filterCategories`. And below you can just use `!!filterGroups.length` or `filterGroups.length > 0` without the `?` because it'll always be defined, just might be empty.

mturley (changes requested):
This is good! One minor improvement we could maybe make to reduce repetition.

Also, there is the weird case where we might get confused why a category isn't showing up if we accidentally have some categories with groups and some without. Not sure if that matters at this stage?

client/src/app/shared/hooks/table-controls/filtering/getFilterHubRequestParams.ts
```diff
@@ -42,6 +42,13 @@ export const getFilterHubRequestParams = <
     // Note: If we need to support more of the logic operators in HubFilter in the future,
     //       we'll need to figure out how to express those on the FilterCategory objects
     //       and translate them here.
+    if (filterCategory.type === "numsearch") {
+      params.filters?.push({
+        field: categoryKey,
+        operator: "=",
+        value: Number(filterValue[0]),
+      });
+    }
```
mturley, line 51: I'm surprised TS is okay with this since `value` of a `HubFilter` has to be a string or an object. We probably want to change the `HubFilter` type (at `client/src/app/api/models.ts`) so the value is `string | number` and also has `list: (string | number)[];` maybe? Also I left behind a TODO comment there that we should remove, saying we should remove HubFilter before I changed my mind 🙃

client/src/app/shared/hooks/table-controls/filtering/getFilterHubRequestParams.ts
```diff
@@ -78,6 +85,8 @@ export const serializeFilterForHub = (filter: HubFilter): string => {
   const joinedValue =
     typeof value === "string"
       ? wrapInQuotesAndEscape(value)
+      : typeof value === "number"
+      ? value
```
mturley, line 89: You probably need it wrapped in quotes here as well, so you could either adapt the `wrapInQuotesAndEscape` function or just do `"${value}"` here.

client/src/app/shared/hooks/table-controls/filtering/getFilterHubRequestParams.ts
mturley: Oh I totally missed that you fixed that

client/src/app/pages/issues/issues.tsx
```diff
@@ -300,9 +304,9 @@ export const Issues: React.FC = () => {
                                   <Tooltip content="View Report">
                                     <Button variant="link" isInline>
                                       <Link
-                                        // to={`/hub/applications/${application.id}/bucket${task?.data?.output}`}
-                                        to={""}
-                                        target="_blank"
+                                        // to={`${ANALYSIS_ISSUES}/${issue.ruleID}`}
+                                        to={`/issues/${issue.ruleID}`}
```
ibolton336, line 308: @mturley I am wondering if you have a better idea of how to persist the filter state down to the affected apps page. This works, but probably need something that also preserves the URL bar so that we can still refresh the browser and remain on the affected apps page. Otherwise may just redirect back to the composite issues page if this isn't feasible.

client/src/app/pages/issues/affected-applications/affected-applications.tsx
```diff
@@ -0,0 +1,237 @@
...
+import { Link, useParams } from "react-router-dom";
+import { useLocalTableControls } from "@app/shared/hooks/table-controls";
+import { useFetchApplications } from "@app/queries/applications";
+import { Application } from "@app/api/models";
+interface IAffectedApplicationsPageParams {
+  ruleid: string;
+}
+export const AffectedApplications: React.FC = () => {
+  const { t } = useTranslation();
+  const { ruleid = null } = useParams<IAffectedApplicationsPageParams>();
```
ibolton336, line 61: @mturley This works for now, but wondering if I should just somehow pass the ruleid from the parent composite issues table in another way.

client/src/app/pages/issues/issues.tsx
ibolton336: https://www.npmjs.com/package/query-string Looking into this as a solution for parsing the rule-id and just keeping the linked url as 

```
to={`/issues/?filter=ruleid%7E%22*${issue.ruleID}*%22&sort=asc:ruleID&limit=10%offset=0`}
```
to satisfy the server requests

ibolton336: Query-string ended up not working out since we are doing some custom filtering. RegEx works fine, though.

mturley (approved):
LGTM

client/src/app/queries/applications.ts
```diff
@@ -26,6 +26,7 @@ export const ApplicationsQueryKey = "applications";
 export const useFetchApplications = () => {
   const queryClient = useQueryClient();
   const { isLoading, error, refetch, data } = useQuery({
+    initialData: [],
```
gildub, line 29: What's for ?

mturley: @ibolton336 I'm also curious why this is necessary. Maybe we can merge this and follow up though, to unblock further work on the Issues page.

mturley: @gildub this came up on slack, it was to prevent having to propagate a bunch of undefined checks everywhere the apps fetch is used. I talked with Ian and in my followup PR I plan to replace this with returning `data || []` from the hook to solve the same problem without changing the initial fetch behavior.

PR #848, opened 2023-05-03
:bug: Fix unnecessary page remounts: don't attempt to render less boilerplate on pages without a drawer

mturley:
Fixes the entire app re-rendering when PageDrawerContent renders for the first time, which was causing all our applications table queries to fetch twice and the initial fetch updating state on an older unmounted instance of the page (resulting in React warnings).


gildub (approved):
LGTM

ibolton336 (approved):

PR #860, opened 2023-05-09
:sparkles: SearchFilterControl: Ignore multiple space

gildub:
Resolves https://issues.redhat.com/browse/MTA-467

### DIFF
--- a/client/src/app/shared/components/FilterToolbar/SearchFilterControl.tsx
+++ b/client/src/app/shared/components/FilterToolbar/SearchFilterControl.tsx
@@ -28,11 +28,11 @@ export const SearchFilterControl = <TItem, TFilterCategoryKey extends string>({
   // Keep internal copy of value until submitted by user
   const [inputValue, setInputValue] = React.useState(filterValue?.[0] || "");
   // Update it if it changes externally
-  React.useEffect(() => {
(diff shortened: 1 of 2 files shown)


client/src/app/shared/components/FilterToolbar/SearchFilterControl.tsx
```diff
@@ -28,11 +28,11 @@ export const SearchFilterControl = <TItem, TFilterCategoryKey extends string>({
   // Keep internal copy of value until submitted by user
   const [inputValue, setInputValue] = React.useState(filterValue?.[0] || "");
   // Update it if it changes externally
-  React.useEffect(() => {
-    setInputValue(filterValue?.[0] || "");
-  }, [filterValue]);
+  React.useEffect(() => setInputValue(filterValue?.[0] || ""), [filterValue]);
```
mturley, line 31: FYI, in general we shouldn't use single-expression inline arrow functions in a useEffect because the return value of the useEffect callback matters (if you return a function there it will be called as cleanup before re-running the effect). I don't think `setInputValue` returns anything so maybe this one is fine, but it's a convention (at least IMO) that I like to stay consistent about.

mturley (approved):
LGTM, but one minor comment you can take or leave.

PR #883, opened 2023-05-11
:bug: Fix RegExp

gildub:
Addressed URL validation for application-form and instance-form after recent change.

### DIFF
--- a/client/src/app/pages/applications/components/application-form/application-form.tsx
+++ b/client/src/app/pages/applications/components/application-form/application-form.tsx
@@ -138,8 +138,11 @@ export const ApplicationForm: React.FC<ApplicationFormProps> = ({
   const customURLValidation = (schema: StringSchema) => {
     const gitUrlRegex =
       /(?:git|ssh|https?|git@[-\w.]+):(\/\/)?(.*?)(\.git)(\/?|\#[-\d\w._]+?)$/;
+
+    const standardUrlRegex = new RegExp(standardURLRegex);
+
     const containsURL = (string: string) =>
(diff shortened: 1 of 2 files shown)


client/src/app/pages/applications/components/application-form/application-form.tsx
```diff
@@ -138,8 +138,11 @@ export const ApplicationForm: React.FC<ApplicationFormProps> = ({
   const customURLValidation = (schema: StringSchema) => {
     const gitUrlRegex =
       /(?:git|ssh|https?|git@[-\w.]+):(\/\/)?(.*?)(\.git)(\/?|\#[-\d\w._]+?)$/;
+
+    const standardUrlRegex = new RegExp(standardURLRegex);
+
     const containsURL = (string: string) =>
-      gitUrlRegex.test(string) || standardURLRegex.test(string);
+      gitUrlRegex.test(string) || standardUrlRegex.test(string);
```
ibolton336, line 145: Slightly confusing naming here with the only difference between standardUrlRegex and standardURLRegex being the case. Just a suggestion.

gildub: @ibolton336,
I haven't been to the bottom of it but I wish we could avoid creating a RegExp from a RegExp.
Naming is hard. 
What about "localStandardURLRegex" ?

ibolton336: Sounds like a move in the right direction. I liked the pattern of wrapping the regex test in a function and just calling that wherever we needed it instead of having to use .test every time.

ibolton336 (approved):

Issue #875, opened 2023-05-10
Finish migration waves application table


ibolton336:
- Wire up remaining fields
- Implement delete

![Image](https://user-images.githubusercontent.com/11218376/237513180-29e78b9f-5e2f-44cf-a61d-52f67a285241.png)

Issue #878, opened 2023-05-10
Add status polling functionality for export issues to issue manager


ibolton336:
- When applications are in the process of being imported to the issue manager, the aggregated status should be *Creating Issues* and the individual application status should be *Creating Issue*.

- Once issues have been created, each issue should display their corresponding status on the issue manager. The aggregated status will display *Issues Created* if the current system date is previous to the Start Date for the wave, and *In progress* if the current system date is posterior to it.

Issue #877, opened 2023-05-10
Update application drawer with migration wave status


ibolton336:
![Image](https://user-images.githubusercontent.com/11218376/237514224-1a9cd07e-d7ec-4ae8-a918-280566528426.png)

Issue #874, opened 2023-05-10
Finish migration waves stakeholder table data


ibolton336:
- Finish wiring up data
- Add delete stakeholder function

![Image](https://github.com/konveyor/tackle2-ui/assets/11218376/ce4121a7-aa33-416b-a928-3d43593d83b6)

PR #889, opened 2023-05-11
:sparkles: Migration wave tables latest and Jira Instance

gildub:
Migration Waves:
- Applications, Stakeholders and Tickets adjustable compound details 
- Activated trash can button to remove application from migration wave
- Follow hub changes now using `Ticket`'s parent.id and kind.id instead of names
- Rationalize/consolidate toastr.success/fail messages
- Move all components and forms under ./components
- Add "No Issues" for aggregated status to cover not documented case when migration wave has applications assigned but no tickets defined yet (not exported).

Jira Configuration:
- Enforce issue manager instance can't be delete when tickets exist with alert modal

Application form: 
- When updating application, added migrationwave field to preserve record otherwise application gets detached from migration wave.

Applications Inventory:
- Detail drawer includes either migration wave information

Queries/Model:
- The Id field of a resource exist for every resource record but at creation time.
  Removing it for being an optional field (`id?: number`) makes handling data much easier because there is no need to test for undefined across all resources. The only case where it's not needed is during creation time and this is where `Omit` is handy.
 Because the migration waves page is consolidating so many resources with Ref[] relationship it was obvious the Ref interface had to be changed.
Which impacted stakeholder, stakeholderGroup, tag, tagCategory changes and triggered a cascade of changes along the way including aligning the involved react query and rest queries.
The final solution (thanks to @mturley) is to use a generic type for all create REST function objects which opts out the id.
`export New<T extends {id: string"} = Omit<T, "id">` 

Resolves https://github.com/konveyor/tackle2-ui/issues/875, https://github.com/konveyor/tackle2-ui/issues/587, https://github.com/konveyor/tackle2-ui/issues/874, https://github.com/konveyor/tackle2-ui/issues/878, https://github.com/konveyor/tackle2-ui/issues/877

image

image

image

image


client/src/app/api/rest.ts
```diff
@@ -227,50 +161,20 @@ export const getJobFunctions = (): AxiosPromise<JobFunction[]> => {
   return APIClient.get(`${JOB_FUNCTIONS}`, jsonHeaders);
 };
 
-// Tag categories
-
-export const getTagCategories = (): AxiosPromise<Array<TagCategory>> => {
```
ibolton336, line 232: Can you move this work to a separate PR?

client/src/app/pages/controls/stakeholder-groups/components/stakeholder-group-form.tsx
```diff
@@ -109,6 +109,55 @@ export const StakeholderGroupForm: React.FC<StakeholderGroupFormProps> = ({
...
+    _: AxiosResponse<StakeholderGroup>
+  ) =>
+    pushNotification({
+      title: t("toastr.success.create", {
+        type: t("terms.stakeholderGroup"),
+      }),
+      variant: "success",
+    });
+
+  const onCreateStakeholderGroupError = (error: AxiosError) => {
```
ibolton336, line 122: Can you move all unrelated work to new PRs with a description & intention behind the changes? Thanks.

ibolton336 (changes requested):

client/src/app/api/rest.ts
gildub: @ibolton336, not really, working with `Ref` took me to make `id` not an option and this triggered a cascade of changes on the involved queries, which I aligned using axios as this is how we handle them.

client/src/app/pages/controls/stakeholder-groups/components/stakeholder-group-form.tsx
gildub: Same response

client/src/app/api/rest.ts
ibolton336: Seems like some good changes here. Thanks.

client/src/app/pages/applications/analysis-wizard/analysis-wizard.tsx
```diff
@@ -303,7 +303,7 @@ export const AnalysisWizard: React.FC<IAnalysisWizard> = ({
     { id, name },
     { prevId, prevName }
   ) => {
-    if (id && stepIdReached < id) setStepIdReached(id as number);
+    if (id && stepIdReached < (id as number)) setStepIdReached(id as number);
```
ibolton336, line 306: Why do we need this assertion?

gildub: Because Id is mandatory for that object anyway.

client/src/app/pages/controls/stakeholder-groups/components/stakeholder-group-form.tsx
ibolton336: Most of these changes are out of the scope of this PR. I know how it goes when you get rolling and want to make mass changes. But we should be looking to avoid this type of PR as a team. If we want to convert all of the rest.ts calls to remove the deprecated APIClient, then we should make a PR for that etc.

client/src/app/pages/applications/analysis-wizard/analysis-wizard.tsx
gildub: More likely for the same reason we need at the end of that line too.

client/src/app/pages/controls/stakeholder-groups/components/stakeholder-group-form.tsx
```diff
@@ -121,36 +170,22 @@ export const StakeholderGroupForm: React.FC<StakeholderGroupFormProps> = ({
         };
       });
 
-    const payload: StakeholderGroup = {
+    const payload: Omit<StakeholderGroup, "id"> = {
```
ibolton336, line 173: Why is this necessary?

client/src/app/pages/controls/stakeholders/stakeholders.tsx
```diff
@@ -61,8 +61,13 @@ export const Stakeholders: React.FC = () => {
...
 
-  const [isNewModalOpen, setIsNewModalOpen] = React.useState(false);
-  const [rowToUpdate, setRowToUpdate] = React.useState<Stakeholder>();
+  const [createUpdateModalState, setCreateUpdateModalState] = React.useState<
+    "create" | Stakeholder | null
+  >(null);
+  const isCreateUpdateModalOpen = createUpdateModalState !== null;
+  const stakeholderToUpdate =
+    createUpdateModalState !== "create" ? createUpdateModalState : null;
+
```
ibolton336, line 70: We should move the modal refactor work to a new PR and address it systematically if we want to do it.

client/src/app/pages/controls/stakeholder-groups/components/stakeholder-group-form.tsx
gildub: Because Id is always present but during creation.
This avoids to test for undefined Id when there is an ID.
I've updated description for more details

ibolton336: Thats a good idea.

client/src/app/pages/controls/stakeholders/stakeholders.tsx
gildub: The change with the Id, triggered a cascade of changes.
This is now aligned with our latest practice.
Sorry for the big PR :(

client/src/app/pages/controls/stakeholder-groups/components/stakeholder-group-form.tsx
gildub: I understand, and I agree with smaller PRs.
Again the cascade of change wasn't triggered to replace APIClient, but I made changes around then a aligned those.

client/src/app/pages/controls/stakeholder-groups/components/stakeholder-group-form.tsx
gildub: Thanks.
I didn't mean to make so many changes but once I touched the ID it was a long journey to fix everything around it.
There are still some resources using optional ID but those weren't impacted so we can leave them as tech debt.

client/src/app/pages/applications/analysis-wizard/analysis-wizard.tsx
ibolton336: Seems as though this is coming from patternfly allowing string or number for ids here.

client/src/app/pages/migration-waves/migration-waves.tsx
```diff
@@ -244,8 +244,13 @@ export const MigrationWaves: React.FC = () => {
     return statuses as TicketStatus[];
   };
 
-  const aggregateTicketStatus = (val: TicketStatus) => {
+  const aggregateTicketStatus = (val: TicketStatus, startDate: string) => {
     const status = ticketStatusToAggreaate.get(val);
+    if (status === "Issues Created") {
+      const now = new Date();
```
ibolton336, line 250: Lets try and use Dayjs here to be consistent with what was done in the wave create modal. Thx.

ibolton336 (approved):

client/src/app/pages/controls/stakeholders/components/stakeholder-form.tsx
```diff
@@ -132,7 +169,7 @@ export const StakeholderForm: React.FC<StakeholderFormProps> = ({
         };
       });
 
-    const payload: Stakeholder = {
+    const payload: Omit<Stakeholder, "id"> = {
```
ibolton336, line 172: Seems like this was a good thought, but no reason we should have to remember to omit an ID every place we define an instance type for submitting payloads or setting the types in the rest.ts file. Seems like we could think of something better here.

ibolton336 (changes requested):

client/src/app/pages/controls/stakeholders/components/stakeholder-form.tsx
ibolton336: What is wrong with leaving the id optional and conditionally merging it to the payload object when we are updating?

gildub: There are no need to worry about Ids on any object with this approach.
The reason is because resources always have an ID but at creation time.
Instead until now we've been looking at it from the other end where Ids are optional because we don't know the value at creation time.  
The benefit mean certainty when handling such objects and not having to test if the object is not "undefined".

mturley (approved):
LGTM after discussion on Slack.

client/src/app/pages/controls/stakeholders/components/stakeholder-form.tsx
ibolton336: Makes sense - thanks.

ibolton336 (approved):

PR #890, opened 2023-05-15
:seedling: Gracefully shutdown the server on SIGINT/SIGTERM

sjd78:
Add graceful shutdown handling for Ctrl-C (SIGINT) and the podman/docker container stop (SIGTERM) signals.  There are a few advantages doing this:

  1. Allow connections to terminate, and

  2. Allow for any resource cleanup, and

  3. Allow Ctrl-C (and `podman container stop` without resorting to SIGKILL) to actually work when running the server as PID 1 in the container.

For some graceful shutdown of node/express as PID 1 in a container references, see:
  - https://expressjs.com/en/advanced/healthcheck-graceful-shutdown.html
  - https://github.com/nodejs/docker-node/blob/main/docs/BestPractices.md#handling-kernel-signals
  - https://emmer.dev/blog/you-don-t-need-an-init-system-for-node.js-in-docker/#express-doesn-t-trap-exit-signals


gildub (approved):
Looks and sounds good !

PR #891, opened 2023-05-16
:warning: Convert labels to source/target

ibolton336:
- Move from xml to yaml for rule file upload - Adds js-yaml parser 
- Add support for labels parsed from the uploaded rule files
- Adds data transformation to rulebundles.ts to support UI components source/target model for label viewing. 
TODO: 
- pass through all labels that don't match source/target syntax to analyzer
- Output is no longer used by tasks. Need to figure out a way to show reports / download links if we plan to continue support of that feature. 
- Remove all CSV download / html download global settings


mturley (approved):
LGTM

PR #895, opened 2023-05-19
:seedling: Refactor server to be an ESM package and use rollup

sjd78:
### Change highlights:
  - Converted the server package to a `type: "module"` ESM module[0]

  - Use rollup[1] in the server package to build to a single bundle
    that can run without external libraries.  Now, `node_modules/` is
    no longer needed to run the built server.

  - Script `npm run dist` collects all of the built artifacts needed
    to run the UI in container

  - Moved envvar `server/helpers.js` from the server package to the
    client package `client/config/envLookup.js`.  The server package
    uses monorepo package link to reference the file in the client
    package.  `webpack.dev.ts` does not work accessing sources from
    the server package with the ESM changes, but server works
    pulling from client/config.

  - Container build is simplified
 
  - Under `npm run start:dev`, the both the client and server will
    auto-reload when source changes are detected

### A number of good things come from using rollup and ESM:
  - `node_modules` is not needed at run time

  - The built server code size is further reduced through tree-shaking

  - The server code looks more similar to the client TS code by using
    `import`, `export`, and `export default` statements

### Container build changes:
  - Added a `.dockerignore` to minimize the number of files copied to
    the build container.  This allows the container to do a fresh
    `npm clean-install` and a fresh `npm build`, even when building the
    container from a developer's repo.  The builds are more repeatable,
    with the same set of files being added from a fresh repo pull or
    from a developer's repo.

  - The builder uses the `default` user to run the build

  - Use `npm clean-install` in place of `npm install` to make sure all
    of the versions are coming from `package-lock.json`

  - Only the project root `dist/` folder needs to be copied to the run
    container since the `npm run dist` script collects all production
    built artifacts needed to run the app.

  - The run container paths adjusted from `/src` to `/dist` to help
    indicate it contains built files and not the sources

  - `entrypoint.sh` node command adjusted for the new paths

### References
[0] - https://nodejs.org/api/esm.html#introduction
[1] - https://rollupjs.org/


sjd78:
Moved `npm run test` and Github Actions ci refactoring to #897

This PR still included the commit since it is needed to pass ci.

sjd78:
Moved npm script refactoring to #898

ibolton336 (approved):
Excellent stuff! Thank you.

gildub (approved):
Looks good.
Thanks for cleaning up and introducing Rollup.js !

PR #897, opened 2023-05-19
:seedling: Add `npm run test` and update github action `ci-actions.yml` to use it

sjd78:
Add a `npm run test` script to the root package and adjust the Github Action to run from the root directory.  This allows for proper installation of any shared dependencies in the root `package.json`, and for test scripts to be run in each workspace.

**Note**: The cli options given in `ci-actions.yml` for the test script were not actually being passed to the jest command.  An extra `--` is needed to send those options.  Those adjustments have also been made.

**Note 2**: Using `npm clean-install` in the `ci-actions.yml` will cause ci tests to fail if `package-lock.json` has not been updated in a PR. This version of install prevents npm from updating the lock file.

**Note 3**: `.gitignore` updated for non-root test, build and dist paths

Extracted from #895

PR #898, opened 2023-05-19
:seedling: Refactor npm scripts, prefer coordination from root package

sjd78:
Change highlights:

  - Script `npm run build:dev` is dropped since it doesn't work. The
    current `_env` injection configurations in `webpack.dev.ts` is not
    compatible with serving the app from the express server.

  - Script `npm run start` builds the production client and starts the
    server, mimicking how the UI runs in container

  - Script `npm run start:dev` is managed at the root `package.json`

  - Script `npm run port-forward` is managed at the root `package.json`

The scripts moved to the root `package.json` allows for the reduction
of the scope of concern for each packages' scripts.  Now, the client only
knows how to run itself.  The root knows how to coordinate the various
pieces (the port mapping, the client's webpack-dev-server, and the
server's server) to get everything running.

~~Requires (and currently includes) #897~~

Extracted from #895


ibolton336 (approved):

mturley:
Love it

Issue #876, opened 2023-05-10
Manage applications view - edit support


ibolton336:
![Image](https://user-images.githubusercontent.com/11218376/237513763-091fb329-09cd-4127-9f7b-f3c10587042f.png)

Manage applications view should reflect currently associated applications

PR #902, opened 2023-05-22
:sparkles: ManageApplicationsForm: Set initialSelected items

gildub:
Resolves https://github.com/konveyor/tackle2-ui/issues/876

### DIFF
--- a/client/src/app/pages/migration-waves/components/manage-applications-form.tsx
+++ b/client/src/app/pages/migration-waves/components/manage-applications-form.tsx
@@ -14,10 +14,9 @@ import {
   ToolbarItem,
 } from "@patternfly/react-core";
 
-import { MigrationWave } from "@app/api/models";
+import { Application, MigrationWave } from "@app/api/models";
 import { ToolbarBulkSelector } from "@app/shared/components";
 import { NotificationsContext } from "@app/shared/notifications-context";
-import { useFetchApplications } from "@app/queries/applications";
 import { useLocalTableControls } from "@app/shared/hooks/table-controls";
 import {
   FilterToolbar,
@@ -41,12 +40,14 @@ import { dedupeFunction } from "@app/utils/utils";
 import { useUpdateMigrationWaveMutation } from "@app/queries/migration-waves";
 
 export interface ManageApplicationsFormProps {
+  applications: Application[];
   migrationWave: MigrationWave;
   migrationWaves: MigrationWave[];
   onClose: () => void;
 }
 
 export const ManageApplicationsForm: React.FC<ManageApplicationsFormProps> = ({
+  applications,
   migrationWave,
   migrationWaves,
   onClose,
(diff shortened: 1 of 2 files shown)


client/src/app/pages/migration-waves/components/manage-applications-form.tsx
```diff
@@ -41,30 +40,47 @@ import { dedupeFunction } from "@app/utils/utils";
...
   migrationWaves,
   onClose,
 }) => {
   const { t } = useTranslation();
   const { pushNotification } = React.useContext(NotificationsContext);
 
-  const { data: applications, isFetching } = useFetchApplications();
-
-  const assignedApplications = migrationWaves
+  const applicationsUsedByOtherMigrationsIds = migrationWaves
```
mturley, line 58: Gotta love a descriptive name 👍 💯

client/src/app/pages/migration-waves/components/manage-applications-form.tsx
```diff
@@ -41,30 +40,47 @@ import { dedupeFunction } from "@app/utils/utils";
...
   const availableApplications = applications.filter(
-    (application) => !assignedApplications.includes(application.id)
+    (application) =>
+      !applicationsUsedByOtherMigrationsIds.includes(application.id)
   );
 
+  const isNewerSelection = (): boolean => {
+    const selectedIds = selectedItems.map((application) => application.id);
+    return (
+      selectedIds.sort().toString() !==
```
mturley, line 79: This is clever! This might be a performance issue in cases where there are hundreds of apps selected, but that doesn't seem likely here right? So this will work fine.

If you want though, here's another solution that will return early as soon as it detects a difference while looping through the arrays. We could export a util for this:
```ts
export const isArrayDifferent = <T>(a: T[], b: T[]) =>
  a.some((val) => !b.includes(val)) || b.some((val) => !a.includes(val));
```

and then the usage here would be
```ts
return isArrayDifferent(selectedIds, initialApplicationUsedByMigrationIds);
```

Whichever you like better is fine, just thinking out loud. I'm actually not sure if this will perform better anyway in the average case. And I'm definitely overthinking it.

mturley (approved):
LGTM. I have one note and then realized I was overthinking, so feel free to ignore it.

client/src/app/pages/migration-waves/components/manage-applications-form.tsx
gildub: @mturley, yeah I took the shortcut for a quick implementation but for scalability the array inspection is better.
So let me take that one in.

gildub: @mturley, I did some tests and the string size blows up before the stack so that's the first limitation of my initial approach. The inspection approach wins anyway being about 140 times faster!

Issue #858, opened 2023-05-08
Jira credential username validation


ibolton336:
For Jira credentials, the username needs to be an email address when the instance type is "jira cloud" We need to handle this in the UI. Connections to Jira trackers will fail silently when a credential is used with the wrong username field. 

`{"level":"error","ts":1683308267.8212721,"logger":"tickets","msg":"Connection test failed.","tracker":1,"error":"401 Unauthorized: <?xml version=\"1.0\" encoding=\"UTF-8\" standalone=\"yes\"?><status><status-code>401</status-code><message>Client must be authenticated to access this resource.</message></status>: request failed. Please analyze the request body for more details. Status code: 401","errorVerbose":"request failed. Please analyze the request body for more details. Status code: 401\n401 Unauthorized: <?xml version=\"1.0\" encoding=\"UTF-8\" standalone=\"yes\"?><status><status-code>401</status-code><message>Client must be authenticated to access this resource.</message></status>","stacktrace":"github.com/go-logr/zapr.(*zapLogger).Error\n\t/Users/ibolton/code/go/pkg/mod/github.com/go-logr/zapr@v0.3.0/zapr.go:132\ngithub.com/konveyor/controller/pkg/logging.(*Logger).Error\n\t/Users/ibolton/code/go/pkg/mod/github.com/konveyor/controller@v0.8.0/pkg/logging/logger.go:109\ngithub.com/konveyor/tackle2-hub/tracker.(*Manager).testConnection\n\t/Users/ibolton/code/go/src/github.com/konveyor/tackle2-hub/tracker/manager.go:85\ngithub.com/konveyor/tackle2-hub/tracker.(*Manager).testConnections\n\t/Users/ibolton/code/go/src/github.com/konveyor/tackle2-hub/tracker/manager.go:67\ngithub.com/konveyor/tackle2-hub/tracker.(*Manager).Run.func1\n\t/Users/ibolton/code/go/src/github.com/konveyor/tackle2-hub/tracker/manager.go:42"}`

PR #916, opened 2023-05-23
:sparkles: IdentityForm: Username validation

gildub:
Simplifies create/update modal
Jira username requires valid email

Resolves https://github.com/konveyor/tackle2-ui/issues/858


ibolton336:
Nice tech debt fixes here!

client/src/app/pages/identities/identities.tsx
```diff
@@ -173,35 +171,9 @@ export const Identities: React.FC = () => {
...
-    }
-
-    closeIdentityModal();
-  };
-
-  const handleOnIdentityUpdated = () => {
-    setRowToUpdate(undefined);
+  const closeCreateUpdateModal = () => {
+    setCreateUpdateModalState(null);
+    refetch;
```
ibolton336, line 176: Is this supposed to be called?

ibolton336 (changes requested):

client/src/app/pages/identities/identities.tsx
```diff
@@ -317,16 +292,27 @@ export const Identities: React.FC = () => {
...
+                })
+          }
+          variant={ModalVariant.medium}
+          isOpen={isCreateUpdateModalOpen}
+          onClose={closeCreateUpdateModal}
+        >
+          <IdentityForm
+            identity={identityToUpdate ? identityToUpdate : undefined}
+            onClose={closeCreateUpdateModal}
+            xmlValidator={validateXML}
```
ibolton336, line 313: Not needed

client/src/app/pages/identities/identities.tsx
```diff
@@ -47,6 +46,8 @@ import {
 } from "@app/queries/identities";
 import { useFetchApplications } from "@app/queries/applications";
 import { NotificationsContext } from "@app/shared/notifications-context";
+import { IdentityForm } from "./components/identity-form";
+import { validateXML } from "./components/identity-form/validateXML";
```
ibolton336, line 50: Not needed

ibolton336 (changes requested):

ibolton336 (changes requested):
See inline.

client/src/app/pages/identities/identities.tsx
gildub: @ibolton336, normally we don't need it with `invalidateQueries()`
Let me double check...

ibolton336: Cool - just saying if we do it needs perens 
```suggestion
    refetch();
```

client/src/app/pages/identities/identities.tsx
ibolton336: I am sorry - we may still need this. Need to check with Jeff/Ramon here about support for mvn identities for mta 7. I Know we will need it for v0.2.0

client/src/app/pages/identities/identities.tsx
ibolton336: I am sorry - we may still need this. Need to check with Jeff/Ramon here about support for mvn identities for mta 7. I Know we will need it for v0.2.0

client/src/app/pages/identities/identities.tsx
gildub: @ibolton336, this is out of scope for this PR. I haven't changed anything from previous modal. So let's open another issue if needed.

client/src/app/pages/identities/identities.tsx
gildub: @ibolton336, redundant, I removed it.

ibolton336:
@gildub Did we confirm we want to constrain the jira cred username to an email?

ibolton336 (changes requested):
One question re: email vs string for jira instance creds.

gildub:
@ibolton336, yes Jira auth is based on Jira email for associated account.

ibolton336:
@gildub If it is always email, we should change the field name to email.

gildub:
@ibolton336, well Jira users know a username is an email address.
It's like most of user accounts on the web, isn't ?
That said we could display `"Username (email)"`.
Not a blocker

ibolton336:
> @ibolton336, well Jira users know a username is an email address. It's like most of user accounts on the web, isn't ? That said we could display `"Username (email)"`. Not a blocker

Last I checked, the username differed based on the type of JIRA instance used. Is that no longer the case? The credential will fail if they use their jira username so I would think we would want to let the user know that the email is required rather than the jira user name.

ibolton336:
> @ibolton336, well Jira users know a username is an email address. It's like most of user accounts on the web, isn't ? That said we could display `"Username (email)"`. Not a blocker

Looks like github accepts either username or email. So if JIRA operates this way, I would think something needs to change with how we display the field. 


Screenshot 2023-05-24 at 3 07 29 PM

gildub:
Not sure about earlier generalization I made as this is semantic.

Meanwhile I based my approach from Jira (documentation says)[https://support.atlassian.com/atlassian-account/docs/manage-api-tokens-for-your-atlassian-account/] says about creating an API token: "You can make HTTP basic authentication requests with your email address as the username and this API token as the password." 

I'm happy to use either "Username" or "Email" or both. We're likely need validation from UXD.

client/src/app/pages/identities/components/identity-form/identity-form.tsx
```diff
@@ -645,7 +645,7 @@ export const IdentityForm: React.FC<IdentityFormProps> = ({
           <HookFormPFTextInput
             control={control}
             name="user"
-            label="Username"
+            label={values.kind === "jira" ? "Username's email" : "Username"}
```
ibolton336, line 648: Per @justinxhale from UXD:

"Jira username or email”

Issue #913, opened 2023-05-22
Bulk delete for migration waves shows wrong success message


ibolton336:
Should show all wave names removed insead of only 1 name. Also I think we are showing a confirmation dialog when bulk deleting in other areas of the app.

PR #711, opened 2023-03-14
:sparkles: Merge `feature-migration-waves` to `main` (disabled behind feature flag)

mturley:
This PR is a placeholder to remind us that we have an ongoing feature branch in `feature-migration-waves` (started in #598). Other PRs will be merged to that branch, and then we can rebase and merge this PR when we're ready to put all those changes into the main branch.

Changes included here are from:
* #598
* #776 
* #740
* #778
* #790

See notes in https://github.com/konveyor/tackle2-ui/pull/740 for some next steps.

Edit: we've added a feature flag in #790 so we can merge this to main even though it's incomplete.

PR #917, opened 2023-05-23
:sparkles: Migration Waves: refactor bulk delete

gildub:
Wraps individual migration waves delete for bulk delete

Resolves https://github.com/konveyor/tackle2-ui/issues/913

### DIFF
--- a/client/src/app/pages/migration-waves/migration-waves.tsx
+++ b/client/src/app/pages/migration-waves/migration-waves.tsx
@@ -27,6 +27,7 @@ import {
 import {
   AppPlaceholder,
   ConditionalRender,
+  ConfirmDialog,
   KebabDropdown,
   ToolbarBulkSelector,
 } from "@app/shared/components";
@@ -107,15 +108,22 @@ export const MigrationWaves: React.FC = () => {
   const openCreateWaveModal = () => setWaveModalState("create");
 
   const [exportIssueModalOpen, setExportIssueModalOpen] = React.useState(false);
-
   const [applicationsToExport, setApplicationsToExport] = React.useState<Ref[]>(
     []
   );
 
   const [waveToManageModalState, setWaveToManageModalState] =
     React.useState<MigrationWave | null>(null);
-
   const closeWaveModal = () => setWaveModalState(null);
+
+  const [migrationWaveToDelete, setMigrationWaveToDelete] = React.useState<{
+    id: number;
+    name: string;
+  } | null>(null);
+
+  const [migrationWavesToDelete, setMigrationWavesToDelete] = React.useState<
+    number[] | null
+  >(null);
 
   const getApplications = (refs: Ref[]) => {
     const ids = refs.map((ref) => ref.id);
@@ -126,15 +134,15 @@ export const MigrationWaves: React.FC = () => {
 
   const onDeleteWaveSuccess = (name: string) => {
     pushNotification({
-      title: t("toastr.success.deleteds", {
+      title: t("toastr.success.deleted", {
         what: name,
-        type: t("terms.migrationWaves"),
+        type: t("terms.migrationWave"),
       }),
       variant: "success",
     });
   };
 
-  const onDeleteWaveError = (error: AxiosError) => {
+  const onError = (error: AxiosError) => {
     pushNotification({
       title: getAxiosErrorMessage(error),
       variant: "danger",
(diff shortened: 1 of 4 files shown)


client/src/app/pages/migration-waves/migration-waves.tsx
```diff
@@ -137,14 +139,37 @@ export const MigrationWaves: React.FC = () => {
...
   };
 
   const { mutate: deleteWave } = useDeleteMigrationWaveMutation(
     onDeleteWaveSuccess,
     onDeleteWaveError
   );
 
+  const onDeleteAllMigrationWavesSuccess = (response: any) => {
+    pushNotification({
+      title: t("toastr.success.deleted", {
```
ibolton336, line 151: Success message currently should read 

Migration wave(s) was successfully deleted.

or if multiple: 

Migration waves were successfully deleted.

ibolton336 (changes requested):

client/src/app/pages/migration-waves/migration-waves.tsx
```diff
@@ -400,14 +425,20 @@ export const MigrationWaves: React.FC = () => {
                             </DropdownItem>,
                             <DropdownItem
                               key="bulk-delete"
+                              isDisabled={selectedItems.length === 0}
                               onClick={() => {
+                                let deleteMigrationWaves: Promise<MigrationWave>[] =
+                                  [];
                                 selectedItems.map((migrationWave) => {
```
ibolton336, line 432: We can use the spread operator here:
```suggestion
                                selectedItems.map((migrationWave) => {
                                  return [
                                    ...deleteMigrationWaves,
                                    migrationWave.id &&
                                      deleteMigrationWave(migrationWave.id),
                                  ];
                                });
```

client/src/app/pages/migration-waves/migration-waves.tsx
```diff
@@ -400,14 +425,20 @@ export const MigrationWaves: React.FC = () => {
...
-                                      id: migrationWave.id,
-                                      name: migrationWave.name,
-                                    });
+                                  if (migrationWave.id) {
+                                    let promise: Promise<MigrationWave>;
+                                    promise = deleteMigrationWave(
+                                      migrationWave.id
+                                    );
+                                    deleteMigrationWaves.push(promise);
+                                  }
```
ibolton336, line 439:
```suggestion
                  
```

client/src/app/pages/migration-waves/migration-waves.tsx
```diff
@@ -400,14 +425,20 @@ export const MigrationWaves: React.FC = () => {
                             </DropdownItem>,
                             <DropdownItem
                               key="bulk-delete"
+                              isDisabled={selectedItems.length === 0}
                               onClick={() => {
+                                let deleteMigrationWaves: Promise<MigrationWave>[] =
+                                  [];
```
ibolton336, line 431:
```suggestion
                                const deleteMigrationWaves: Promise<MigrationWave>[] =
                                  [];
```
const is fine here since we do not plan to reassign references to the object.

client/src/app/pages/migration-waves/migration-waves.tsx
ibolton336: What do we do in the case of only 1 wave slated for delete? The current form would lead to the wrong message displayed. 😅

client/src/app/pages/migration-waves/migration-waves.tsx
```diff
@@ -124,9 +126,9 @@ export const MigrationWaves: React.FC = () => {
 
   const onDeleteWaveSuccess = (name: string) => {
     pushNotification({
-      title: t("toastr.success.deleted", {
+      title: t("toastr.success.deleteds", {
```
ibolton336, line 129: maybe `multipleDelete`? 🤷

ibolton336:
Should we add a confirm dialog to be consistent?
Screenshot 2023-05-23 at 3 08 21 PM

client/src/app/pages/migration-waves/migration-waves.tsx
```diff
@@ -105,16 +108,23 @@ export const MigrationWaves: React.FC = () => {
...
 
+  const [migrationWaveToDelete, setMigrationWaveToDelete] = React.useState<{
+    id: number;
+    name: string;
+  } | null>(null);
+
+  const [migrationWavesToDelete, setMigrationWavesToDelete] = React.useState<
+    number[] | null
+  >(null);
+
```
ibolton336, line 127: I think we should avoid using useState here but I see why you did it for the confirm modal. Hmm..

ibolton336: This is probably fine. Just wondering if there is a better way rather than cluttering our components up. Can revisit later.

ibolton336 (approved):

client/src/app/pages/migration-waves/migration-waves.tsx
gildub: @ibolton336, 

We have to use state because we capture the item to delete which will be deleted after Confirmation Dialog.
Combining the state and flag was introduced by @mturley (https://github.com/konveyor/tackle2-ui/pull/711) and I think it helps simplifying.

ibolton336: Yeah I understand that. But now we have duplicated some state here it appears.

gildub: @ibolton336, hmmm, we have state for exporting, managing, editing and deleting, let me harmonize the names and align them.

ibolton336: Ok. I think you could drop the duplicate wave / waves state and just use one slice of state for that. Not sure I follow on your other point.

PR #919, opened 2023-05-23
:seedling: Fix container build for old versions of docker

sjd78:
With docker version `Docker version 20.10.22, build 3a2c30b`, the build fails not knowing how to map a username to a uid in the `COPY` command.

Do two things to prevent this error, and potentially others, going forward:

  - explicitly use the `:latest` tag on the FROM containers

  - explicitly use `USER 1001` and `COPY --chown:1001` to avoid any user name to uid lookup problems


gildub (approved):

ibolton336:
Thanks @sjd78!

Issue #914, opened 2023-05-22
Should not be able to export a wave with no associated applications


ibolton336:
Disable the export buttons when no selection is made (for bulk) or when no applications are assigned to migration waves.

PR #922, opened 2023-05-24
:sparkles: Export only when selection and applications are available

gildub:
Note. This is not about refactoring bulk ticket creation with newly available HUB API request which will be done separately.

Resolves https://github.com/konveyor/tackle2-ui/issues/914

### DIFF
--- a/client/src/app/pages/migration-waves/migration-waves.tsx
+++ b/client/src/app/pages/migration-waves/migration-waves.tsx
@@ -73,7 +73,6 @@ import { WaveStatusTable } from "./components/status-table";
 import { WaveForm } from "./components/migration-wave-form";
 import { ManageApplicationsForm } from "./components/manage-applications-form";
 import { useFetchStakeholders } from "@app/queries/stakeholders";
-import { useFetchStakeholderGroups } from "@app/queries/stakeholdergoups";
 import { deleteMigrationWave } from "@app/api/rest";
 dayjs.extend(utc);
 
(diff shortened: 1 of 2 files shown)


ibolton336 (approved):

client/src/app/pages/migration-waves/migration-waves.tsx
```diff
@@ -104,11 +104,9 @@ export const MigrationWaves: React.FC = () => {
     migrationWaveModalState !== "create" ? migrationWaveModalState : null;
   const openCreateWaveModal = () => setWaveModalState("create");
 
-  const [exportIssueModalOpen, setExportIssueModalOpen] = React.useState(false);
-
-  const [applicationsToExport, setApplicationsToExport] = React.useState<Ref[]>(
-    []
-  );
+  const [applicationsToExport, setApplicationsToExport] = React.useState<
```
ibolton336, line 107: No reason to use react state here.

gildub: Using only that state allowing `null` value simplifies modal handling

gildub:
@ibolton336, the refactor is to use one state (allowing `null`) variable to handle modal.

client/src/app/pages/migration-waves/migration-waves.tsx
ibolton336: Ok. Can we do that without using react state? Maybe we can use a variable instead.

ibolton336 (approved):

Issue #921, opened 2023-05-23
Delete application is currently broken


ibolton336:
FOREIGN KEY constraint failed

ibolton336:
Update only happens with apps created from cli

gildub:
@ibolton336 applications assigned to migration waves cannot be delete.
So we need to disable the button and display an alert message.

PR #923, opened 2023-05-24
:sparkles: Applications assigned to migration waves cannot be deleted

gildub:
Resolves https://github.com/konveyor/tackle2-ui/issues/921


client/src/app/api/models.ts
```diff
@@ -118,7 +118,7 @@ export interface Application {
   identities?: Ref[];
   repository?: Repository;
   binary?: string;
-  migrationWave?: Ref;
+  migrationWave: Ref | null;
```
ibolton336, line 121: Why the change to null here?

gildub: Because migrationWave is a field that exists in the record meanwhile its value can be `null`.

ibolton336: Undefined typically means a variable has been declared but not defined yet. This aligns with how the hub uses our payloads that we pass. We don't pass the hub "null" values for fields that do not exist. We simply omit the field when making the request. Adding explicit null values is useful when we need a special user defined null case that sits outside of our normal usage.

ibolton336: Take a look at the payload we send to the hub 

image

ibolton336:
```suggestion

export interface Application {
  id: number;
  name: string;
  description?: string;
  comments?: string;
  businessService?: Ref;
  tags?: TagRef[];
  owner?: Ref;
  contributors?: Ref[];
  review?: Ref;
  identities?: Ref[];
  repository?: Repository;
  binary?: string;
  migrationWave?: Ref;
}
```

ibolton336: How the hub stores the values in the database doesn't really matter to us. We are working in TS world where the creators/maintaners  themselves have recommended  usage "undefined" and not use "null". I personally find that changing our pattern to "null" at this stage doesn't make much sense as both are acceptable and can be valid. There is no real strong point one way or another but I enjoy using the '?' syntax as I have come to find it makes the code more readable.

gildub: Here is the reason optional fields are only convenient at creation time:

If an application is updated (i.e AppicationForm) and the `migrationWave` field is omitted then the relationship is silently severed and the application disappears from the migration wave:

```json
{
	"binary": "::",
	"comments": "",
	"contributors": [],
	"description": "App1",
	"id": 1,
	"migrationWave": {
		"id": 3,
		"name": "Wave1"
	},
	"name": "App1",
	"tags": []
}
```

gildub: So maybe we're in a special case where `migrationWave` is a mandatory field.
The rest might be a question of view but I'll tend to make any recorded field not optional and avoid testing all the time if such field is defined and focus on the job to know if what value it contains or not.

ibolton336: I  think I see your point. So instead of having to add 

```
       migrationWave: application?.migrationWave,
````
to the update payload when updating an application from the form that has no migrationWave field included ( it should already exist on the application from other views. just like we have other fields not in the application edit form (review, bucket, etc)), we can do 

````
    if (application) {
      updateApplication({ ...application, ...payload });
    } else {
      createApplication(payload);
    }
````
to make sure we are not stripping any data from the application object by not including it in the update request.

ibolton336: @gildub Yes that is what they decided on for their project. 


https://github.com/Microsoft/TypeScript/wiki/Coding-guidelines#null-and-undefined

ibolton336 (approved):

PR #926, opened 2023-05-24
:ghost: Change references of "instance" to "tracker"

ibolton336:
- Use the same name ("trackers") for all of the UI components (forms/tables) & underlying architecture (hooks, queries/ etc) dealing with JIRA trackers. 
- Keeps usage of instance in form fields displayed to user and other "copy".

### DIFF
--- a/client/src/app/pages/external/jira/tracker-form.tsx
+++ b/client/src/app/pages/external/jira/tracker-form.tsx
@@ -43,13 +43,13 @@ interface FormValues {
   insecure: boolean;
 }
 
-export interface InstanceFormProps {
-  instance?: Tracker;
+export interface TrackerFormProps {
+  tracker?: Tracker;
   onClose: () => void;
 }
 
-export const InstanceForm: React.FC<InstanceFormProps> = ({
-  instance,
+export const TrackerForm: React.FC<TrackerFormProps> = ({
+  tracker,
   onClose,
 }) => {
   const { t } = useTranslation();
@@ -57,12 +57,12 @@ export const InstanceForm: React.FC<InstanceFormProps> = ({
   const [axiosError, setAxiosError] = useState<AxiosError>();
   const [isLoading, setIsLoading] = useState(false);
 
-  const { trackers: instances } = useFetchTrackers();
+  const { trackers: trackers } = useFetchTrackers();
   const { identities } = useFetchIdentities();
 
   const { pushNotification } = React.useContext(NotificationsContext);
 
-  const onCreateInstanceSuccess = (_: AxiosResponse<Tracker>) =>
+  const onCreateTrackerSuccess = (_: AxiosResponse<Tracker>) =>
     pushNotification({
       title: t("toastr.success.save", {
         type: t("terms.instance").toLocaleLowerCase(),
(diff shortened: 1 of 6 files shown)


client/src/app/shared/hooks/useCreateEditModalState.ts
```diff
@@ -0,0 +1,21 @@
+import React from "react";
+import { useContext, useState } from "react";
+
+export default function useCreateEditModalState<T>() {
+  const [modalState, setModalState] = React.useState<"create" | T | null>(null);
+  const isModalOpen = modalState !== null;
+  const resourceToUpdate = modalState !== "create" ? modalState : null;
+
+  const [resourceToDelete, setResourceToDelete] = React.useState<T | null>(
```
gildub, line 9: The delete state doesn't belong here. Not part of CreateEdit.

gildub (changes requested):
Yes "instance" is user view where "tracker" is technical view.
We have to maintain both views.
The interface belongs more on the UI interface.
The tracker is on the record side.
So this change sounds extreme, unnecessary and absolutely not priority. 
Also note I was aware of those views when I coded and I managed to use instance in a way that made sens. That said it's ok to replace either one if that make sense, in its context.

Let's move the hook code to another PR as it's out of scope.
From practice I think a hook such as `useCreateEditModal` does help remove some clutter  and bring more consistency.
This is why I'm using more `modalState`.

Although I realized the following icing sugar is actually, I think, cluttering the code unnecessarily: 
```
const isModalOpen = modalState !== null;
 const resourceToUpdate = modalState !== "create" ? modalState : null;
``` 
Because we test once `isModalOpen` so we can just test the value (null).
And the resource value is used once or twice so we can just use the value.

Therefore for best of both worlds we don't need a hook and use state which takes null or <T> or "create" or "whatever-we-need" and we test its value.

For createEdit:
`const [createEditModalState, setCreateEditModalState] = React.useState<"create" | T | null>(null);`

For delete (or some other action):  
`const [deleteModalState, setDeleteModalState] = React.useState<| T | null>(null);`

client/src/app/shared/hooks/useCreateEditModalState.ts
ibolton336: > Yes "instance" is user view where "tracker" is technical view. We have to maintain both views. The interface belongs more on the UI interface. The tracker is on the record side.

The only thing we need to maintain is the translation to user for form labels, descriptions, and column names. Makes more sense to use the same name for all of the UI components (forms/tables)  & underlying architecture (hooks, queries/ etc). Not sure what is so extreme about the change. Just helps when reasoning about the code.

ibolton336:
>Yes "instance" is user view where "tracker" is technical view. We have to maintain both views. The interface belongs more on the UI interface. The tracker is on the record side.

The only thing we need to maintain is the translation to user for form labels, descriptions, and column names. Makes more sense to use the same name for all of the UI components (forms/tables) & underlying architecture (hooks, queries/ etc). Not sure what is so extreme about the change. Just helps when reasoning about the code.

gildub:
@ibolton336, I understand your point and I agree about being consistent. 
Sometimes we need to use different name for the variety of the words.

Can we keep the hook separate though ?

gildub (approved):
LGTM

PR #930, opened 2023-05-25
:ghost: Support both xml and yaml files for custom rules

ibolton336:
- Adds support for XML rule files alongside yaml rule files. Conditionally reinstates the xml validation based on rule type
- Adds label parser function for source/target label display purposes in the wizard


client/src/app/common/CustomRules/rules-utils.tsx
```diff
@@ -1,25 +1,82 @@
 import { IReadFile, ParsedRule, Rule, Ruleset } from "@app/api/models";
 import yaml from "js-yaml";
+type RuleFileType = "YAML" | "XML" | null;
+export const checkRuleFileType = (filename: string): RuleFileType => {
+  const fileExtension = filename.split(".").pop();
```
mturley, line 5: I wonder if we want to do `filename.toLowerCase().split(".").pop();` here in case someone has capital `.YAML`/`.YML`/`.XML`. Not that that's common

client/src/app/common/CustomRules/rules-utils.tsx
```diff
@@ -1,25 +1,82 @@
...
-      target: allLabels?.targetLabel,
-      otherLabels: allLabels?.otherLabels,
-      allLabels: allLabels?.allLabels,
-      total: 0,
-      ...(file.responseID && {
-        fileID: file.responseID,
-      }),
-    };
+    if (checkRuleFileType(file.fileName) === "YAML") {
+      const payload = atob(file.data.substring(31));
```
mturley, line 16: Where do these magic numbers come from?

mturley (approved):
LGTM if you want to merge, one minor comment that may not matter and a question though.

client/src/app/common/CustomRules/rules-utils.tsx
ibolton336: Was reading the file in wrong. Fixed!

gildub (approved):
LGTM

PR #937, opened 2023-05-25
:seedling: [backport release-0.2] downgrade `rimraf` to `@^4.x` for downstream builds

sjd78:
Downstream builds use an npmjs package caching system called cachito[1].  It currently does not support dependencies that use a package-spec alias[2][3].  Downgrading `rimraf` to version `^4.x` removed any aliases from our `package-lock.json` and should allow downstream builds to work again.

[1] - https://github.com/containerbuildsystem/cachito
[2] - https://docs.npmjs.com/cli/v9/using-npm/package-spec?v=true#aliases
[3] - https://github.com/containerbuildsystem/cachito/issues/859

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -15,7 +15,7 @@
       "devDependencies": {
         "concurrently": "^8.0.1",
         "copyfiles": "^2.4.1",
-        "rimraf": "^5.0.1"
+        "rimraf": "^4.4.1"
       }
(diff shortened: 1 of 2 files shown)


rayfordj:
This pleases Cachito.  :+1:

ibolton336 (approved):

PR #966, opened 2023-06-01
:sparkles: Issues: Add affected file details modal with incident details and code snippet viewer, adjust for hub change from IssueReport to AppReport

mturley:
Initial implementation of the `FileIncidentsDetailModal`, which shows incident details within a file. Some details are incomplete pending hub implementation and design review, and will be finished outside the scope of this PR.

Screenshot 2023-06-07 at 3 01 24 PM

Screenshot 2023-06-07 at 3 01 30 PM

Full flow to reach this modal:
* On the Issues page, click the number of affected applications for an issue (RuleReport)
* On the Affected Applications page, click an application (AppReport) to open the drawer
* In the drawer's Affected Files tab, click a file (FileReport)

qi904nHhzs

This PR:

* Replaces the `AnalysisIssueReport` with `AnalysisAppReport` to match new hub changes (see https://github.com/konveyor/tackle2-hub/pull/368).
  * The AppReport object's top level properties are Application properties, and issue properties are now nested under `issue`. This aligns better with the "Affected applications" page we are using these objects for and better supports the sorting/filtering on that page.
  * Changes the hub path `/analyses/report/issues` to `/analyses/report/applications` to match.
  * Changes the Affected Applications page / table / drawer code to use the new type and naming
  * Adds `incidents` and `effort` columns to the Affected Applications table
  * Re-enables sorting on this table now that it is working in the hub
  * Removes the logic to manually cross-reference an IssueReport with an Application, because now the AppReport contains the application properties we need.
* Adds the `AnalysisIncident` type (corresponds to the hub type `Incident`).
  * This is the actual incident from the analyzer including line number, code snippet and markdown message.
  * Adds a hub request function `getIncidents` and query `useFetchIncidents` for fetching these.
* Adds the new `FileIncidentsDetailModal` component which is opened by clicking a file in the `IssueAffectedFilesTable`.
  * At the top level, this modal only fetches the first 5 incidents for the file (useFetchIncidents is used without any table hooks, so the `hubRequestParams` are passed in by hand to filter by file and limit to 5 results instead of using `getHubRequestParams`).
  * Tabs are dynamically rendered for each of the first 5 incidents. In each of these tabs, a grid is rendered with details including the `IncidentCodeSnipViewer` (see below).
    * Installs the `react-markdown` package for rendering the incident message. Adds a custom object in `client/src/app/components/markdown-pf-components.tsx` for configuring react-markdown to use PF-specific components for common HTML tags.
  * A 6th tab for "All incidents" is rendered if there are more than 5 total incidents. In this tab, the `FileAllIncidentsTable` is rendered (see below).
* Adds the `IncidentCodeSnipViewer` component, which renders the `incident.codeSnip` in PatternFly's CodeEditor component in read-only mode with a marker (red squiggly underline and icon) on the affected line.
  * Installs required packages `@patternfly/react-code-editor`, `monaco-editor`, `react-monaco-editor`, and `monaco-editor-webpack-plugin`. See https://github.com/patternfly/patternfly-react/blob/main/packages/react-code-editor/README.md#usage
  * Configures `monaco-editor-webpack-plugin`-related loaders in webpack config. See https://github.com/microsoft/monaco-editor/tree/main/webpack-plugin#using
  * Includes logic to reason about relative and absolute line numbers. The CodeEditor methods deal with relative line numbers (1 is the first line in the codeSnip string, even if the codeSnip starts on line 15 of the source file), and the hub and display logic deal with absolute line numbers (the line numbers from the source file).
    * Note that for now, the analyzer/hub does not indicate what line the codeSnip starts on. This PR hard-codes `codeSnipStartLine = 1`, and we will need to later add logic to strip line numbers out of the codeSnip string and determine this number from them (see https://github.com/konveyor/enhancements/issues/124)
   * Includes a commented-out `language` prop, which can be set to enable syntax highlighting.
     * Note that for now, the analyzer/hub does not indicate what language the codeSnip is written in (see https://github.com/konveyor/enhancements/issues/125). We will need to identify this and use it for the `language` prop, and then we may want to limit the UI to a static list of supported languages and configure the monaco webpack plugin to only include resources for those languages in the build. See https://github.com/microsoft/monaco-editor/tree/main/webpack-plugin#options
* Includes some miscellaneous code cleanup in files surrounding the Issues and Affected Apps pages.
* Adds a `noMargin` prop to `SimplePagination` for rendering it inline somewhere (it is aligned with column headers in the "all incidents" table as per the mockup)
* Adds a `forceNumRenderedColumns` argument to `useTableControlProps` to override the automatically calculated `numRenderedColumns` when using custom `colSpan`s in the table (for adding this extra pagination column header that overlaps the last data column)
* Includes some TODO comments to be addressed in future PRs.


client/src/app/api/models.ts
```diff
@@ -599,12 +610,24 @@ export interface AnalysisRuleReport extends BaseAnalysisRuleReport {
...
+export interface AnalysisIncident {
+  id: number;
+  file: string;
+  line: number;
+  message: string;
+  codeSnip: string;
+  facts: Record<string, string>; // TODO what's actually in here?
+  createTime?: string; // ISO-8601 timestamp
+  createUser?: string;
+  updateUser?: string;
```
gildub, line 628: Maybe we don't need those in the model unless we use them ?

gildub (commented):
Looks good, just one question about the model.
I haven't been able to test yet.

client/src/app/api/models.ts
mturley: Yeah, probably not. I was just putting all the properties in before I started using any, but we probably don't care about these. Will remove

gildub (approved):
Looks nice to me !

ibolton336:
Nice work! Cool to see this coming together.

PR #977, opened 2023-06-02
:bug: Do not allow deletion of jira issue from ui

### LINKED ISSUE #975: Do not allow deletion of tickets

ibolton336:
Closes #975


gildub (changes requested):
Removing applications from migration wave doesn't remove the ticket.
I think we need to discuss the related issue.

Issue #911, opened 2023-05-22
After deleting the last application in the status nested table, a blank status is shown


ibolton336:

ibolton336:
![Image](https://github.com/konveyor/tackle2-ui/assets/11218376/9e37ac6c-0a3f-4ca8-a588-0981fc3f0294)

gildub:
@ibolton336, 

We don't display the section when there are no issues.

This is following the same behavior described in (Migration Waves Management([https://github.com/konveyor/enhancements/tree/master/enhancements/migration-waves#migration-waves-management-1] were we don't display the section if there are no issue manager:

"This section will only be displayed if there is at least one Issue Manager configured in the Konveyor instance" 

Confirmed by @rromannissen.

PR #979, opened 2023-06-02
:bug: Add JIRA status values

### LINKED ISSUE #974: Migration waves aggregated status value shows "In Progress" although there are no tickets
To reproduce, 

Create migration wave, add applications, export them to Jira instance.
Then remove tickets on Jira side.

ibolton336:
Closes #974 #911 

- Adds Completed status
- Adds In Progress status
- Filters out empty status from derived status calculation. This could mean the issue has been deleted from the underlying JIRA instance OR that the jira ticket is in the process of being created. We have no way of knowing how to interpret empty string, so this filters it out which gives us a more accurate representation of aggregated status.

### DIFF
--- a/client/src/app/utils/waves-selector.ts
+++ b/client/src/app/utils/waves-selector.ts
@@ -34,17 +34,12 @@ export const getWavesWithStatus = (
     const statuses = getTicketStatus(wave, tickets);
     if (statuses.includes("Error")) {
       return "Error";
-    } else if (statuses.includes("")) {
-      const now = dayjs.utc();
-      const start = dayjs.utc(wave.startDate);
-      var duration = now.diff(start);
-      if (duration > 0) {
-        return "In Progress";
-      } else {
-        return "Not Started";
-      }
     } else if (statuses.includes("New")) {
       return "Issues Created";
+    } else if (statuses.includes("In Progress")) {
+      return "In Progress";
+    } else if (statuses.every((status) => status === "Done")) {
+      return "Completed";
     } else {
       return "Not Started";
     }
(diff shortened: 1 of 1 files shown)


client/src/app/utils/waves-selector.ts
```diff
@@ -34,17 +34,12 @@ export const getWavesWithStatus = (
     const statuses = getTicketStatus(wave, tickets);
     if (statuses.includes("Error")) {
       return "Error";
-    } else if (statuses.includes("")) {
-      const now = dayjs.utc();
```
gildub, line 38: Are we not handling the starting date anymore for the "In progress" aggregate status value ?

gildub (commented):
Just wondering if requirements have changed

ibolton336:
> 

@gildub Ok understood. Wondering how we can best handle the empty ticket status false positive when issue has been deleted from the underlying JIRA instance.

gildub:
@rromannissen, as of now when clicking on the aggregated status tab of a migration wave which has applications assigned to it, either because not exported yet or because the issues were removed in Jira, we're displaying an empty line.
Mabye we need to be provide more information to the user and display explicitly that there are no tickets.
What do you think ?

client/src/app/utils/waves-selector.ts
ibolton336: I was thinking of this and wondering where the start date logic falls in with the other status we are deriving from JIRA. When would we prioritize using the date to derive status over using the status of the issues to derive status? Seems as if there should be a state for when all issues are in "New" or "Not started" state but the migration date is technically "In progress", for example.

rromannissen:
@gildub sounds like a good idea to me! If I understood correctly, it's just about displaying a message saying something like "No issues created yet" when there are no issues (initial state) or all issues associated to the applications have been deleted. Is that right?

gildub:
@rromannissen, that's right.

@ibolton336, let's check with UXD for the best way to present empty applications tickets to the user.
Maybe on a follow-up PR if easier.

ibolton336:
Sounds great @gildub @rromannissen I will capture the new status proposition in an issue.

PR #989, opened 2023-06-06
:bug: Credential modal, reset value and revalidate form for password onFocus

sjd78:
When `onFocus` to a password or token input field, reset the value back to default and have the form revalidate. Use `resetField()` instead of `setValue("password", "")` to achieve this behavior.

The field reset and form invalidation causes the Create button to be disabled and the old password removed when focus enters the password field.

Bug-Url: https://issues.redhat.com/browse/MTA-474

Behavior with this change:

https://github.com/konveyor/tackle2-ui/assets/3985964/7c757a42-b83a-44a7-935f-d76a9ddbabd8


gildub (approved):
LGTM

PR #996, opened 2023-06-07
:seedling: Add password blur test to identity form

sjd78:
Followup PR #989 and extend the `identity-form.test.tsx` test "Identity form validation test - source - username/password" to include blur of the password field and refocus to make sure the password is cleared and the submit button is disabled.


gildub (approved):
LGTM

Issue #865, opened 2023-05-09
Migration waves: name is not mandatory


gildub:
A migration wave name must be changed to be optional.

For details see https://github.com/konveyor/enhancements/tree/master/enhancements/migration-waves#create-new-waves

PR #997, opened 2023-06-08
:sparkles: Migration Waves:  name is not required, use start and end dates if needed

gildub:
The name of migration is optional, the start and end dates are used if needed to distinguish migration waves.
Also migration wave names are not unique.

Resolves https://github.com/konveyor/tackle2-ui/issues/865

### DIFF
--- a/client/src/app/pages/migration-waves/components/manage-applications-form.tsx
+++ b/client/src/app/pages/migration-waves/components/manage-applications-form.tsx
@@ -223,7 +223,7 @@ export const ManageApplicationsForm: React.FC<ManageApplicationsFormProps> = ({
         <Text component={TextVariants.h5}>Selected wave</Text>
         <TextInput
           value={
-            migrationWave?.name === ""
+            !migrationWave?.name
               ? `${dayjs(migrationWave.startDate).format(
                   "MM/DD/YYYY"
                 )} - ${dayjs(migrationWave.endDate).format("MM/DD/YYYY")}`
--- a/client/src/app/pages/migration-waves/components/manage-applications-form.tsx
+++ b/client/src/app/pages/migration-waves/components/manage-applications-form.tsx
@@ -203,7 +203,7 @@ export const ManageApplicationsForm: React.FC<ManageApplicationsFormProps> = ({
       applications: selectedItems.map((application) => {
         return { id: application.id, name: application.name };
       }),
-      name: migrationWave?.name ? migrationWave.name.trim() : "",
+      name: migrationWave?.name?.trim() || "",
       startDate: migrationWave?.startDate || "",
       endDate: migrationWave?.endDate || "",
       stakeholders: migrationWave?.stakeholders || [],
@@ -216,8 +216,6 @@ export const ManageApplicationsForm: React.FC<ManageApplicationsFormProps> = ({
 
     onClose();
   };
-
-  console.log(migrationWave);
 
   return (
     <Form onSubmit={onSubmit}>
--- a/client/src/app/pages/migration-waves/components/manage-applications-form.tsx
+++ b/client/src/app/pages/migration-waves/components/manage-applications-form.tsx
@@ -13,6 +13,7 @@ import {
   ToolbarContent,
   ToolbarItem,
 } from "@patternfly/react-core";
+import dayjs from "dayjs";
 
 import { Application, MigrationWave } from "@app/api/models";
 import { ToolbarBulkSelector } from "@app/shared/components";
@@ -202,7 +203,7 @@ export const ManageApplicationsForm: React.FC<ManageApplicationsFormProps> = ({
       applications: selectedItems.map((application) => {
         return { id: application.id, name: application.name };
       }),
-      name: migrationWave?.name.trim() || "",
+      name: migrationWave?.name ? migrationWave.name.trim() : "",
       startDate: migrationWave?.startDate || "",
       endDate: migrationWave?.endDate || "",
       stakeholders: migrationWave?.stakeholders || [],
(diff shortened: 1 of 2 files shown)


client/src/app/pages/migration-waves/components/migration-wave-form.tsx
```diff
@@ -58,7 +58,7 @@ const stakeholderToOption = (
 });
 
 interface WaveFormValues {
-  name: string;
+  name: string | undefined;
```
ibolton336, line 61:
```suggestion
  name?: string;
```

ibolton336 (changes requested):

client/src/app/pages/migration-waves/components/manage-applications-form.tsx
```diff
@@ -202,7 +203,7 @@ export const ManageApplicationsForm: React.FC<ManageApplicationsFormProps> = ({
       applications: selectedItems.map((application) => {
         return { id: application.id, name: application.name };
       }),
-      name: migrationWave?.name.trim() || "",
+      name: migrationWave?.name ? migrationWave.name.trim() : "",
```
mturley, line 206: This is fine as-is, but I think you could have continued the optional chaining for a shorter line. If any part of the optional chain is undefined the whole expression evaluates to undefined, so the `||` is enough to catch you.

```
name: migrationWave?.name?.trim() || "",
```

client/src/app/pages/migration-waves/components/manage-applications-form.tsx
```diff
@@ -216,12 +217,20 @@ export const ManageApplicationsForm: React.FC<ManageApplicationsFormProps> = ({
...
+  console.log(migrationWave);
+
   return (
     <Form onSubmit={onSubmit}>
       <TextContent>
         <Text component={TextVariants.h5}>Selected wave</Text>
         <TextInput
-          value={migrationWave.name}
+          value={
+            migrationWave?.name === ""
```
mturley, line 228: Are we sure `migrationWave.name` will be `""` here or could it be undefined? Elsewhere you check if it is `null`. I wonder if to be safe any place that needs to fall back if the name is missing should work for `""`, `null` or `undefined` values, so it could just check if the name is falsy.

```
!migrationWave?.name ? ... : ...
```

client/src/app/pages/migration-waves/components/migration-wave-form.tsx
```diff
@@ -181,7 +173,8 @@ export const WaveForm: React.FC<WaveFormProps> = ({
   } = useForm<WaveFormValues>({
     mode: "onChange",
     defaultValues: {
-      name: migrationWave?.name || "",
+      name:
+        migrationWave && migrationWave.name !== null ? migrationWave.name : "",
```
mturley, line 177: I'm pretty sure this change has no effect, since a `null` name value is falsy and would cause `migrationWave?.name || ""` to evaluate to `""`. I would vote to revert this for simplicity.

client/src/app/pages/migration-waves/components/migration-wave-form.tsx
```diff
@@ -200,7 +193,8 @@ export const WaveForm: React.FC<WaveFormProps> = ({
   const onSubmit = (formValues: WaveFormValues) => {
     const payload: New<MigrationWave> = {
       applications: migrationWave?.applications || [],
-      name: formValues.name.trim(),
+      name:
```
mturley, line 196: Similarly, this can be simplified with optional chaining:

```
name: formValues.name?.trim() || ""
```

client/src/app/pages/migration-waves/components/manage-applications-form.tsx
gildub: @mturley, yes the backend always has `migrationWave.name`  field defined where it could be empty (""). 
Your approach is slicker.

client/src/app/pages/migration-waves/components/migration-wave-form.tsx
gildub: Right, I restored it.

client/src/app/pages/migration-waves/components/migration-wave-form.tsx
```diff
@@ -138,15 +138,7 @@ export const WaveForm: React.FC<WaveFormProps> = ({
     name: yup
       .string()
       .trim()
-      .min(3, t("validation.minLength", { length: 3 }))
-      .max(120, t("validation.maxLength", { length: 120 }))
-      .test(
-        "Duplicate name",
-        "An identity with this name already exists. Use a different name.",
```
ibolton336, line 145: Maybe we should keep this for duplicate migration wave names.

gildub: @ibolton336, I thought about it but it isn't a requirement.

ibolton336: Seems like an implied requirement that we wouldn't want a user creating duplicates IMO. Is there a case FOR having duplicate migration waves? 

Screenshot 2023-06-12 at 3 55 36 PM

gildub: @ibolton336, I don't think there is any implied requirement.
Maybe a migrator would like to break down a migration in separate waves starting at different times.
@rromannissen, what do you think ?

ibolton336 (approved):

client/src/app/pages/migration-waves/components/migration-wave-form.tsx
gildub: Let's follow-up with core team to have a composite primary key.

ibolton336: > I don't think there should be duplicate waves.
So sounds like we should just open up a new issue and add the duplicate check back in?

gildub: @ibolton336, the duplicate check will need to include start and end dates.
And make sure the hub enforces that too.

ibolton336: Gotcha. So the name field is set to start date/ end date when no name exists, right? So wouldn't the existing check work for that, too?

PR #1000, opened 2023-06-09
:sparkles: Forms: use isDirty instead of isTouched for immediate feedback to user

gildub:
Make `getValidatedFromErrorTouched()` to use `isDirty` instead of `isTouched` 

Resolves https://issues.redhat.com/browse/MTA-658


sjd78 (commented):
The change makes sense. I have a suggestion to make it smoother.

mturley (approved):
I think I prefer @sjd78's suggestion that results in less awkward naming, but this fixes the issue and isn't disruptive so it LGTM. Maybe we can follow up to clean it up.

gildub:
@mturley, @sjd78, I changed it for the better name `getValidatedFromErrors()` .

PR #1005, opened 2023-06-09
:bug: Fix broken axios get calls for fetching tags and tag categories by id

mturley:
In https://github.com/konveyor/tackle2-ui/pull/889, the `getTagById` and `getTagCategoryById` functions were changed to use axios, but the `.then` callback was left out. That means a Promise is being returned instead of a Tag or TagCategory, and TypeScript isn't catching it because axios doesn't know what type it is fetching. This led to tags not being displayed in the app details drawer on the app inventory page.

### DIFF
--- a/client/src/app/pages/applications/components/application-tags/application-tags.tsx
+++ b/client/src/app/pages/applications/components/application-tags/application-tags.tsx
@@ -59,7 +59,6 @@ export const ApplicationTags: React.FC<ApplicationTagsProps> = ({
 
   useEffect(() => {
     if (application.tags) {
-      console.log({ application, tags });
       setIsFetching(true);
 
       Promise.all(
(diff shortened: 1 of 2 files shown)


client/src/app/pages/applications/components/application-tags/application-tags.tsx
```diff
@@ -59,6 +59,7 @@ export const ApplicationTags: React.FC<ApplicationTagsProps> = ({
 
   useEffect(() => {
     if (application.tags) {
+      console.log({ application, tags });
```
gildub, line 62: Do we need to keep the log ?

client/src/app/pages/applications/components/application-tags/application-tags.tsx
```diff
@@ -68,6 +69,7 @@ export const ApplicationTags: React.FC<ApplicationTagsProps> = ({
       )
         .then((tags) => {
           const tagsWithSources = tags.reduce((prev, current, index) => {
+            console.log({ prev, current });
```
gildub, line 72: Do we need to keep the log ?

gildub (commented):
Thanks for fixing.
See inline for console.log question

client/src/app/pages/applications/components/application-tags/application-tags.tsx
mturley: Ah, sorry about that

PR #1006, opened 2023-06-09
:bug: MTA-479: Use tag category fallback colors in tags table and tag category edit form

mturley:
Fixes https://issues.redhat.com/browse/MTA-479


gildub (approved):
LGTM

PR #1042, opened 2023-06-15
:bug: Don't show empty {{what}} value in success notifications

ibolton336:
- Fixes empty string issue which is breaking QE tests
[Update notifications include two spaces](https://issues.redhat.com/browse/MTA-753)


mturley (approved):
LGTM

PR #1046, opened 2023-06-15
:bug: TrackerForm: Fix identityOptions

gildub:
Trap undefined which slipped through the cracks of React Hook Form.


mturley (commented):
Can we also change `kind` to `kind?` in `FormValues` in case there is any other code in the future that breaks when kind is undefined?

mturley (approved):

mturley (approved):

PR #1048, opened 2023-06-16
:sparkles: Add missing filters on Issues page

mturley:
* Adds optional properties to `FilterCategory` to support cases where the field and value we want to give to the hub are different from the ones we store in UI state (since UI filterValues keys must be unique, but hub filter keys can be repeated):
  * `serverFilterField` is used instead of the key from the `filterValues` object when serializing for the hub
  * `getServerFilterValue` takes the `value` array from our `filterValues` object and returns the array of values for the hub
  * If either of these are omitted, the existing behavior is used (keys and values of `filterValues` from state are used for the hub).
* Adds missing filters for the issues page:
  * Business service
    * Also added to the affected apps page
    * Select options generated from fetching all apps, which should already be in the query cache if the user has visited the app inventory page
  * Tags
    * Also added to the affected apps page
    * Select options generated from fetching all tags
    * Uses `getServerFilterValue` to filter by tag id in the hub while displaying tag name in the UI
    * See `NOTE` comment on line 74 of `client/src/app/pages/issues/helpers.ts`
  * Category (regular text search filter)
  * Source
    * In the hub this is filtering on `labels` (via `serverFilterField`) with the prefix `konveyor.io/source=` (via `getServerFilterValue`).
  * Target
    * In the hub this is filtering on `labels` (via `serverFilterField`) with the prefix `konveyor.io/target=` (via `getServerFilterValue`).
* Factors out `useSharedFilterCategoriesForIssuesAndAffectedApps` hook which returns `filterCategories` used for both the issues and affected apps pages (and abstracts away fetching the resources we need to generate the select options for these).
* Adds business service and application tags to the filters we carry between the issues and affected apps pages.
* In `MultiselectFilterControl`, changes how the option search box in the filter dropdown works so that it can match based on either the `key` or the `value` of the option object. I changed this to support an earlier version of the tags filter where I tried relying on `key` being the tag id, before realizing that didn't work and I needed `serverFilterField`. I left it in because I think it is a net good change anyway. All our existing filters have `key` and `value` as the same value, so this does not affect any existing filters.
* In `getFilterHubRequestParams`:
  * Adds support for the `serverFilterField` and `getServerFilterValue` properties of `FilterCategory`
  * For "search"-type filters, does not wrap the value in wildcards (`*`) so that we can decide where to put the wildcards on a per-filter level using `getServerFilterValue`. This lets us put the wildcard before and after the source and target query strings even though they are not at the beginning of the full label string.
  * Changes how we join multiple filters for the hub request, to support ANDing multiple UI filters that use the same hub field (here used for `labels` filters for source and target). Now multiple `HubFilter`s using the same field get merged into a single `HubFilter` before passing to the hub (where possible - nested lists are not supported), and all filters are in a single comma-separated `filters` hub URL param instead of multiple URL params.
    * Depends on https://github.com/konveyor/tackle2-hub/pull/409


ibolton336 (approved):

client/src/app/shared/components/FilterToolbar/MultiselectFilterControl.tsx
```diff
@@ -92,12 +92,16 @@ export const MultiselectFilterControl = <
...
-        // which could become a problem maybe?
-        return optionProps?.key
-          ?.toLowerCase()
-          .includes(textInput?.toLowerCase());
+        // Note: The in-dropdown filter can match the option's key or value. This may not be desirable?
+        if (!textInput) return false;
+        const optionValue =
+          typeof optionProps?.value === "string"
+            ? optionProps.value
+            : optionProps.value.toString();
```
sjd78, line 100: What's wrong with `String.toString()`?
```suggestion
        const optionValue = optionProps?.value?.toString();
```

mturley: Ah, fair point. Forgot the string type had that.

Issue #855, opened 2023-05-05
Add error message for failed jira tracker creation


ibolton336:
Currently, if a create request fails for create jira tracker, the UI fails silently with no error message

In the hub logs, I am seeing:

```
[GIN] 2023/05/05 - 13:23:52 | 200 |     930.791µs |       127.0.0.1 | GET      "/trackers"
{"level":"error","ts":1683307432.8720489,"logger":"tickets","msg":"Connection test failed.","tracker":1,"error":"No response returned: Get \"/slucidi.atlassian.net/rest/api/2/myself\": unsupported protocol scheme \"\"","errorVerbose":"Get \"/slucidi.atlassian.net/rest/api/2/myself\": unsupported protocol scheme \"\"\nNo response returned","stacktrace":"github.com/go-logr/zapr.(*zapLogger).Error\n\t/Users/ibolton/code/go/pkg/mod/github.com/go-logr/zapr@v0.3.0/zapr.go:132\ngithub.com/konveyor/controller/pkg/logging.(*Logger).Error\n\t/Users/ibolton/code/go/pkg/mod/github.com/konveyor/controller@v0.8.0/pkg/logging/logger.go:109\ngithub.com/konveyor/tackle2-hub/tracker.(*Manager).testConnection\n\t/Users/ibolton/code/go/src/github.com/konveyor/tackle2-hub/tracker/manager.go:85\ngithub.com/konveyor/tackle2-hub/tracker.(*Manager).testConnections\n\t/Users/ibolton/code/go/src/github.com/konveyor/tackle2-hub/tracker/manager.go:67\ngithub.com/konveyor/tackle2-hub/tracker.(*Manager).Run.func1\n\t/Users/ibolton/code/go/src/github.com/konveyor/tackle2-hub/tracker/manager.go:42"}
{"level":"error","ts":1683307442.888748,"logger":"tickets","msg":"Connection test failed.","tracker":1,"error":"No response returned: Get \"/slucidi.atlassian.net/rest/api/2/myself\": unsupported protocol scheme \"\"","errorVerbose":"Get \"/sl
```

``` 
[GIN] 2023/05/05 - 13:35:47 | 200 |     393.625µs |       127.0.0.1 | GET      "/trackers"
[...]

PR #1049, opened 2023-06-16
:bug: Update jira instance connection column to use popover

sjd78:
When a jira instance on the main jira configuration table has a connectivity issue, the connection column will show "Not connected". Clicking the text will now open a popover to display additional details about the state of the instance.  The popover replaces a standard modal box with similar information.

Fixes: #855

What it looked like before:
image

What it looks like now:
image


gildub:
@sjd78, that sounds good, although during a previous conversation with @rromannissen the idea is to display errors from the Jira in code block (same for the tickets errors). 
Be aware the message can't be several lines long too.
So this needs to be validated by UXD.
Thansk

gildub:
@sjd78 , sounds good, I was concerned about the style change but I can see Ramon validation on the related issue.
Could you please rebase ?

gildub (changes requested):
Looks good,
please use I18n for text as in initial code ?

sjd78:
> Looks good, please use I18n for text as in initial code ?

I swapped the `"Connection"` text back to `t("terms.instance")` for the popover title.  I'll leave the description in the body above the code block since the popover design guidelines always include a header and descriptive text.  Having just the header with the code block text looks a bit awkward.

(rebased also)

client/src/app/pages/external/jira/components/tracker-status.tsx
```diff
@@ -26,28 +29,28 @@ const TrackerStatus = ({ connected, message }: ITrackerStatusProps) => {
...
-              variant="link"
-              onClick={() => setCodeModalState(message)}
+            <Popover
+              aria-label="More information about no connection"
+              alertSeverityVariant="danger"
+              headerIcon={<ExclamationCircleIcon />}
+              headerContent={t("composed.error", { what: t("terms.instance") })}
+              bodyContent={
+                <TextContent>
+                  <Text>Jira instance {name} is not connected.</Text>
```
gildub, line 39: Can we use i18n text here ?

gildub (commented):
So we' re consistent please use i18n whenever there is user facing text.

ibolton336 (approved):

gildub (approved):
LGTM, thanks !

PR #1054, opened 2023-06-20
:sparkles: Issues: Parse and strip line numbers from codeSnip in Incident viewer

mturley:
`incident.codeSnip` now includes line numbers embedded in the string with the following properties:
* There may or may not be leading whitespace before the line number
* The line numbers are always right-aligned (end on the same character index in the line)
* The line number is followed by 2 spaces, then by the code for that line (which possibly includes more leading whitespace)
* If there is no code on that line, the line number may be followed by only 1 space or no whitespace

See:
* https://github.com/konveyor/enhancements/issues/124
* https://github.com/konveyor/analyzer-lsp/issues/200
* https://github.com/konveyor/tackle2-hub/pull/403
* https://github.com/konveyor/tackle2-hub/pull/414

This PR adds a regex for matching each line of the codeSnip to pull out the line number and code, tolerating leading whitespace before the number and preserving original whitespace from the code. It uses the number found on the first line as `codeSnipStartLine` for the purposes of custom rendering of line numbers in the CodeEditor, and strips the numbers and 2-space padding out before rendering the code.


gildub (approved):
LGTM

Issue #1053, opened 2023-06-20
Update tracker metadata to use new API route


ibolton336:
> The issue that is preventing QE from connecting to the Red Hat Jira instance is related to an API endpoint that a Jira admin can disable. Unfortunately, it's an endpoint that I was using to populate the metadata field on the tracker that you guys were getting the projects and issue types from. I have a PR up to deal with this, but unfortunately I can't provide you that information in the metadata field now. There's going to be a Hub endpoint /trackers/:id/projects that will give you a list of projects (name and id), and then another endpoint /trackers/:id/projects/:id2/issuetypes that will give you the issue types for that project (also name and id). The metadata field on the Tracker resource will go away. (edited) 

We need to update our trackers.ts to include a new slice of state retrieved from the api for issueTypes & projects. This may be able to be handled using the selector pattern for derived state we have established for wave status.

PR #1056, opened 2023-06-21
:sparkles: Refactor HUB API Jira changes

gildub:
Resolves https://github.com/konveyor/tackle2-ui/issues/1053

### DIFF
--- a/client/src/app/pages/migration-waves/components/status-table.tsx
+++ b/client/src/app/pages/migration-waves/components/status-table.tsx
@@ -88,7 +88,7 @@ export const WaveStatusTable: React.FC<IWaveStatusTableProps> = ({
   const getTicketByApplication = (tickets: Ticket[], id: number = 0) =>
     tickets.find((ticket) => ticket.application?.id === id);
 
-  const getTicketIssue = (appId: number) => {
+  const useTicketIssue = (appId: number) => {
     const ticket = getTicketByApplication(tickets, appId);
     const types = useTrackerTypesByProjectId(
       ticket?.tracker?.name,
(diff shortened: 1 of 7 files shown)


client/src/app/pages/migration-waves/components/status-table.tsx
```diff
@@ -84,18 +84,17 @@ export const WaveStatusTable: React.FC<IWaveStatusTableProps> = ({
...
     if (appId) {
       const ticket = getTicketByApplication(tickets, appId);
       if (ticket) {
-        const types = getTrackerTypesByProjectId(
-          trackers,
+        const types = useTrackerTypesByProjectId(
           ticket.tracker.name,
-          ticket?.parent
+          ticket.parent
         );
```
mturley, line 98: This has the same issue we identified before with `useTrackerProjectsByTracker` where it is conditionally calling a hook. It should be renamed `useTicketIssue`, and the `useTrackerTypesByProjectId` should be called outside the `if (ticket)` block, which will require its params to be optional (internally passing `enabled: false` to useQuery if they are missing). Then make sure `useTicketIssue` is not being called in a conditional or callback (called at render time, every render).

mturley (changes requested):
One more hook rules issue here, otherwise LGTM.

client/src/app/pages/migration-waves/components/status-table.tsx
gildub: Thanks @mturley , I removed the hook from being under condition.

mturley (approved):
LGTM

PR #1057, opened 2023-06-21
:sparkles: Add facts modal/code viewer & detail drawer tab 

ibolton336:
- Adds initial implementation of the facts modal, analysis detail drawer tab for facts, and code viewer for facts json.

Notes: This does not cover the filter/sort logic for facts based on sources. There is no mock data available from the api for this yet, but we will be required to collate the fact sources ourselves on the UI side based on the fact keys returned from the hub. https://github.com/konveyor/tackle2-hub/pull/396 for more info on how the api will look. 

TODO: https://github.com/konveyor/tackle2-ui/issues/1061 
TODO: [How should we type facts with an unknown data type? ](https://github.com/konveyor/tackle2-ui/issues/1062)

### DIFF
--- a/client/src/app/api/models.ts
+++ b/client/src/app/api/models.ts
@@ -666,4 +666,7 @@ export interface Ticket {
   error?: boolean;
 }
 
-export type Fact = any;
+export type Fact = {
+  name: string;
+  data: any;
+};
--- a/client/src/app/api/models.ts
+++ b/client/src/app/api/models.ts
@@ -653,19 +653,6 @@ export interface WaveWithStatus extends MigrationWave {
   allStakeholders: StakeholderWithRole[];
 }
 
-export interface Ticket {
-  id: number;
-  application?: Ref | null;
-  tracker: Ref;
-  kind: string;
-  parent?: string;
-  fields?: Ref | null;
-  readonly message?: string | null;
-  reference?: string | null;
-  readonly status?: TicketStatus | null;
-  error?: boolean;
-}
-
 export type Fact = {
   name: string;
   data: any;
--- a/client/src/app/api/models.ts
+++ b/client/src/app/api/models.ts
@@ -652,3 +652,18 @@ export interface WaveWithStatus extends MigrationWave {
   fullApplications: Application[];
   allStakeholders: StakeholderWithRole[];
 }
+
+export interface Ticket {
+  id: number;
+  application?: Ref | null;
+  tracker: Ref;
+  kind: string;
+  parent?: string;
+  fields?: Ref | null;
+  readonly message?: string | null;
+  reference?: string | null;
+  readonly status?: TicketStatus | null;
+  error?: boolean;
+}
+
+export type Fact = any;
--- a/client/src/app/api/models.ts
+++ b/client/src/app/api/models.ts
@@ -652,8 +652,10 @@ export interface WaveWithStatus extends MigrationWave {
   fullApplications: Application[];
   allStakeholders: StakeholderWithRole[];
 }
+export type UnstructuredFact = any;
 
 export type Fact = {
   name: string;
+  //TODO: Address this when moving to structured facts api
   data: any;
 };
--- a/client/src/app/api/rest.ts
+++ b/client/src/app/api/rest.ts
@@ -701,6 +701,7 @@ export const updateTagCategory = (obj: TagCategory): Promise<TagCategory> =>
   axios.put(`${TAG_CATEGORIES}/${obj.id}`, obj);
 
 // Facts
+
 export const getFacts = (id: number | string | undefined) =>
   id
     ? axios.get(`${APPLICATIONS}/${id}/facts`).then((response) => response.data)
--- a/client/src/app/api/rest.ts
+++ b/client/src/app/api/rest.ts
@@ -45,6 +45,7 @@ import {
   TrackerProject,
   TrackerProjectIssuetype,
   Fact,
+  UnstructuredFact,
 } from "./models";
 import { QueryKey } from "@tanstack/react-query";
 import { serializeRequestParamsForHub } from "@app/shared/hooks/table-controls";
@@ -702,7 +703,10 @@ export const updateTagCategory = (obj: TagCategory): Promise<TagCategory> =>
 
 // Facts
 
-export const getFacts = (id: number | string | undefined) =>
+export const getFacts = (
+  id: number | string | undefined
+): Promise<UnstructuredFact> =>
+  //TODO: Address this when moving to structured facts api
   id
     ? axios.get(`${APPLICATIONS}/${id}/facts`).then((response) => response.data)
     : Promise.reject();
--- a/client/src/app/api/rest.ts
+++ b/client/src/app/api/rest.ts
@@ -701,6 +701,7 @@ export const updateTagCategory = (obj: TagCategory): Promise<TagCategory> =>
   axios.put(`${TAG_CATEGORIES}/${obj.id}`, obj);
 
 // Facts
-export const getFacts = (id: number | string): AxiosPromise<Fact> => {
-  return APIClient.get(`${APPLICATIONS}/${id}/facts`, jsonHeaders);
-};
+export const getFacts = (id: number | string | undefined) =>
+  id
+    ? axios.get(`${APPLICATIONS}/${id}/facts`).then((response) => response.data)
+    : Promise.reject();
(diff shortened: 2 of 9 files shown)


client/src/app/api/models.ts
```diff
@@ -658,3 +658,8 @@ export interface WaveWithStatus extends MigrationWave {
   fullApplications: Application[];
   allStakeholders: StakeholderWithRole[];
 }
+
+export type Fact = {
+  name: string;
+  data: any;
```
mturley, line 664: I was going to suggest we replace `any` with `unknown` here to prevent any issues with code assuming it's a string (JSON.stringify won't care what it is), but since we're going to revisit this to support structured facts it's probably fine to leave it as-is (and it stands out as something to be fixed anyway which is maybe good)

client/src/app/pages/applications/components/application-detail-drawer/application-facts.tsx
```diff
@@ -0,0 +1,98 @@
...
+              setFilterValues={setFilterValues}
+              showFiltersSideBySide
+            />
+          </ToolbarToggleGroup>
+        </ToolbarContent>
+      </Toolbar>
+      {filteredFacts.map((fact) => {
+        return (
+          <div>
+            <Button
```
mturley, line 80: This flat list of link-buttons with no real layout aside from line breaks really bothers me for some reason, but that's a design issue. Maybe we can revisit this with Justin. I think even just a compact table with one unlabeled column, or maybe a `DataList` (https://www.patternfly.org/v4/components/data-list), something to make it look less plain. Outside the scope of this PR though, just thinking out loud.

client/src/app/pages/applications/components/application-detail-drawer/fact-detail-modal/fact-code-snip-viewer.tsx
```diff
@@ -0,0 +1,22 @@
...
+  fact,
+}) => {
+  return (
+    <CodeEditor
+      isReadOnly
+      isDarkTheme
+      isLineNumbersVisible
+      language={Language.json}
+      height="450px"
+      code={JSON.stringify(fact.data)}
```
mturley, line 19: The design calls for displaying this in YAML rather than JSON right? Since we already have the `js-yaml` package installed we could `import yaml from 'js-yaml';` and then use `yaml.dump(fact.data)` below instead of `JSON.stringify(fact.data)`, and use `Language.yaml` instead of `Language.json`. I think that should be sufficient, although we may want to wrap the `yaml.dump` in a try-catch by lifting this code out into a variable. I suppose the API always gives us JSON and all JSON can be stringified to YAML so maybe that's fine.

client/src/app/queries/facts.ts
```diff
@@ -0,0 +1,34 @@
...
+import { Application, Fact } from "@app/api/models";
+import { ApplicationsQueryKey } from "./applications";
+
+export const FactsQueryKey = "facts";
+
+export const useFetchFacts = (applicationID: number | string | undefined) => {
+  const { data, isLoading, error, refetch } = useQuery(
+    [FactsQueryKey, applicationID],
+    {
+      queryFn: async () => await getFacts(applicationID),
```
mturley, line 14: I noticed this recently in a lot of our queries (and I am guilty of copying it), but we don't actually need async/await for these queryFns if all we're doing is calling a function that already returns a promise. We may want to simplify all these at some point to:
```ts
queryFn: () => getFacts(applicationID)
```
however this is harmless as-is, just a nitpick.

mturley (commented):
Just a couple things.

client/src/app/pages/applications/components/application-detail-drawer/fact-detail-modal/fact-code-snip-viewer.tsx
mturley: Ah, looking at [the docs](https://www.npmjs.com/package/js-yaml) we can avoid this ever throwing an exception by using the `skipInvalid` option:
```ts
code={yaml.dump(fact.data, { skipInvalid: true })}
```

client/src/app/api/rest.ts
```diff
@@ -678,3 +680,10 @@ export const createTagCategory = (
 
 export const updateTagCategory = (obj: TagCategory): Promise<TagCategory> =>
   axios.put(`${TAG_CATEGORIES}/${obj.id}`, obj);
+
+// Facts
+
+export const getFacts = (id: number | string | undefined) =>
```
sjd78, line 686: The other query functions have an explicit return type Promise<?>.  Does it make sense to add one here?

client/src/app/queries/facts.ts
```diff
@@ -0,0 +1,34 @@
...
+      onError: (error: AxiosError) => console.log("error, ", error),
+      select: (facts): Fact[] => {
+        return Object.keys(facts).map((fact) => {
+          return {
+            name: fact,
+            data: facts[fact],
+          };
+        });
+      },
+    }
```
sjd78, line 25: There is potential to make this shorter (and look more functional)....

```suggestion
      select: (facts): Fact[] =>
        Object.keys(facts).map(
          (fact) => ({ name: fact, data: facts[fact] })
        ),
```

client/src/app/pages/applications/components/application-detail-drawer/application-facts.tsx
```diff
@@ -0,0 +1,98 @@
...
+              variant="link"
+              isInline
+              onClick={() => setSelectedFactForDetailModal(fact)}
+            >
+              {fact.name}
+            </Button>
+          </div>
+        );
+      })}
+      {selectedFactForDetailModal ? (
```
sjd78, line 90: You use the form `{renderCondition && (<element />)}` form in other places, why ternary here?

sjd78 (commented):
Really just some style questions

client/src/app/api/rest.ts
ibolton336: This one may fall under the same category as [this comment](https://github.com/konveyor/tackle2-ui/pull/1057/files#r1237604675). I am unsure how to type this unstructured fact that is just a json/yaml object that hasn't been parsed yet.

client/src/app/pages/applications/components/application-detail-drawer/application-facts.tsx
ibolton336: I was trying to match the style we used in issue-affected-files-table for consistencies sake.

client/src/app/pages/applications/components/application-detail-drawer/application-facts.tsx
ibolton336: Opened an issue for this & linked in the description.

client/src/app/pages/applications/components/application-detail-drawer/fact-detail-modal/fact-code-snip-viewer.tsx
```diff
@@ -0,0 +1,22 @@
...
+    />
```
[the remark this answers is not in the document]
mturley: It took a little bit of configuring in https://github.com/konveyor/tackle2-ui/pull/966/files, but not much! now we just have it for wherever we want it

client/src/app/pages/applications/components/application-detail-drawer/application-facts.tsx
mturley: We're not consistent about it in this repo, but I've always tried to favor the ternary because `&&` can fall back to falsy values that still render something (like `0`). https://kentcdodds.com/blog/use-ternaries-rather-than-and-and-in-jsx

client/src/app/pages/applications/components/application-detail-drawer/fact-detail-modal/fact-code-snip-viewer.tsx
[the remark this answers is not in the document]
sjd78: Now it would be really good to get code splitting working so the main app bundle isn't too huge.

PS - I know how to get it working, just haven't worked it up into a skinny PR yet.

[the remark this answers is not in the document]
mturley: We would love the help with code splitting! It's something I've never really taken the time to figure out.

mturley (approved):
LGTM assuming we revisit the Fact/UnstructuredFact types in a followup.

PR #1071, opened 2023-06-24
:bug: added dynamic year updation

### LINKED ISSUE #1007: :bug: Year updation to be done
On seeing the info section of Konveyor , Copyright @ 2022 is mentioned. First of all it needs to be updated and I would like to replace the regular update by adding `Date.getFullYear()`. this will update the year automatically. Following is the screenshot
image

khareyash05:
Closes #1007 
Added dynamic year update


mturley (approved):
LGTM, although I'm not sure how the copyright itself works (whether it is safe to assume that the copyright will always be up to date in the current year). I think maybe we can merge this and come back to it, do you have an opinion @ibolton336 ?

ibolton336:
I'm fine with merging @mturley . I would like to see us use DayJS for consistency, but that is all I have.

mturley:
Ah, here's the problem though -- CI failed because of the snapshot test. We can't have text that changes depending on the time it is rendered in our snapshot tests or they will suddenly fail CI in January on other PRs. We should probably talk to the team and see if we even need this copyright line to be present.

mturley (changes requested):
I think we cannot merge this unless we either remove the snapshot test on the AboutModal or just change the text to statically say 2023 and update that every year. Otherwise our CI tests will suddenly start to fail for all PRs on January 1st 2024.

Issue #835, opened 2023-04-25
View report / task data in yaml


ibolton336:
- Display api fetched data for task details in yaml format

gildub:
@ibolton336, is this still on the radar as the analyzers report format was going to cover for it ?

sjd78:
Duplicated by https://issues.redhat.com/browse/MTA-773

With the current way of opening the details, only the default mime type for the service endpoint can be used:

```jsx
        onClick: () => {
          const task = getTask(row);
          if (task) window.open(`/hub/tasks/${task.id}`, "_blank");
        },
```

```jsx
  const openAnalysisDetails = () => {
    if (task) window.open(`/hub/tasks/${task.id}`, "_blank");
  };
```

A reasonable way forward would be to use a `CodeEditor` component in read-only mode similar to the `FactCodeSnipViewer` of #1057.

Maybe navigate to a new route in the normal application page?

Maybe have a skinny viewer app using a separate webpack entry point all together?

Thoughts @ibolton336 , @gildub , @mturley ?

mturley:
I think we probably don't want to take the user away from their current browser tab for this if we can help it (I know we currently do, but I think that was out of convenience since this was the easiest way to pull up raw data?)

My vote would be to open a modal with a read-only code viewer, just like #1057 as suggested.

PR #1072, opened 2023-06-27
:sparkles: Add `SimpleDocumentViewerModal` to display analysis details

sjd78:
Instead of opening a new browser window to show a user a json formatted analysis details document, open a modal on page with an embedded and enhanced CodeEditor.

The `SimpleDocumentViewerModal` component allows fetching and displaying the document as either a json or yaml document.  Users can request the other format on screen.  By using the `@patternfly/react-code-editor` component, a number of good things are enabled:
  - the content is displayed on page and is code colored per language
  - the contents can be copied to the clipboard
  - the contents can be save to a file
  - the user can easily view either the json or yaml format of the same document, as provided by the rest api endpoint

Resolves: #835
Resolves: https://issues.redhat.com/browse/MTA-773


sjd78:
## How it looked before
Firefox:
Screenshot from 2023-06-27 18-59-07

Chrome:
Screenshot from 2023-06-27 18-58-39

## How it looks now
Firefox:
Screenshot from 2023-06-27 19-12-15


Chrome:
Screenshot from 2023-06-27 18-57-24

## Swapping between yaml and json:


https://github.com/konveyor/tackle2-ui/assets/3985964/733c30bd-fe37-40d2-8eea-95370d2d2a6c

sjd78:
I did not add any unit tests (snapshot or otherwise) so far.  I was trying to do it, but the testing framework does not like a component that does a `useEffect()` to grab data and update it's own display.  I have not figured out how to successfully do that yet.  As far as I got was for the component to render the empty state and snapshot it.  Trying to get the component rendered and a snapshot with data it mock fetches has been elusive.

Net result is that Codecov does NOT like me right now.

client/src/app/pages/applications/applications-table-analyze/applications-table-analyze.tsx
```diff
@@ -749,6 +757,13 @@ export const ApplicationsTableAnalyze: React.FC = () => {
           }}
         />
       )}
+
+      <SimpleDocumentViewerModal<Task | string>
```
ibolton336, line 761: LGTM! 

I was wondering why you chose to pass the fetch functions down as props rather than use them inside the component itself. Seems like since we are already using conditional logic to determine the language type [here](https://github.com/konveyor/tackle2-ui/pull/1072/files#diff-4835af0eed4a8f71b4f6d20783d916be40890f398377685be99c7db990be71b9R79-R88), then we can just use the appropriate fetch functions within that conditional? Does that make sense? 

Either way, I like the use of typescript to make the component more extensible down the road. Maybe that was your intention. Just curious.

ibolton336 (approved):

client/src/app/pages/applications/applications-table-analyze/applications-table-analyze.tsx
sjd78: Initially I wanted to have a more generic component that could be used elsewhere, and doing the fetch as a prop allows the location and methodology for the fetch can vary as needed.

In this PR, I eventually found 3 places that did a `window.open` or a Link with `target="_blank"`.  Two of those places used 1 URL, and the Link style was another.  So more generic FTW!  And playing with some generic typing with function type overloading got interesting.

One use:
https://github.com/konveyor/tackle2-ui/blob/aeeba069881d5bbb390b873199f7dd424029f654/client/src/app/pages/applications/components/application-detail-drawer/application-detail-drawer-analysis.tsx#L137-L142


The other use:
https://github.com/konveyor/tackle2-ui/blob/aeeba069881d5bbb390b873199f7dd424029f654/client/src/app/pages/applications/components/application-detail-drawer/application-detail-drawer-analysis.tsx#L113-L118

Issue #1074, opened 2023-06-29
tech stack: enable code splitting to allow React.lazy() components


sjd78:
In current webpack, code splitting is enabled by default.  In the project's configurations, there are mismatches that prevent code splitting from working.

A only base requirement to make code splitting work is to reconfigure `tsconfig.json` so it outputs in a style that webpack can use for code splitting.  For example, the transpiled js code needs to have `import` instead of `require`.

There are a few related changes that would be nice at the same time
  - reconfigure webpack to output assets, chunks, modules etc in nicer (for humans to look at) places
  - drop use of `path.resolve(__dirname, ...)` to something that would be easier to maintain going forward

PR #1075, opened 2023-06-29
:seedling: Enable webpack to do code splitting

sjd78:
## Summary
By updating `tsconfig.json` and making a few changes in webpack configuration, code splitting is now working.

For developers, this means that the previous 100MB+ app bundle that could frequently crash my browser when trying to debug code has been replaced by a small initial bundle with dynamic loading of each React.lazy() component.  Code debugging and source viewing works much better now.

For end users, this means the app can start up faster.  Less network traffic is great, but really only has an effect if the network latency is high (e.g. if the app is located on a different continent from the user).

Resolves: #1074 

### Refactor the client `tsconfig.json`
  - By updating the `tsconfig.json` to use `module: es2020`, this allows
    webpack to do actual code splitting based on `React.lazy()`.

  - Many of the previous settings that were removed are now redundant
    with either the `strict: true` or with the current `target`, `module`,
    and `moduleResolution` settings.

  - The `ts-node` section has been added to allow the webpack configuration
    files to be written in typescript.

### webpack config refactoring
  - Use `pathTo()` instead of `path.resolve(__dirname)`.  This
    will help any migration to use ESM modules in future.

  - Setup `output` file names for dev and prod builds

  - Have monaco workers generate to `monaco/...`

  - Use `CaseSensitivePathsPlugin` since it is a devDep but
    currently unused

  - Use consistent ordering for imports

### updates to `client/package.json`
  - Add script `build:dev`

  - Remove `GENERATE_SOURCEMAP=true` from the `start:dev` script since
    it does not impact the build at all

  - Rearrange the `prebuild` script, and force `--noEmit`

  - Drop dependencies that are not being used:
    - `webpack-manifest-plugin`
    - `react-dev-utils`
    - `react-refresh`

### DIFF
--- a/client/config/webpack.common.ts
+++ b/client/config/webpack.common.ts
@@ -1,26 +1,27 @@
 import path from "path";
+import { Configuration, WatchIgnorePlugin } from "webpack";
+import CaseSensitivePathsPlugin from "case-sensitive-paths-webpack-plugin";
 import CopyPlugin from "copy-webpack-plugin";
 import Dotenv from "dotenv-webpack";
 import { TsconfigPathsPlugin } from "tsconfig-paths-webpack-plugin";
-import { Configuration, WatchIgnorePlugin } from "webpack";
 import MonacoWebpackPlugin from "monaco-editor-webpack-plugin";
+
 import { LANGUAGES_BY_FILE_EXTENSION } from "./monacoConstants";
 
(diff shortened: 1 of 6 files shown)


mturley:
This is **_awesome_**. I want to spend a little more time playing with it and then will give a review

mturley (approved):
Looks great to me @sjd78.

PR #1078, opened 2023-06-29
Migration to PF V5

gildub:
The initial driver behind the need for PatternFly Version 5 is for the `datePicker` component which get bug fixes.
But we'll need to migration to V5 anyway sooner than later.

Some component (Table, Select, Dropdown, Wizard) have v4 version maintained in @patternfly/react-core/deprecated allowing for a smooth transition and make sure there is no visual changes. 
Once migrated we can progressively move to the newer components.

This PR addresses the 150+ errors initially triggered by porting to pre-release V5.


mturley:
@ibolton336 for certain components that had major changes, they released them early under `@patternfly/react-core/next`. But I hadn't realized that was only for the new `Dropdown`, `Select` and `Wizard`. https://github.com/patternfly/patternfly-react/tree/v4/packages/react-core/src/next/components

client/src/app/app.css
```diff
@@ -1,63 +1,64 @@
...
-  color: var(--pf-c-icon--m-warning--Color);
+.pf-v5-c-icon .pf-v5-m-warning {
+  color: var(--pf-v5-c-icon--m-warning--Color);
 }
 
-.pf-c-icon .pf-m-danger {
-  color: var(--pf-c-icon--m-danger--Color);
+.pf-v5-c-icon .pf-v5-m-danger {
+  color: var(--pf-v5-c-icon--m-danger--Color);
 }
```
mturley, line 49: I'm surprised that this CSS was ever necessary. I wonder if we just remove all the `pf-v5-c-icon` stuff here if icons work as expected?

@gildub it looks like this was added in https://github.com/konveyor/tackle2-ui/pull/300/files#diff-acb6efd446c502574b8cfc71c2c9fda16948c4811f0040ed7efcbfd1ebdee22f, do you remember why it was needed?

mturley (approved):
I haven't found any more glaring behavior regressions yet. I think this is at a point where I would be comfortable merging it as long as we make a followup effort over the next few days to thoroughly compare the before/after visually and open issues for each visual/style regression, then prioritize getting those fixed before long.

client/src/app/app.css
gildub: Thanks for point it that.
I'll check.

gildub: @mturley, Effectively those classes seems unnecessary, I removed them. I can't remember why I added them in the first place.

Issue #1067, opened 2023-06-23
Jira configuration: Not connected message needs adjustment 


gildub:
The error message of a failed connection can be quite long and needs to be display correctly:

![image](https://github.com/konveyor/tackle2-ui/assets/1901741/b2221097-4d24-480b-bcd1-2056e1eaa956)

sjd78:
To get this crazy looking error:
 1. Create a Jira credential with something that will not work (bearer or basic doesn't matter)
 2. Create a new jira configuration to a real jira instance
 3. Use the bad credential created in 1
 4. Wait a bit and the "Not connected" popover will populate with a crazy long error message

When I took a look at the error message, it turns out to be the full HTTP response with error code 403...

A snip of what I saw:
```
403 Forbidden: 










<html>

<head>
    <title>Forbidden (403)</title>
    





[removed the script blocks]

    <meta name=\"application-name\" content=\"JIRA\" data-name=\"jira\" data-version=\"9.4.2\">
</head>
<body id=\"jira\" class=\"aui-layout aui-style-default page-type-message \" data-version=\"9.4.2\" >

<!-- HTTPError403 -->
    <div class=\"aui-page-panel\" ><div class=\"aui-page-panel-inner\">
            <main role=\"main\" id=\"main\" class=\"aui-page-panel-content\" >
                    <div class=\"aui-page-header\" ><div class=\"aui-page-header-inner\">
                            <div class=\"aui-page-header-main\" >
                                    <h1>Forbidden (403)</h1>
                                </div>
                        </div></div>
                    <div class=\"aui-message aui-message-warning warning\">
                            <p>Encountered a <code>&quot;403 - Forbidden&quot;</code> error while loading this page.</p>
                            <p>Basic Authentication Failure - Reason : AUTHENTICATION_DENIED</p>
                            <p><a href=\"/secure/MyJiraHome.jspa\">Go to Jira home</a></p>
                        </div>
                </main>
        </div></div>
</body>
</html>
: request failed. Please analyze the request body for more details. Status code: 403
```

I can adjust the display to handle content like this.

However, why is content like this even passed back to us in the REST api as a message?  Does it make sense to open a related issue on hub?

cc: @gildub , @ibolton336

PR #1079, opened 2023-06-29
:bug: Update the `TrackerStatus` popover to handle long error messages

sjd78:
Since a Jira tracker "Not connected" error message can be an unlimited length, update the popover such that:
  - The width will be between 30em and 50em
  - The message height itself will be a maximum of 450px tall
  - Message content will overflow with scroll bars as needed
  - If the message is longer than 300 characters, an expandable toggle is added to the code block header to allow showing the message in full

These changes should allow long messages to be displayed in a reasonable way.

Resolves: #1067

### DIFF
--- a/client/public/locales/en/translation.json
+++ b/client/public/locales/en/translation.json
@@ -309,6 +309,8 @@
     "scheduled": "Scheduled",
     "select": "Select",
     "settingsAllowApps": "Allow reviewing applications without running an assessment first",
+    "showLess": "Show less",
+    "showMore": "Show more",
     "source": "Source",
     "sourceBranch": "Branch",
     "sourceCode": "Source code",
--- a/client/src/app/pages/external/jira/components/tracker-status.css
+++ b/client/src/app/pages/external/jira/components/tracker-status.css
@@ -1,0 +1,9 @@
+.tracker-status-code {
+  min-width: 30em;
+  max-width: 60em;
+}
+
+.tracker-status-code .pf-c-code-block__content {
+  max-height: 450px;
+  overflow: auto;
+}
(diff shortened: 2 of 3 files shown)


sjd78:
# Screenshots

Short error message:
screenshot-localhost_9000-2023 06 29-17_57_06

Long error message (the same one displayed in https://github.com/konveyor/tackle2-ui/issues/1067#issuecomment-1613747587) initially displayed with truncated content:
screenshot-localhost_9000-2023 06 29-17_57_17

Long error message with expanded content (scrolled to the end of content):
screenshot-localhost_9000-2023 06 29-17_57_26

mturley:
Hmm, I wonder if we can do anything better than display the HTML to the user? we could clean it with something like [sanitize-html](https://www.npmjs.com/package/sanitize-html) and render it, but then we'd get a "Go to Jira home" link in it that we don't want the user to click. I guess any further parsing of that HTML means we're hard-coding assumptions about the response body from Jira. Maybe nothing to be done, just thinking out loud

sjd78:
> Hmm, I wonder if we can do anything better than display the HTML to the user? we could clean it with something like [sanitize-html](https://www.npmjs.com/package/sanitize-html) and render it, but then we'd get a "Go to Jira home" link in it that we don't want the user to click. I guess any further parsing of that HTML means we're hard-coding assumptions about the response body from Jira. We could disallow/remove `<a>` tags when santizing? We could also just extract any text nodes (maybe excluding `<a>`) and render them in simple paragraphs instead of trying to render the given HTML structure.
> 
> Maybe nothing to be done, just thinking out loud

I'd rather have the backend save a meaningful erorr message instead of the full HTTP response when it encounters a 403 response code trying to connect to a service.  I see this change as a stop-gap until the error message can make sense to an end user.

See https://github.com/konveyor/tackle2-ui/issues/1067#issuecomment-1613747587

mturley:
Agreed, that would be better. Sounds good for a stopgap.

mturley (approved):

PR #1081, opened 2023-06-30
:bug: Handle uploaded yaml with a single parsed value

ibolton336:
- Fixes a bug with the rules selector that was writing ruleset data to each rule in the map. This caused all uploaded rules to show the ruleset name rather than the rule name. 
- Fixes an issue with yaml parsing. yaml.load returns an object when only one value is found rather than an array when multiple values are found from an uploaded rule file. This caused the app to blow up & freeze when trying to iterate through the single object. 
- TODO: Fix the enabled save button on edit with no dirty fields for the custom-target-form #1073

### DIFF
--- a/client/src/app/common/CustomRules/rules-utils.tsx
+++ b/client/src/app/common/CustomRules/rules-utils.tsx
@@ -11,15 +11,21 @@ export const checkRuleFileType = (filename: string): RuleFileType => {
     return "XML";
   } else return null;
 };
+type ParsedYamlElement = { labels?: string[] };
+type ParsedYaml = ParsedYamlElement[] | {};
 
 export const parseRules = (file: IReadFile): ParsedRule => {
   if (file.data) {
     if (checkRuleFileType(file.fileName) === "YAML") {
-      const yamlDoc = yaml.load(file.data) as any[];
-      const yamlLabels = yamlDoc?.reduce((acc, parsedLine) => {
-        const newLabels = parsedLine?.labels ? parsedLine?.labels : [];
-        return [...acc, ...newLabels];
-      }, []);
+      const yamlDoc: ParsedYaml = yaml.load(file.data) as ParsedYaml;
+      const yamlList = Array.isArray(yamlDoc) ? yamlDoc : [yamlDoc];
+      const yamlLabels = Array.from(
+        new Set(
+          yamlList?.flatMap((parsedLine) => {
+            return parsedLine?.labels ? parsedLine?.labels : [];
+          }) || []
+        )
+      );
       const allLabels = getLabels(yamlLabels);
       return {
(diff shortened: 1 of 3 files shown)


client/src/app/common/CustomRules/rules-utils.tsx
```diff
@@ -11,12 +11,13 @@ export const checkRuleFileType = (filename: string): RuleFileType => {
     return "XML";
   } else return null;
 };
-
+type ParsedYaml = any[] | {};
```
mturley, line 14: Since we are accessing the `labels` property on the parsed YAML array elements, can we use a type here that contains that property instead of `any`? Maybe something like:

```ts
type ParsedYamlElement = { labels?: string[]; };
type ParsedYaml = ParsedYamlElement[] | {};
```

client/src/app/common/CustomRules/rules-utils.tsx
```diff
@@ -11,12 +11,13 @@ export const checkRuleFileType = (filename: string): RuleFileType => {
...
 export const parseRules = (file: IReadFile): ParsedRule => {
   if (file.data) {
     if (checkRuleFileType(file.fileName) === "YAML") {
-      const yamlDoc = yaml.load(file.data) as any[];
-      const yamlLabels = yamlDoc?.reduce((acc, parsedLine) => {
+      const yamlDoc: ParsedYaml = yaml.load(file.data) as ParsedYaml;
+      const yamlList = Array.isArray(yamlDoc) ? yamlDoc : [yamlDoc];
+      const yamlLabels = yamlList?.reduce((acc, parsedLine) => {
```
mturley, line 20: Also (and it's fine if we leave this because the `reduce` was pre-existing) if your reduce is just adding elements to an accumulated array, you can simplify it by using [flatMap](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flatMap) instead. Each iteration of a flatMap returns an array, and the arrays are all flattened/appended together at the end.

```ts
const yamlLabels = yamlList?.flatMap((parsedLine) => {
  return parsedLine?.labels ? parsedLine?.labels : [];
}) || [];
```

Note that this is the same as `.map(...).flat(1)`.

Also, looking at the behavior of `getLabels`, if the same label appears more than once in the YAML it's going to end up duplicated in `otherLabels` and `allLabels`, right? I wonder if that's a problem. If it is, we could deduplicate the array by using a Set.

```ts
const yamlLabels = Array.from(new Set(yamlList?.flatMap((parsedLine) => {
  return parsedLine?.labels ? parsedLine?.labels : [];
}) || []);
```

mturley (commented):
A couple thoughts -- might not be a problem?

mturley (approved):
LGTM!

PR #1082, opened 2023-06-30
:bug: Save button always enabled for custom-target edit

ibolton336:
- Fixes an issue where the save button is always enabled for custom-target-update form. This is because the default handler calls setValue due to the file upload component form state being manually managed by the useRuleFiles component. This fix bypasses the setValue call for initially loaded placeholder files. 
Closes https://github.com/konveyor/tackle2-ui/issues/1073


mturley (approved):
Ha! that'll do it

PR #1085, opened 2023-07-03
:bug: Pass all associated rule labels to analyzer

ibolton336:
- Passes a list of all labels that are associated with rule files along to the analyzer.


gildub (approved):
LGTM. 
Test wise, it doesn't break anything but I don't have a rule set file to fully exploit those labels.

ibolton336:
> LGTM. Test wise, it doesn't break anything but I don't have a rule set file to fully exploit those labels.

Thanks Gilles - I have been using the rule files here: https://github.com/pranavgaikwad/rulesets/tree/main/default/generated/eap7

PR #1098, opened 2023-07-06
:sparkles: Tags: Mark stall data after create/edit mutation

gildub:
Resolves https://issues.redhat.com/browse/MTA-802

### DIFF
--- a/client/src/app/queries/tags.ts
+++ b/client/src/app/queries/tags.ts
@@ -31,7 +31,7 @@ export const useFetchTags = () => {
 
 export const useFetchTagCategories = () => {
   const { data, isLoading, error, refetch } = useQuery({
-    queryKey: [TagCategoriesQueryKey],
+    queryKey: [TagCategoriesQueryKey, TagsQueryKey],
     queryFn: getTagCategories,
     onError: (error: AxiosError) => console.log("error, ", error),
   });
@@ -53,13 +53,11 @@ export const useCreateTagMutation = (
     mutationFn: createTag,
     onSuccess: (res) => {
       onSuccess(res);
-      queryClient.invalidateQueries([TagsQueryKey]);
-      queryClient.invalidateQueries([TagCategoriesQueryKey]);
+      queryClient.invalidateQueries([TagCategoriesQueryKey, TagsQueryKey]);
     },
     onError: (err: AxiosError) => {
       onError(err);
-      queryClient.invalidateQueries([TagsQueryKey]);
-      queryClient.invalidateQueries([TagCategoriesQueryKey]);
+      queryClient.invalidateQueries([TagCategoriesQueryKey, TagsQueryKey]);
     },
   });
 };
@@ -75,12 +73,10 @@ export const useCreateTagCategoryMutation = (
     onSuccess: (res) => {
       onSuccess(res);
       queryClient.invalidateQueries([TagCategoriesQueryKey]);
-      queryClient.invalidateQueries([TagsQueryKey]);
     },
     onError: (err: AxiosError) => {
       onError(err);
       queryClient.invalidateQueries([TagCategoriesQueryKey]);
-      queryClient.invalidateQueries([TagsQueryKey]);
     },
   });
 };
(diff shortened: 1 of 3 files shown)


client/src/app/queries/tags.ts
```diff
@@ -54,10 +54,12 @@ export const useCreateTagMutation = (
     onSuccess: (res) => {
       onSuccess(res);
       queryClient.invalidateQueries([TagsQueryKey]);
+      queryClient.invalidateQueries([TagCategoriesQueryKey]);
```
sjd78, line 57: Since most of the mutations seem to need to invalidate both queries, why not just have a helper to make sure things are handled that same way for all?

E.g.
```ts
const useInvalidatingMutation = <T>(
  mutationFn: MutationFunction<T, T>,
  onSuccess: (res: any) => void,
  onError: (err: AxiosError) => void
) => {

  return useMutation({
    mutationFn,
    onSuccess: (res) => {
      onSuccess(res);
      queryClient.invalidateQueries([TagsQueryKey]);
      queryClient.invalidateQueries([TagCategoriesQueryKey])
    },
    onError: (err: AxiosError) => {
      onError(err);
      queryClient.invalidateQueries([TagsQueryKey]);
      queryClient.invalidateQueries([TagCategoriesQueryKey])
    },
  });
}
```

And then this mutation can just look like:
```ts
export const useCreateTagMutation = (
  onSuccess: (res: any) => void,
  onError: (err: AxiosError) => void
) => {
  return useInvalidatingMutation(createTag, onSuccess, onError);
};
```

sjd78 (commented):
Just a suggestion

client/src/app/queries/tags.ts
mturley: I would say we should perhaps include an additional `mutationOptions` object that is a `Partial<UseMutationOptions>` and gets spread into the inner useMutation, or perhaps just have the arguments for `useInvalidatingMutation` be themselves the same `UseMutationOptions` object, but the onSuccess and onError just get overridden? TypeScript generics might get weird I suppose.

Also, we may want to consider using compound query key arrays that contain some common element we can pass to invalidate any query that has to do with tags. The `queryKey` can be an array with any number of elements to identify the query, and if any of those elements is passed into `invalidateQueries` it will invalidate all matching queries. https://tanstack.com/query/v4/docs/react/guides/query-keys

That would at least mean we can do a single `queryClient.invalidateQueries([TagsQueryKey])` if we just include `TagsQueryKey` along with `TagCategoriesQueryKey` in the tag categories query hook. It still doesn't prevent all the duplication @sjd78's suggestion would address though, so maybe we can do both.

gildub: Thanks @sjd78 and @mturley , for the review.

I've pushed an update using compound query key which simplifies and therefore it doesn't need to use an helper.
Effectively on the tags page, the tags are depending upon the tagCategories since we're displaying the tags for each tag category they belong to, therefore when a tag is touched (Created/updated/deleted) then we want to invalidate the same query key we use for fetching the tagCategories.

client/src/app/queries/tags.ts
```diff
@@ -53,11 +53,11 @@ export const useCreateTagMutation = (
     mutationFn: createTag,
     onSuccess: (res) => {
       onSuccess(res);
-      queryClient.invalidateQueries([TagsQueryKey]);
+      queryClient.invalidateQueries([TagCategoriesQueryKey, TagsQueryKey]);
```
sjd78, line 56: I debugged through the react-query code -- this only invalidates the query that matches both keys.

Effectively, this change invalidates the `useFetchTagCategories` query now instead of the `useFetchTags` query.

sjd78: AFAICT, the only way to invalidate 2 queries is to either:
 1. call invalidate twice with exactly matching `queryKey`, or
 2. setup multiple values in the initial `queryKey` and use one of those across all queries you want to invalidate at the same time.

For example:
```tsx
export const useFetchTags = () => {
  const { data, isLoading, error, refetch } = useQuery({
    queryKey: ["allTags", TagsQueryKey],
    queryFn: getTags,
  });
  return {
    tags: data || [],
    isFetching: isLoading,
    fetchError: error,
    refetch,
  };
};

export const useFetchTagCategories = () => {
  const { data, isLoading, error, refetch } = useQuery({
    queryKey: ["allTags", TagCategoriesQueryKey],
    queryFn: getTagCategories,
  });
  return {
    tagCategories: data || [],
    isFetching: isLoading,
    fetchError: error,
    refetch,
  };
};

export const useCreateTagMutation = (
  onSuccess: (res: any) => void,
  onError: (err: AxiosError) => void
) => {

  return useMutation({
    mutationFn: createTag,
    onSuccess: (res) => {
      onSuccess(res);
      queryClient.invalidateQueries(["allTags"]);
    },
    onError: (err: AxiosError) => {
      onError(err);
      queryClient.invalidateQueries(["allTags"]);
    },
  });
};
```

The extra `"allTags"` shouldn't do anything to the query itself other than make it easier to partial match on the `invalidateQueries()` call

sjd78 (changes requested):
See inline comments

sjd78 (approved):
With invalidation only needed on the tag category query, this change LGTM.

client/src/app/queries/tags.ts
mturley: Yeah, option 2 here is what I meant above.

PR #1103, opened 2023-07-10
:bug: Always show analysis details in drawer

### LINKED ISSUE #1013: Don't show drawer spinner for ongoing analysis run
We should be able to view analysis report details while the analysis is running. We should link to app details like we currently do in the kebab menu for analysis details.

### LINKED ISSUE #1100: Facts tab shows when no facts are present in the app drawer

ibolton336:
- It was reported (no bz) that the analysis details should always be accessible from the side drawer. This PR removes the spinner for in-progress analysis status & replaces it with a link to the code viewer modal. Resolves #1013 
- Also, this PR hides the facts tab when no facts are present. Resolves #1100

### DIFF
--- a/client/src/app/pages/applications/components/application-detail-drawer/application-detail-drawer-analysis.tsx
+++ b/client/src/app/pages/applications/components/application-detail-drawer/application-detail-drawer-analysis.tsx
@@ -25,6 +25,7 @@ import { useFetchFacts } from "@app/queries/facts";
 import { ApplicationFacts } from "./application-facts";
 import { SimpleDocumentViewerModal } from "@app/shared/components/simple-task-viewer";
 import { getApplicationAnalysis, getTaskById } from "@app/api/rest";
+import { COLOR_NAMES_BY_HEX_VALUE } from "@app/Constants";
 
 export interface IApplicationDetailDrawerAnalysisProps
   extends Pick<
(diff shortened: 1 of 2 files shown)


ibolton336:
Screenshot 2023-07-10 at 12 24 23 PM

client/src/app/pages/applications/components/application-detail-drawer/application-detail-drawer-analysis.tsx
```diff
@@ -134,25 +131,50 @@ export const ApplicationDetailDrawerAnalysis: React.FC<
...
+                  variant="link"
+                  onClick={() => setTaskIdToView(task?.id)}
+                  className={spacing.ml_0}
+                  style={{ margin: "0", padding: "0" }}
+                >
+                  Analysis details
+                </Button>
+              ) : null}
+
+              {notAvailable}
```
mturley, line 164: This `notAvailable` should be on the else end of the `task ? <Button> : null` ternary instead of null, right? Currently you get both the "Analysis details" button and then "not available" text below it.

mturley (changes requested):

client/src/app/pages/applications/components/application-detail-drawer/application-detail-drawer-analysis.tsx
ibolton336: Yeah - I was wondering what we should show in this case for the "catch all status". I think your suggestion makes sense.

PR #1104, opened 2023-07-10
:bug: Fix double tab pane in issue incidents detail modal from PF upgrade

mturley:
I missed this when reviewing #1078. cc @gildub

I noticed that when clicking an affected file in the drawer on the Issues pages to open the incident details modal, the last tab ("All incidents" table) was split out into its own tab bar and rendered below the contents of the selected tab. This is because of this change: https://github.com/konveyor/tackle2-ui/pull/1078/files#diff-e40974378de11be89b5b1ca02db26b8fed80482f468154ebb84807af7f0e7273L125-R147 (link takes a while to scroll the diff after loading)

When I reversed the change, I saw the reason for it: the new Tabs component doesn't like to have its children expressed as a combination of an array and extra nodes (like we had been doing when mapping over the first 5 incidents and conditionally including a 6th tab). It gave this error:

Screenshot 2023-07-10 at 3 38 00 PM

It appears to be due to this type change in PF: https://github.com/patternfly/patternfly-react/pull/8217

The solution was to make sure the children of `<Tabs>` is always a single array of `<Tab>` elements.

### DIFF
--- a/client/src/app/pages/issues/issue-detail-drawer/file-incidents-detail-modal/file-incidents-detail-modal.tsx
+++ b/client/src/app/pages/issues/issue-detail-drawer/file-incidents-detail-modal/file-incidents-detail-modal.tsx
@@ -94,7 +94,7 @@ export const FileIncidentsDetailModal: React.FC<
           }
         >
           {[
-            ...firstFiveIncidents.map((incident, index) => (
+            firstFiveIncidents.map((incident, index) => (
               <Tab
                 key={incident.id}
                 eventKey={incident.id}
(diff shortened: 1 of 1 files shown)


ibolton336 (approved):

client/src/app/pages/issues/issue-detail-drawer/file-incidents-detail-modal/file-incidents-detail-modal.tsx
```diff
@@ -121,30 +123,26 @@ export const FileIncidentsDetailModal: React.FC<
...
+                      isInline
+                      variant="info"
+                      className={spacing.mtMd}
+                      title="TODO"
+                    />
+                    <FileAllIncidentsTable fileReport={fileReport} />
+                  </Tab>,
+                ]
+              : []),
+          ]}
```
sjd78, line 144: As an alternative to the array spread (which does look ok here), you could throw in a `.flat().filter(Boolean)` and not worry about it:

```tsx
  {[
    firstFiveIncidents.map((incident, index) => ( /* snip */ )),
    totalNumIncidents > 5 && [ /* snip */ ],
  ].flat(1).filter(Boolean)}
```

sjd78 (commented):
Just a thought on an alternate form

client/src/app/pages/issues/issue-detail-drawer/file-incidents-detail-modal/file-incidents-detail-modal.tsx
mturley: Ah, that's a good idea! Nice and clean.

sjd78 (approved):
So pretty!  LGTM

PR #1110, opened 2023-07-11
:bug: [release-0.2] Fix state management and validation of date pickers in migration wave modal

mturley:
Currently, the date pickers in the create/edit migration wave modal misbehave when the user attempts to enter a date string manually.
* The required MM/DD/YYYY format is not enforced.
* Invalid date formats are coerced to the nearest matching date, which leads to inaccurate validation running before the user is done typing.
* Entering a start date that is out of the allowable range (before the current day) disables the field.

The problem seemed to mostly stem from the fact that these dates were stored in form state as `Date` objects, which made it impossible to reference the user's entered string in validation if it was formatted incorrectly. This PR changes the form state to use strings instead of Dates, updates the validation accordingly, and transforms the strings into the correct format on submit.

Note also that the interdependent validation has been changed: before, there was a validation error on the start date if it was after the end date AND a validation error on the end date if it was before the start date. Since these validations have been moved to use `.when` and `.test` (because `.min` with `yup.ref` is unreliable when the value could be in an invalid format), yup gives an error if there is a circular dependency in validation. To address this, only the end date field is validated to enforce the requirement that it is after the start date, and if the start date is changed, validation is re-triggered on the end date.

### DIFF
--- a/client/src/app/pages/migration-waves/components/migration-wave-form.tsx
+++ b/client/src/app/pages/migration-waves/components/migration-wave-form.tsx
@@ -24,6 +24,7 @@ import {
 } from "@app/queries/migration-waves";
 import dayjs from "dayjs";
 import utc from "dayjs/plugin/utc";
+import customParseFormat from "dayjs/plugin/customParseFormat";
 import {
   Stakeholder,
   StakeholderGroup,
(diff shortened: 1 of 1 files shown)


djzager (approved):

sjd78 (approved):
LGTM

gildub (commented):
LGTM and works well.
Thanks @mturley

PR #1124, opened 2023-07-12
:bug: Fix aggregated wave status. Reflect all states in Jira instance 

ibolton336:
Resolves https://issues.redhat.com/browse/MTA-877
Resolves https://github.com/konveyor/tackle2-ui/issues/974 https://github.com/konveyor/tackle2-ui/issues/911

- Adds Completed status
- Adds In Progress status
- Filters out empty status from derived status calculation. This could mean the issue has been deleted from the underlying JIRA instance OR that the Jira ticket is in the process of being created. We have no way of knowing how to interpret empty string, so this filters it out which gives us a more accurate representation of aggregated status.
- Add TODO for info text around auth types for JIRA credential creation
- Remove Created By field in identity page table
- Fixes stale tickets data in migration waves table by mounting the tickets query

### DIFF
--- a/client/src/app/queries/migration-waves.ts
+++ b/client/src/app/queries/migration-waves.ts
@@ -8,7 +8,10 @@ import {
   deleteAllMigrationWaves,
 } from "@app/api/rest";
 import { getWavesWithStatus } from "@app/utils/waves-selector";
-import { TicketsQueryKey } from "./tickets";
+import { useFetchTickets } from "./tickets";
+import { TrackersQueryKey } from "./trackers";
+import { useFetchApplications } from "./applications";
+import { useFetchStakeholders } from "./stakeholders";
 
 export const MigrationWavesQueryKey = "migration-waves";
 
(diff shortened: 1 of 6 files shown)


client/src/app/queries/migration-waves.ts
```diff
@@ -35,7 +36,8 @@ export const useFetchMigrationWaves = () => {
     queryFn: getMigrationWaves,
     refetchInterval: 5000,
     onError: (error) => console.log("error, ", error),
-    onSuccess: () => queryClient.invalidateQueries([TicketsQueryKey]),
+    onSuccess: () =>
+      queryClient.invalidateQueries([TicketsQueryKey, TrackersQueryKey]),
```
mturley, line 40: @ibolton336 does this behave as expected? I think the invalidateQueries array param is intended to be components that have to all be part of a queryKey to invalidate that query, so this should probably be two separate `invalidateQueries` calls for tickets and trackers. Confusing API!

sjd78: Same kind of discussion as https://github.com/konveyor/tackle2-ui/pull/1098/files/83a6edbd8d8db959d1aecd53e1901d213c4cb13d#r1259920085.

"Quick" fix is to have all target queries have a single shared element in the key array and then partial matching do its thing.

FWIW, debugging in to react-query is pretty easy.

client/src/app/utils/waves-selector.ts
```diff
@@ -143,6 +149,7 @@ export const getWavesWithStatus = (
   };
   const wavesWithStatus: WaveWithStatus[] = waves.map(
     (wave): WaveWithStatus => {
+      console.log({ ticketStatus: aggregatedTicketStatus(wave, tickets) });
```
sjd78, line 152: Do you want to keep the log there?

client/src/app/utils/waves-selector.ts
```diff
@@ -56,14 +49,16 @@ export const getWavesWithStatus = (
...
+    wave.applications
+      .map((application): TicketStatus => {
+        const matchingTicket = getTicketByApplication(tickets, application.id);
+        if (matchingTicket?.error) {
+          return "Error";
+        } else if (matchingTicket?.status) {
+          return matchingTicket.status;
+        } else return "";
+      })
+      .filter((ticketStatus) => ticketStatus !== "");
```
sjd78, line 61: `.filiter(Boolean)` will remove anything that is falsey ("", 0, false, undefined, null)

mturley (approved):

PR #1133, opened 2023-07-13
:seedling: Adopt `react-refesh` in dev mode to allow hot component replacement

sjd78:
By adopting react-refresh in webpack dev configurations, via `ReactRefreshWebpackPlugin` and `ReactRefreshTypeScript`, developers can now live edit react components without loosing state.  This means components are replaced on change without forcing a page reload.

The `webpack-dev.ts` configuration is using `mergeWithRules()` to both merge the common configuration, and to replace the `ts-loader` options.  The dev options enable react-refresh.

Add `ForkTsCheckerWebpackPlugin` to dev configuration to do async typescript type checking.  It isn't needed on the prod config since the build runs full `tsc` before running webpack.  Any type errors will stop the build before progressing to webpack.

In all webpack configs, `module/rules` now come before `plugins` to maintain visual consistency.

### DIFF
--- a/client/config/webpack.common.ts
+++ b/client/config/webpack.common.ts
@@ -25,12 +25,13 @@ const config: Configuration = {
   module: {
     rules: [
       {
-        test: /\.tsx?$/,
-        loader: "ts-loader",
+        test: /\.[jt]sx?$/,
         exclude: /node_modules/,
-        options: {
-          // disable type checker for fork-ts-checker-webpack-plugin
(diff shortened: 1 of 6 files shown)


sjd78:
An example of changing a component that does not force a page reload:

https://github.com/konveyor/tackle2-ui/assets/3985964/f86c9f8b-fe1f-484b-adfc-7d342502c394

mturley (approved):
This is the dream right here

PR #1149, opened 2023-07-17
:bug: Applications: bulk delete disabled when no selection

gildub:
Resolves https://issues.redhat.com/browse/MTA-898

Note: This fix work on PF V4 but the tooltip is not displayed on PF V5 therefore this fix addresses the issue on release-0.2 but we need to lodge a bug PF V5 for main branch.


client/src/app/pages/applications/applications-table-analyze/applications-table-analyze.tsx
```diff
@@ -419,6 +419,9 @@ export const ApplicationsTableAnalyze: React.FC = () => {
     importWriteAccess = checkAccess(userScopes, importsWriteScopes),
     applicationWriteAccess = checkAccess(userScopes, applicationsWriteScopes);
 
+  const areAppsInWaves = () =>
```
mturley, line 422: This doesn't need to be a function, it can just be a boolean variable. It is immediately called twice in the same render when it only needs to run the `some` loop once

client/src/app/pages/applications/applications-table-assessment/applications-table-assessment.tsx
```diff
@@ -608,6 +608,9 @@ export const ApplicationsTable: React.FC = () => {
     importWriteAccess = checkAccess(userScopes, importsWriteScopes),
     applicationWriteAccess = checkAccess(userScopes, applicationsWriteScopes);
 
+  const areAppsInWaves = () =>
```
mturley, line 611: Same here

mturley (commented):
Just a minor nitpick

mturley (approved):
LGTM 👍

PR #1150, opened 2023-07-17
:bug: Restore deprecated `Dropdown` in `KebabDropdown`, use matching deprecated `DropdownItems`, use `isAriaDisabled` prop for items with tooltips

mturley:
Fixes a regression caused by #1126 where dropdown items that are disabled with tooltips could not trigger the tooltips. Discovered by @gildub when working on #1149.

It appears the `isAriaDisabled` prop which we need for retaining the mouse hover event on a disabled DropdownItem is not present in the new PF5 DropdownItem. This is due to missing styles in PF core that are still being addressed (see https://github.com/patternfly/patternfly/issues/5528, https://github.com/patternfly/patternfly/pull/5692, and related comment here: https://github.com/patternfly/patternfly-react/pull/8992#issuecomment-1530234144)

Until `isAriaDisabled` is supported in the new PF5 DropdownItem we must restore the usage of the deprecated component for these tooltips to work.

### DIFF
--- a/client/src/app/components/target-card.tsx
+++ b/client/src/app/components/target-card.tsx
@@ -7,13 +7,13 @@ import {
   Card,
   CardBody,
   Text,
-  DropdownItem,
   Flex,
   FlexItem,
   Button,
   ButtonVariant,
   Label,
 } from "@patternfly/react-core";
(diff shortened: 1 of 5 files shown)


gildub (approved):
LTGM

PR #1152, opened 2023-07-17
:sparkles: Remove empty name column for issues in favor of description column

mturley:
In @pranavgaikwad's demo we saw that the issue `name` field is not useful and perhaps always empty in real data, and instead the first line of the `description` field is the more meaningful identifier for a row in this table. This change removes the name column and renames the description column header to "Issue" as discussed.

Screenshot 2023-07-17 at 1 48 36 PM


gildub (approved):
LGTM

PR #1154, opened 2023-07-17
:bug: Adjust Dropdown for deprecated table support in PF V5

gildub:
Disable properly the row delete option (dropdown) when an application is involved in migration wave.

Note: doesn't need to be back-ported

image

### DIFF
--- a/client/src/app/pages/applications/applications-table-analyze/applications-table-analyze.tsx
+++ b/client/src/app/pages/applications/applications-table-analyze/applications-table-analyze.tsx
@@ -356,7 +356,7 @@ export const ApplicationsTableAnalyze: React.FC = () => {
         {
           title: t("actions.delete"),
           ...(row.migrationWave !== null && {
-            isAriaDisabled: true,
+            isDisabled: true,
             tooltip: "Cannot delete application assigned to a migration wave.",
             tooltipProps: { postition: TooltipPosition.top },
           }),
--- a/client/src/app/pages/applications/applications-table-analyze/applications-table-analyze.tsx
+++ b/client/src/app/pages/applications/applications-table-analyze/applications-table-analyze.tsx
@@ -357,8 +357,11 @@ export const ApplicationsTableAnalyze: React.FC = () => {
           title: t("actions.delete"),
           ...(row.migrationWave !== null && {
             isAriaDisabled: true,
-            tooltip: "Cannot delete application assigned to a migration wave.",
-            tooltipProps: { position: TooltipPosition.top, content: "blah" },
+            tooltipProps: {
+              position: TooltipPosition.top,
+              content:
+                "Cannot delete application assigned to a migration wave.",
+            },
           }),
           onClick: () => deleteRow(row),
         }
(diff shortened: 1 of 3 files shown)


mturley (changes requested):
@gildub I'm not sure I understand the purpose of this PR, didn't we just realize we do need the `isAriaDisabled` prop and that's why we are keeping the deprecated Dropdown for now? Testing this branch, the tooltip does not work (where it does without this change)

mturley:
Oh sorry, I was testing the bulk dropdown and not the table row one. Hmm, this does break the tooltip but it fixes the disabled behavior. We need to figure out how to keep both -- maybe we need to custom render a KebabDropdown in a cell on each row instead of using the built in actions array? I suppose that is a big refactor.

gildub:
@mturley , in which case might want to look at migrating applications pages away from the deprecated tables the new table giving us more flexibility for those complicated cases.
What do you think ?

mturley:
Yeah that is definitely something we should do eventually but is probably a lot of work and should be done carefully. I would be in support though if we decided as a team that there is a good time to do it

mturley:
Ideally we can find a way to fix this issue in the short term before we go for the full refactor

gildub:
@mturley,
I'm trying to understand why the tooltip is now broken.
The `IAction` type we use for the action array is coming from "@patternfly/react-table" where the Table used (by AppTableWithControls -> AppTable) is  from "@patternfly/react-table/deprecated".

Even though the implementation of IAction hasn't changed it seems the integration of the two implies a different behavior.
If that's the case I don't see how we can make that work as there is no "deprecated" IAction version.

mturley:
Yeah, I don't think that workaround will work, we want to retain the hover behavior. I think the reason this happened is because the legacy/deprecated Table in PF was refactored to render the composable Table components under the hood, so that it was easier to maintain when there were style changes to table elements. My guess is that this shared rendering code was updated to use the new dropdown components (or something like that).

The solution that comes to mind for me is to stop using the `IAction` objects and instead render our own column that uses our `KebabDropdown` in the `title` property of its cells. That will be a bit of a refactor but it won't be throwaway work because we'll need the dropdown to be in that format when we eventually refactor the whole table to use the composable components.

mturley:
I would be happy to take a crack at that if you want me to @gildub

gildub:
@mturley, in the meantime I got that tracker for the issue about tooltip support on menu for legacy dropdown: https://github.com/patternfly/patternfly-react/issues/9285

gildub:
@mturley, to follow-up and answer you, I'm trying to see if PF fix is going to be addressing the issue soon enough so we don't have to refactor. In case they don't then we'll do as you described, use our own column and dropdown.

gildub:
@mturley, good news, PF team is working on the issue and are hoping to fix it before PF GA.

mturley:
Awesome! maybe we can revert my reverting of the upgrade to the new dropdowns in KebabDropdown after that then 😆

mturley (commented):
LGTM, but deleting and recreating the package-lock file resulted in locking new versions of dependencies unrelated to this fix. Ideally we should keep those to a dedicated PR if we want to bump them. You can solve this by reverting to the prior package-json version:

```
git checkout main package-lock.json
```

and then updating only PF:

```
npm update @patternfly/react-core
```

mturley (approved):
LGTM!

PR #1155, opened 2023-07-18
:sparkles: Use latest prerelease version for PF V5

gildub:
The prerelease version stanza pull the latest prerelease versions of PF allowing us to not have to update those dependencies every couple of days as this is moving quite fast.


client/package.json
```diff
@@ -21,12 +21,12 @@
...
-    "@patternfly/react-code-editor": "5.0.0-prerelease.23",
-    "@patternfly/react-core": "5.0.0-prerelease.23",
-    "@patternfly/react-table": "5.0.0-prerelease.23",
-    "@patternfly/react-tokens": "5.0.0-prerelease.8",
+    "@patternfly/patternfly": "prerelease",
+    "@patternfly/react-charts": "prerelease",
+    "@patternfly/react-code-editor": "prerelease",
+    "@patternfly/react-core": "prerelease",
+    "@patternfly/react-table": "prerelease",
+    "@patternfly/react-tokens": "prerelease",
```
sjd78, line 29: I think the PF guys meant "^5.0.0-prerelease" to get the latest prerelease versions.

From the semver docs:

> ^1.2.3-beta.2 := >=1.2.3-beta.2 <2.0.0-0 Note that prereleases in the 1.2.3 version will be allowed, if they are greater than or equal to beta.2. So, 1.2.3-beta.4 would be allowed, but 1.2.4-beta.2 would not, because it is a prerelease of a different [major, minor, patch] tuple.

sjd78 (changes requested):

client/package.json
sjd78: - https://github.com/npm/node-semver#prerelease-tags

sjd78 (approved):

PR #1156, opened 2023-07-18
:bug: Migration wave name is min 3/max 20 chars only when provided

gildub:
Resolves https://issues.redhat.com/browse/MTA-906

### DIFF
--- a/client/src/app/pages/migration-waves/components/migration-wave-form.tsx
+++ b/client/src/app/pages/migration-waves/components/migration-wave-form.tsx
@@ -153,7 +153,6 @@ export const WaveForm: React.FC<WaveFormProps> = ({
           return true;
         }
       ),
-    }),
     startDateStr: yup
       .string()
       .required(t("validation.required"))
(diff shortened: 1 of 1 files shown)


client/src/app/pages/migration-waves/components/migration-wave-form.tsx
```diff
@@ -136,11 +136,14 @@ export const WaveForm: React.FC<WaveFormProps> = ({
...
-      .min(3, t("validation.minLength", { length: 3 }))
-      .max(120, t("validation.maxLength", { length: 120 })),
+    name: yup.string().when("name", {
+      is: true,
+      then: yup
+        .string()
+        .trim()
+        .min(3, t("validation.minLength", { length: 3 }))
+        .max(120, t("validation.maxLength", { length: 120 })),
+    }),
```
ibolton336, line 146:
```suggestion
    name: yup
      .string()
      .defined()
      .test(
        "min-char-check",
        "Name is invalid. The name must be between 3 and 120 characters ",
        (value) => {
          if (!!value) {
            const schema = yup
              .string()
              .min(3, t("validation.minLength", { length: 3 }))
              .max(120, t("validation.maxLength", { length: 120 }));
            return schema.isValidSync(value);
          }
          return true;
        }
      ),
    }),
```

ibolton336 (approved):

PR #1157, opened 2023-07-18
:ghost: Change default application inventory tab

ibolton336:
- Change default tab to analysis 
Screenshot 2023-07-18 at 3 44 25 PM


gildub:
@ibolton336, I have no pb with with this change, I'm just wondering why analysis tab is becoming the default ?

mturley (approved):

gitdallas (approved):
seems we could make the tabs an object with a few props and use mapping in a couple places rather than hard coded order matching, but just a nit with only 2 items.

PR #1172, opened 2023-07-20
:ghost: Re-enable createdBy field for auth required deployments

ibolton336:
- By default, do not show the createdBy field when Authentication is disabled as this just leads to a table column with auth.noAdmin as the createdBy value for every entry in the table.

https://issues.redhat.com/browse/MTA-1015


gildub (approved):
LGTM

PR #1180, opened 2023-07-20
:bug: Add url validation to rules step

ibolton336:
Resolves https://issues.redhat.com/browse/MTA-985

### DIFF
--- a/client/src/app/utils/utils.test.ts
+++ b/client/src/app/utils/utils.test.ts
@@ -1,9 +1,14 @@
+import { act, render } from "@testing-library/react";
+import { renderHook } from "@testing-library/react-hooks";
 import { AxiosError } from "axios";
 import {
   getAxiosErrorMessage,
   getValidatedFromError,
   getValidatedFromErrors,
   getToolbarChipKey,
+  customURLValidation,
+  gitUrlRegex,
+  standardURLRegex,
 } from "./utils";
 
 describe("utils", () => {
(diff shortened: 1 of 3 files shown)


client/src/app/utils/utils.ts
```diff
@@ -94,10 +94,10 @@ export const getValidatedFromError = (error: unknown | undefined) => {
 };
 
 export const standardURLRegex =
-  /(https?:\/\/(?:www\.|(?!www))[a-zA-Z0-9][a-zA-Z0-9-]+[a-zA-Z0-9]\.[^\s]{2,}|www\.[a-zA-Z0-9][a-zA-Z0-9-]+[a-zA-Z0-9]\.[^\s]{2,}|https?:\/\/(?:www\.|(?!www))[a-zA-Z0-9]+\.[^\s]{2,}|www\.[a-zA-Z0-9]+\.[^\s]{2,})/gi;
+  /(https?:\/\/(?:www\.|(?!www))[a-zA-Z0-9][a-zA-Z0-9-]+[a-zA-Z0-9]\.[^\s]{2,}|www\.[a-zA-Z0-9][a-zA-Z0-9-]+[a-zA-Z0-9]\.[^\s]{2,}|https?:\/\/(?:www\.|(?!www))[a-zA-Z0-9]+\.[^\s]{2,}|www\.[a-zA-Z0-9]+\.[^\s]{2,})/;
```
sjd78, line 97: So that's it?! I see the ignore case flag is redundant.  The global flag removal also makes sense (apart from the back reference craziness) since you want the full string to be a valid url.

WDYT about adding a "^" to the beginning and a "$" to the end to make sure the whole string is exactly a URL?

ibolton336: Just updated with your suggestions.

client/src/app/utils/utils.test.ts
```diff
@@ -89,4 +93,20 @@ describe("utils", () => {
...
+
+    const customRuleRepoURLExample =
+      "https://github.com/ibolton336/tackle-testapp.git";
+    const URLvalidationResult = gitUrlRegex.test(customRuleRepoURLExample);
+    expect(URLvalidationResult).toBe(true);
+
+    const brokenURLExample = "github.com/ibolton336/tackle-testapp.git";
+    const brokenURLvalidationResult = gitUrlRegex.test(brokenURLExample);
+    expect(brokenURLvalidationResult).toBe(false);
+  });
```
sjd78, line 111: I suggest adding a few more strings to the tests, and have an `it()` section for `gitUrlRegex` and `standardURLRegex`.

A few url string suggestions:
  - `""`
  - `"     http://www.foo.bar"`
  - `"     http://www.foo.bar     "`
  - `"www.foo.bar"`
  - `"foo.bar"`
  - `"www.foo.b"`
  - `"foo.ba"`

Having a separate `it()` just for a double test would be good as well:
```js
it("URL should match the same multiple times in a row", () => {
  const url = "https://github.com/ibolton336/tackle-testapp.git";

  expect(standardURLRegex.test(url).toBe(true);
  expect(standardURLRegex.test(url).toBe(true);
  expect(standardURLRegex.test(url).toBe(true);
  expect(standardURLRegex.test(url).toBe(true);
  expect(standardURLRegex.test(url).toBe(true);
});
```

And that test should fail if you put the `/g` back on the pattern...

mturley: jeez, I would not have expected a regex `test` to have state like that

sjd78 (approved):
LGTM!

The tests look good so they capture all of the troubles that have been dealt with. :+1:

PR #1200, opened 2023-07-25
:bug: Delete associated tickets on wave delete

ibolton336:
Resolves https://issues.redhat.com/browse/MTA-1014


gildub:
@ibolton336, I believe this one is on hold until Ramon is back next week.

gildub:
@ibolton336, also we don't want to delete tickets when a migration wave is deleted as the use might want/need to move the applications involved in a different migration wave. 
The tickets, if that approach is confirmed by Ramon, will be cascade deleted when the tracker is removed (and only if there are no migration waves left)

PR #1202, opened 2023-07-25
:bug: Fix crashing app when deleting

ibolton336:
Resolves https://issues.redhat.com/browse/MTA-1041


gildub (approved):
@ibolton, works fine now, just a detail, don't we want to make the refetch() statement last, after the detail drawer is gone ?

PR #1204, opened 2023-07-25
:book: Update README to use operator repo install script

mturley:
Updates the README to use the [`install-tackle.sh` script provided by the operator repo](https://github.com/konveyor/operator/blob/main/hack/install-tackle.sh) instead of us having to separately maintain install instructions that may change.

Also moves the info about configuring `kubectl` for minikube to before the installation step because the script expects `kubectl` on your path.

[The rich diff can be viewed here for easier review](https://github.com/konveyor/tackle2-ui/pull/1204/files?short_path=b335630#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5).

### DIFF
--- a/README.md
+++ b/README.md
@@ -97,6 +97,12 @@ $ export TACKLE_FEATURE_AUTH_REQUIRED=true
 
 #### Running the install script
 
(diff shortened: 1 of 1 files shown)


sjd78 (approved):
Makes sense.  They're reasonable changes to setup the dev env.

Issue #1093, opened 2023-07-05
Hide RuleSets with names that start with a `.`


mansam:
The Hub has a convention of using a leading period to indicate a system-owned resource that should be hidden and read only. We're using this to indicate certain built-in rulesets that other rulesets depend upon. The UI should ignore and not display any RuleSets that begin with a `.`.

PR #1205, opened 2023-07-25
:bug: Hide hidden rulesets from seeded hub list

ibolton336:
Resolves https://github.com/konveyor/tackle2-ui/issues/1093

### DIFF
--- a/client/src/app/pages/migration-targets/migration-targets.tsx
+++ b/client/src/app/pages/migration-targets/migration-targets.tsx
@@ -44,11 +44,7 @@ export const MigrationTargets: React.FC = () => {
   const { t } = useTranslation();
   const { pushNotification } = React.useContext(NotificationsContext);
 
-  const {
-    rulesets,
-    isFetching: isFetchingrulesets,
-    refetch: refetchrulesets,
-  } = useFetchRulesets();
+  const { rulesets, refetch: refetchrulesets } = useFetchRulesets();
 
   const rulesetOrderSetting = useSetting("ui.ruleset.order");
   const rulesetOrderSettingMutation = useSettingMutation("ui.ruleset.order");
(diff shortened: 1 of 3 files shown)


client/src/app/queries/rulesets.ts
```diff
@@ -19,25 +19,25 @@ export const useFetchRulesets = () => {
...
+              };
+
+              return {
+                ...rule,
+                metadata: transformedMetadata,
+              };
+            });
+            return { ...ruleset, rules: mappedRules };
+          })
+          .filter((ruleset) => !ruleset.name.startsWith("."));
```
ibolton336, line 40: Adds this filter for hidden rulesets.

client/src/app/pages/migration-targets/migration-targets.tsx
```diff
@@ -207,28 +203,34 @@ export const MigrationTargets: React.FC = () => {
...
-                      deleteRuleset(matchingRuleset.id);
-                    }
-                  }}
-                />
-              ))}
+              rulesetOrderSetting.data.map((id) => {
+                const matchingRuleset = rulesets.find(
+                  (Ruleset) => Ruleset.id === id
+                );
+                if (matchingRuleset) {
```
ibolton336, line 210: Don't display ordered items which are hidden.

gildub (changes requested):
The system owned resource is hidden in the Administration -> Custom migration targets page but there is a blank target showing in the Set Targets during Application analysis.

The concern is a succession of blank targets if there are several hidden targets:

image

ibolton336:
> The system owned resource is hidden in the Administration -> Custom migration targets page but there is a blank target showing in the Set Targets during Application analysis.
> 
> The concern is a succession of blank targets pushing everything down in case when there are several hidden targets:
> 
> image

Good catch. The blank targets are filtered in the migration targets page, but not in the wizard.

gildub (approved):
LGTM

client/src/app/queries/rulesets.ts
sjd78, line 40: Very very minor nit-pick:  I suggest moving the filter() above the map() so the map() only needs to process things that will end up being selected.

ibolton336: Done!

PR #1207, opened 2023-07-26
:sparkles: Applications: Update Notification and refactors update/create modal

gildub:
While resolving the related notification issues (see associated MTA-1024) this refactors create/update modal.

Includes following changes :

- Push edit/create notifications down to ApplicationForm
- Use onClose to replace onSave and onCancel
- Consolidate New/Edit Modal duplication

Partially address https://issues.redhat.com/browse/MTA-1024

### DIFF
--- a/client/src/app/pages/applications/components/application-form/application-form.tsx
+++ b/client/src/app/pages/applications/components/application-form/application-form.tsx
@@ -274,25 +274,24 @@ export const ApplicationForm: React.FC<ApplicationFormProps> = ({
     }
   };
 
-  const onCreateUpdateApplicationSuccess = (
-    response: AxiosResponse<Application>
-  ) => {
-    if (application) {
-      pushNotification({
-        title: t("toastr.success.save", {
-          type: t("terms.application"),
-        }),
-        variant: "success",
-      });
-    } else {
-      pushNotification({
-        title: t("toastr.success.createWhat", {
-          type: t("terms.application"),
-          what: response.data.name,
-        }),
-        variant: "success",
-      });
-    }
+  const onCreateApplicationSuccess = (response: AxiosResponse<Application>) => {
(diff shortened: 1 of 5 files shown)


client/src/app/pages/applications/components/application-form/application-form.tsx
```diff
@@ -274,8 +274,26 @@ export const ApplicationForm: React.FC<ApplicationFormProps> = ({
     }
   };
 
-  const onCreateUpdateApplicationSuccess = (response: any) => {
-    onSaved(response);
+  const onCreateUpdateApplicationSuccess = (
+    response: AxiosResponse<Application>
```
ibolton336, line 278: I think we could use our custom useUpdateApplicationMutation & useCreateApplicationMutation hooks to 1) narrow this type to something like "AppName" & 2) Pass the name from the mutation function variable object param to the success handler.

ibolton336 (changes requested):
See inline.

client/src/app/pages/applications/components/application-form/application-form.tsx
gildub: @ibolton336, generalizing this approach is, I believe out of scope of this PR, please let's discuss how we are going to do that here : https://github.com/konveyor/tackle2-ui/issues/1214.

ibolton336: Np - we can address the missing entity name in the update notification later. However, I think it would make more sense to have two separate handlers for create / update here since there.

gildub: @ibolton336, I agree it makes sense to have separate handlers. Let me change that.

ibolton336 (approved):

PR #1210, opened 2023-07-26
:sparkles: BusinessServices: Update Notification and refactors update/create modal and queries

gildub:
While resolving the related notification issues (see associated MTA-1024) this also refactors create/update modal and tries to bring more consistency.

Includes following changes :

- Push edit/create notifications down to BusinessServiceForm
- Use onClose to replace onSave and onCancel
- Replace REST functions to use axios instead of deprecated APIClient().
- Use React Query mutations instead of legacy queries
- Consolidate New/Edit Modal duplication

Partially address https://issues.redhat.com/browse/MTA-1024


client/src/app/api/rest.ts
```diff
@@ -721,6 +693,28 @@ export const updateStakeholderGroup = (
...
+
+export const getBusinessServices = (): Promise<BusinessService[]> =>
+  axios.get(BUSINESS_SERVICES).then((response) => response.data);
+
+export const deleteBusinessService = (
+  id: number | string
+): Promise<BusinessService> => axios.delete(`${BUSINESS_SERVICES}/${id}`);
+
+export const createBusinessService = (
+  obj: New<BusinessService>
```
ibolton336, line 706: I would like to see a more descriptive param name here.

client/src/app/pages/controls/business-services/components/business-service-form.tsx
```diff
@@ -94,6 +92,59 @@ export const BusinessServiceForm: React.FC<BusinessServiceFormProps> = ({
...
+      title: t("toastr.success.createWhat", {
+        type: t("terms.businessService"),
+        what: response.data.name,
+      }),
+      variant: "success",
+    });
+    onClose();
+  };
+
+  const onUpdateBusinessServiceSuccess = (
```
ibolton336, line 108: We should be using a combined success handler here like we are for applications. Or at least we should agree on a style and be consistent across slices of the app.

client/src/app/queries/businessservices.ts
```diff
@@ -1,45 +1,75 @@
...
+    onSuccess: (res) => {
+      onSuccess(res);
+      queryClient.invalidateQueries([BusinessServicesQueryKey]);
+    },
+    onError,
+  });
+};
+
+export const useUpdateBusinessServiceMutation = (
+  onSuccess: (res: any) => void,
```
ibolton336, line 59:
```suggestion
  onSuccess: (businessServiceName: string) => void,
```

client/src/app/queries/businessservices.ts
```diff
@@ -1,45 +1,75 @@
...
+  onError: (err: AxiosError) => void
+) => {
+  const queryClient = useQueryClient();
+  return useMutation({
+    mutationFn: updateBusinessService,
+    onSuccess: (res) => {
+      onSuccess(res);
```
ibolton336, line 66:
```suggestion
  onSuccess: (_, vars) => {
      onSuccess(vars.businessService.name);
```

ibolton336 (changes requested):

client/src/app/pages/controls/business-services/business-services.tsx
```diff
@@ -304,16 +272,25 @@ export const BusinessServices: React.FC = () => {
...
+          {
+            what: t("terms.businessService").toLowerCase(),
+          }
+        )}
+        variant="medium"
+        isOpen={isCreateUpdateModalOpen}
+        onClose={closeCreateUpdateModal}
+      >
+        <BusinessServiceForm
+          businessService={
```
ibolton336, line 288: Make prop optional instead of passing undefined.

client/src/app/queries/businessservices.ts
```diff
@@ -1,45 +1,75 @@
...
-  const { data, isLoading, error, refetch } = useQuery<BusinessService[]>(
-    [BusinessServicesQueryKey],
-    async () => (await getBusinessServices()).data,
-    {
-      onError: (error) => console.log("error, ", error),
-    }
-  );
+  const { data, isLoading, error, refetch } = useQuery({
+    queryKey: [BusinessServicesQueryKey],
+    queryFn: getBusinessServices,
```
ibolton336, line 18: Nice! Much easier to follow 👍

client/src/app/pages/controls/business-services/components/business-service-form.tsx
gildub: @ibolton336,

This seems to be out of scope, please https://github.com/konveyor/tackle2-ui/issues/1214 to following-up.

client/src/app/queries/businessservices.ts
gildub: @ibolton336, all queries are using `res` as parameter for onSuccess().
Therefore I don't see the value of such change and the gain in the meaning.
I'm not necessarily against it meanwhile if we decide such change then we'd have to make it consistent across the all queries.

client/src/app/queries/businessservices.ts
gildub: @ibolton336,

This seems to be out of scope, please see why here https://github.com/konveyor/tackle2-ui/issues/1214 to following-up.

client/src/app/pages/controls/business-services/components/business-service-form.tsx
ibolton336: Thx. I read the issue but I still don't see why the application handler is a special case. Maybe it would just be easier to move the on create / on update success handlers to their own functions for the applications? Just for consistency.

client/src/app/queries/businessservices.ts
ibolton336: I don't mind not doing this. Just seemed like an easy solution that would help standardize notifications with more info.

client/src/app/pages/controls/business-services/components/business-service-form.tsx
```diff
@@ -35,7 +35,7 @@ export interface FormValues {
 }
 
 export interface BusinessServiceFormProps {
-  businessService?: BusinessService;
+  businessService: BusinessService | null;
```
ibolton336, line 38: This seems to go against what we have done across the app for optional props. Having the prop be undefined makes sense as it is omitted for new businessServices & doesn't need to be explicitly set.

client/src/app/queries/businessservices.ts
gildub: @ibolton336, I agree but I would like to have an agreed overall approach to address this across the board.

client/src/app/pages/controls/business-services/components/business-service-form.tsx
gildub: @ibolton336, I thought about that but this isn't an optional prop anymore as it was before.
The Form requires the `businessService` prop.
The modal state value  provides either the object value itself or null:
```ts
const businessServiceToUpdate =
    createUpdateModalState !== "create" ? createUpdateModalState : null;
```

client/src/app/queries/businessservices.ts
gildub: @ibolton336, you had a good point with the `any` type.
This isn't the case anymore as there is no need to pass anything for now (see update) until we eventually address that later if we decide to pass object name to update queries.

client/src/app/pages/controls/business-services/components/business-service-form.tsx
gildub: @ibolton336, 
you're right, the Application isn't a special case, I confused it with the MigrationTargets where we have to pass the reference to scroll down the page.
So revisiting this, we want to be consistent and have separate handlers, I'm changing the handlers in https://github.com/konveyor/tackle2-ui/pull/1207.

We might eventually revisit that, consistently of course, or not once we decide to pass along the object name to the save/edit.

ibolton336 (approved):

PR #1191, opened 2023-07-21
:bug: Adjust Tags tab on Application Details Drawer when a Task is running

sjd78:
When an analysis task is running, and the application drawer is open, show both loading text and spinner and the existing tags. Once the analysis is complete, the loading text/spinner will be removed and any new tags will be shown.

Bug-Url: https://issues.redhat.com/browse/MTA-465

The Reports tab is also mentioned on MTA-465.  This was already addressed with #1103.

PR #1216, opened 2023-07-27
:sparkles: Refactor useFetchTasks()'s select function

sjd78:
The motivation for this refactoring came while working on #1191 and trying to understand what tasks were given to the analysis table.  The select function used by `useFetchTasks()` to filter by addon and to keep only the newest tasks per application id was confusing to read as written.  This refactor simplifies the steps involved to use a typical stream processing approach.

Output of the refactored function is functionally identical to the original implementation.


sjd78:
I submitted this PR mostly because I did the refactor when working on the other PR and didn't want to throw it out.

ibolton336 (approved):

PR #1218, opened 2023-07-27
:seedling: 1070 remove lib-ui statusicon dep

### LINKED ISSUE #1070: Duplicate StatusIcon references
We are using a few components with the same name "statusIcon" that operate differently & exist in two different locations (one locally & one in the external lib-ui package)

gitdallas:
closes #1070 

lib-ui statusicon dep replaced with an updated app/shared/components statusicon

### DIFF
--- a/client/src/app/shared/components/status-icon/status-icon.tsx
+++ b/client/src/app/shared/components/status-icon/status-icon.tsx
@@ -1,6 +1,11 @@
 import React from "react";
 import { useTranslation } from "react-i18next";
-import { Flex, FlexItem, SpinnerProps } from "@patternfly/react-core";
+import {
+  Flex,
+  FlexItem,
+  SpinnerProps,
+  TextContent,
+} from "@patternfly/react-core";
 import CheckCircleIcon from "@patternfly/react-icons/dist/esm/icons/check-circle-icon";
 import TimesCircleIcon from "@patternfly/react-icons/dist/esm/icons/times-circle-icon";
 import InProgressIcon from "@patternfly/react-icons/dist/esm/icons/in-progress-icon";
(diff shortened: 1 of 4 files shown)


client/src/app/shared/components/status-icon/status-icon.tsx
sjd78, line 1: Honest the following line really bugs me (you didn't touch it):
```tsx
const Icon = iconList[status].Icon;
```

That assumes the status is going to actually going to be one of the `StatusIconType` strings ... and the typescript type only binds it at runtime.  To be safe and avoid nasty runtime errors, I'd much rather see something like:
```ts
const iconType = iconList[status] ?? iconList.Unknown;
const Icon = iconType.Icon;
```

sjd78 (commented):
Looks ok, just a few comments

client/src/app/shared/components/status-icon/status-icon.tsx
```diff
@@ -80,44 +97,42 @@ export const StatusIcon: React.FC<IStatusIconProps> = ({
...
-      label = t("terms.scheduled");
-      break;
-    default:
-      label = t("terms.unknown");
-      break;
+  if (!label) {
+    switch (status) {
+      case "Canceled":
+        label = t("terms.canceled");
+        break;
```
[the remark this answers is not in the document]
ibolton336: This is some Tackle v1 code. Would be great to implement both of your suggestions here. +1 to getting rid of switch statements & +1 to use the null coalescing operator. I wonder if it would be worth adding some stricter TS/ eslint rules during this downtime between features.

client/src/app/shared/components/status-icon/status-icon.tsx
gitdallas: see comment above - planning to remove this component altogether unless anyone objects

ibolton336 (approved):

sjd78 (approved):
With #1236, LGTM

PR #1226, opened 2023-07-28
:seedling: Bump all PF react-core deps to V5 release

gildub:


sjd78 (approved):
Other than some confusion on my part about why use `^5.1.0` instead of `^5.0.0`, LGTM

Issue #1136, opened 2023-07-13
[PF5 Regression] - Toggle buttons on "Proxy configuration" page don't work


sjd78:
On the Proxy page, the HTTP Proxy and HTTPS Proxy buttons do not toggle when clicked.

punithnayak:
Hey Will work on this issue
![image](https://github.com/konveyor/tackle2-ui/assets/110713318/85b0384a-f96f-4dfb-92a5-e089be324f75)
these buttons ryt?

sjd78:
Resolved by #1230

PR #1230, opened 2023-07-28
:bug: Fix onChange handlers

ibolton336:
- Addresses a small regression that showed up as a part of the pf5 migration/upgrade.


sjd78:
See #1136

gitdallas (approved):

Issue #1234, opened 2023-07-29
Replace useEntityModal with local state approach


gildub:
Replace legacy `useEntityModal` with local state approach equivalent of :`
```ts
  const [createUpdateModalState, setCreateUpdateModalState] = React.useState<
    "create" | JobFunction | null
  >(null);
  const isCreateUpdateModalOpen = createUpdateModalState !== null;
  const jobFunctionToUpdate =
    createUpdateModalState !== "create" ? createUpdateModalState : null;
```
We want to generalize the latter approach we're implementing for consistency.

To track down the removal of `useEntityModal`: 
- [ ] ApplicationAssessmentPageHeader
- [x] ApplicationTableAnalyze 
- [x] ApplicationTableAssessment
- [ ] Finaly remove hook and test file

PR #1235, opened 2023-07-29
:sparkles: Application tables: Refactor useEntityModal

gildub:
Replace useEntityModal with in-file state.

See https://github.com/konveyor/tackle2-ui/issues/1234 for details

Also:
- Add missing notification when managing applications credentials.
- Application-form: Replace Alert with Notification


client/src/app/pages/applications/applications-table-analyze/applications-table-analyze.tsx
```diff
@@ -660,91 +634,46 @@ export const ApplicationsTableAnalyze: React.FC = () => {
...
+        cancelBtnLabel={t("actions.cancel")}
+        onCancel={() => setApplicationsToDelete([])}
+        onClose={() => setApplicationsToDelete([])}
+        onConfirm={() => {
+          let ids: number[] = [];
+          applicationsToDelete.forEach((application) => {
+            if (application.id) ids.push(application.id);
+          });
+          if (ids) bulkDeleteApplication({ ids: ids });
+          setApplicationsToDelete([]);
```
ibolton336, line 674:
```suggestion
          const ids = applicationsToDelete
          .filter(application=>application?.id)
          .map(application => application.id)
          
          //TODO should have some sort of loading state / error state before closing the modal just incase the bulk delete fails. Then we could move the setApplicatoinsToDelete to a success/error callback after the async operation rather than it being synchronous as it sits here. 
          if (ids) bulkDeleteApplication({ ids: ids });
          setApplicationsToDelete([]);
```

ibolton336 (changes requested):
Small change requested for old forEach / array.push style we are stylistically moving away from.

client/src/app/pages/applications/applications-table-analyze/applications-table-analyze.tsx
gildub: @ibolton336, that's a good opportunity indeed.

gildub: @ibolton336, good idea to move setApplicationsToDelete to the success/error callbacks.
We don't really need to handle loading or error state when deleting items, the notification whether the delete operation is successful or fails, by providing information to the user.

ibolton336 (approved):

PR #1239, opened 2023-07-31
:ghost: Target api updates

ibolton336:
UI changes relating to https://github.com/konveyor/tackle2-hub/pull/464

- Rename setting for ruleset/target order to match api changes
- Change TS model for Ruleset to match api changes. Ruleset is now a field on the target model and can be appended to directly / read from directly when performing CRUD operations against a custom target.
- Drive target card selection from label rather than ruleset/target since we no longer need to store the entire ruleset/target state in the wizard. This is because the analyzer no longer requires it to be passed in the TaskData. We are now only passing labels. This simplifies the wizard state & allows tracking of individual ruleset selection/ source selection/ target selection to be removed. 
-  Now we can drive the select targets / select sources dropdowns on the set-options analysis wizard form from labels rather than tracking specific state for either of those. 
- Changes format of default sources/targets to match new hub api label format


client/src/app/api/rest.ts
```diff
@@ -463,7 +480,7 @@ export const createFile = ({
   file: IReadFile;
 }) =>
   axios
-    .post<Ruleset>(`${FILES}/${file.fileName}`, formData, fileHeaders)
```
mturley, line 466: Why remove the type param here? Is there a suitable replacement or is inference enough here?

client/src/app/pages/migration-targets/components/custom-target-form.tsx
```diff
@@ -485,7 +552,7 @@ export const CustomTargetForm: React.FC<CustomTargetFormProps> = ({
                     file={file.fullFile}
                     key={file.fileName}
                     customFileHandler={(file) => {
-                      if (file.type === "placeholder") {
+                      if (file.type === "place as Blobholder") {
```
mturley, line 555: Was this intentional?

mturley (commented):
A couple tiny things, but mostly LGTM. There is probably some amount of preexisting repetition around this code that we can reduce in the future but that's out of scope here.

client/src/app/pages/migration-targets/components/custom-target-form.tsx
ibolton336: Good catch - thanks.

client/src/app/api/rest.ts
ibolton336: Added a new type for HubFile here. Fixed.

client/src/app/common/CustomRules/rules-utils.tsx
```diff
@@ -137,8 +144,10 @@ export const getLabels = (labels: string[]) =>
...
 
-export const getRulesetTargetList = (ruleset: Ruleset) => {
-  return ruleset.rules.reduce((acc: string[], rule) => {
-    return [...acc, rule?.metadata?.target || ""];
-  }, []);
+export const getAllLabelsFromTarget = (target: Target) => {
+  return (
+    target?.labels?.reduce((acc: TargetLabel[], label) => {
+      return [...acc, label || ""];
+    }, []) || []
```
sjd78, line 151: This doesn't seem right.  `label` would be a `TagetLabel` so you're iterating to rebuild the source array?

Wouldn't it be that same to just do:
```js
    target?.labels ?? []
```

Further, it appears that this function is only used in `set-options.tsx` and only the `.label` field of the `TargetLabel` seems to be used.  Does the full object need to be there?  If it _is_ sufficient to just use `TargetLabel.label`, the code in both locations could be reduced, and this function could just be a simple pluck (and fancy with destructuring):
```js
    target?.labels?.map(({ label }) => label) ?? []
```

client/src/app/data/targets.ts
```diff
@@ -1,33 +1,51 @@
...
-  "konveyor.io/target=linux",
-  "konveyor.io/target=openjdk",
-  "konveyor.io/target=openjdk11",
-  "konveyor.io/target=openjdk17",
-  "konveyor.io/target=openliberty",
-  "konveyor.io/target=quarkus",
-  "konveyor.io/target=resteasy",
-  "konveyor.io/target=rhr",
-  "konveyor.io/target=azure-appservice",
+const openTargets: TargetLabel[] = [
```
sjd78, line 4: Is this a structure that needs to be kept in sync with the backend?  Any change it would be available via api instead?

client/src/app/pages/applications/analysis-wizard/set-options.tsx
```diff
@@ -21,39 +21,53 @@ import spacing from "@patternfly/react-styles/css/utilities/Spacing/spacing";
...
   const [isSelectTargetsOpen, setSelectTargetsOpen] = React.useState(false);
   const [isSelectSourcesOpen, setSelectSourcesOpen] = React.useState(false);
-  const { rulesets } = useFetchRulesets();
+  const { targets } = useFetchTargets();
 
-  const allRulesetTargets = rulesets
-    .map((Ruleset) => getRulesetTargetList(Ruleset))
-    .flat();
+  const allLabelsFromTargets = targets
+    .map((target) => getAllLabelsFromTarget(target))
```
sjd78, line 45: If the attribute pluck is all you need, the helper function isn't really even needed and this could be written:
```js
    .map(({ labels }) => labels)
```

sjd78 (commented):
I didn't get through everything so far, but here are a few minor things I noticed.

client/src/app/common/CustomRules/rules-utils.tsx
ibolton336: Good catch. This function is a leftover from the previous implementation. Will update.

client/src/app/data/targets.ts
ibolton336: These are old rules leftover from windup that we were asked to keep around. Maybe worth a follow up issue to move these out of our UI codebase?

client/src/app/pages/applications/analysis-wizard/set-options.tsx
ibolton336: +1 Will clean this up. No need to have this utility function around any more.

mturley (approved):

sjd78 (approved):
LGTM

PR #1240, opened 2023-08-01
:sparkles: Proxies: rest, queries and notifications

gildub:
In continuation of work already done to align, feature by feature, here the proxy configuration to :
- Have it's REST function to use axios instead of APIConfig and uUse React Query consitenlty
- Use Notification` instead for `Alert`.


client/src/app/queries/proxies.ts
```diff
@@ -1,51 +1,34 @@
...
     proxies: data || [],
     isFetching: isLoading,
     fetchError: error as AxiosError,
   };
 };
 
-export const useUpdateProxyMutation = (onSuccess: any) => {
-  const [putResult, setPutResult] = useState<any>(null);
+export const useUpdateProxyMutation = (
+  onSuccess: (res: any) => void,
```
ibolton336, line 22: Let's add a type here for the response if it is needed.

client/src/app/api/rest.ts
```diff
@@ -309,22 +309,6 @@ export const deleteIdentity = (identity: Identity): AxiosPromise => {
   return APIClient.delete(`${IDENTITIES}/${identity.id}`);
 };
 
-export const getProxies = (): AxiosPromise<Array<Proxy>> => {
-  return APIClient.get(`${PROXIES}`, jsonHeaders);
```
ibolton336, line 313: Nice!

ibolton336 (changes requested):
One small comment. Looking good though!

ibolton336 (approved):

PR #1246, opened 2023-08-03
:sparkles: ApplicationAssessmentPageHeader refactor to remove useEntityModal

gildub:
- Replace UseEntityModal with local state "standard" approach we now use
- Remove useEntityModal hook which is not used anymore.

Resolves https://github.com/konveyor/tackle2-ui/issues/1234


ibolton336:
LGTM

PR #1247, opened 2023-08-04
:seedling: Upgrade jest, and related, to version 29

sjd78:
- Upgrading to current `jest`, `@testing-library`, and `ts-jest` libraries brings them current and will allow upgrading typescript to version 5.

  - The jest snapshot format changed between version 29 and 29.  All of the snapshots are brought current in this commit.


gildub (approved):
LGTM

PR #1259, opened 2023-08-07
:bug: Fix application update

ibolton336:
- This is a regression that is causing the app to be overwritten on update


jortel (approved):
VISACK

PR #1267, opened 2023-08-08
:bug: Reset wizard form after analysis run

### LINKED ISSUE #1256: Sometimes wizard state is persisted between analysis runs
I think that we need to refactor the analysis wizard to use our new modal paradigm. Seems that there is an intermittent bug involving multiple modals with separate state that isn't being cleaned up.

ibolton336:
- Resolves #1256 
- When closing the wizard, the taskgroup state hangs around between analysis runs causing problems. This PR cleans up the react state when closing/submitting the analysis wizard

### DIFF
--- a/client/src/app/pages/applications/analysis-wizard/analysis-wizard.tsx
+++ b/client/src/app/pages/applications/analysis-wizard/analysis-wizard.tsx
@@ -102,12 +102,6 @@ export const AnalysisWizard: React.FC<IAnalysisWizard> = ({
   const [stepIdReached, setStepIdReached] = React.useState(1);
   const isMutating = useIsMutating();
 
-  React.useEffect(() => {
-    if (!currentTaskgroup) {
-      createTaskgroup(defaultTaskgroup);
-    }
-  }, []);
-
   const onCreateTaskgroupSuccess = (data: Taskgroup) => {
     setCurrentTaskgroup(data);
   };
@@ -287,12 +281,22 @@ export const AnalysisWizard: React.FC<IAnalysisWizard> = ({
 
   const isModeValid = applications.every((app) => isModeSupported(app, mode));
 
+  const handleCancel = () => {
+    if (currentTaskgroup && currentTaskgroup.id) {
+      deleteTaskgroup(currentTaskgroup.id);
+    }
+    setCurrentTaskgroup(null);
+    reset();
+    onClose();
+  };
+
   const onSubmit = (fieldValues: AnalysisWizardFormValues) => {
     if (currentTaskgroup) {
       const taskgroup = setupTaskgroup(currentTaskgroup, fieldValues);
-
       submitTaskgroup(taskgroup);
     }
+    setCurrentTaskgroup(null);
+    reset();
     onClose();
   };
 
@@ -301,13 +305,11 @@ export const AnalysisWizard: React.FC<IAnalysisWizard> = ({
     { prevId, prevName }
   ) => {
     if (id && stepIdReached < (id as number)) setStepIdReached(id as number);
-  };
-
-  const handleClose = () => {
-    reset();
-    if (currentTaskgroup && currentTaskgroup.id)
-      deleteTaskgroup(currentTaskgroup.id);
-    onClose();
+    if (id === StepId.SetTargets) {
+      if (!currentTaskgroup) {
+        createTaskgroup(defaultTaskgroup);
+      }
+    }
   };
 
   const analyzableApplications = useAnalyzableApplications(applications, mode);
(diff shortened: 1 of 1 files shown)


gitdallas (approved):
although would it make sense to call `handleClose` here? does both `reset()` and `onClose()`, i'm unsure about the delete task group stuff though.

ibolton336:
> although would it make sense to call `handleClose` here? does both `reset()` and `onClose()`, i'm unsure about the delete task group stuff though.

You're right on the money. Our delete task group stuff was very wrong and needs to be refactored.

ibolton336:
@gitdallas Updated - good idea to consolidate that logic.

client/src/app/pages/applications/analysis-wizard/analysis-wizard.tsx
```diff
@@ -286,28 +280,37 @@ export const AnalysisWizard: React.FC<IAnalysisWizard> = ({
...
-    onClose();
+    resetWizard();
   };
 
   const onMove: WizardStepFunctionType = (
     { id, name },
     { prevId, prevName }
   ) => {
     if (id && stepIdReached < (id as number)) setStepIdReached(id as number);
+    if (id === 2) {
```
gitdallas, line 302: What is 2? Any way we can use a step property or something instead?

Line 301 indicates that id is a string (otherwise why `as number`?), but line 302 is treating it as a number.

ibolton336: Good point - we used to have an enum for this. I will create another one.

ibolton336: Screenshot 2023-08-08 at 2 54 07 PM
Looks like the string/number type is coming from PF.

client/src/app/pages/applications/applications-table-analyze/applications-table-analyze.tsx
```diff
@@ -84,7 +84,10 @@ export const ApplicationsTableAnalyze: React.FC = () => {
   const { t } = useTranslation();
   const { pushNotification } = React.useContext(NotificationsContext);
 
-  const [isAnalyzeModalOpen, setAnalyzeModalOpen] = React.useState(false);
+  const [createUpdateModalState, setCreateUpdateModalState] = React.useState<
+    Application | Application[] | null
+  >(null);
```
gildub, line 89: The default value should be empty array `[]` and the null type is not needed.

client/src/app/pages/applications/applications-table-analyze/applications-table-analyze.tsx
```diff
@@ -84,7 +84,10 @@ export const ApplicationsTableAnalyze: React.FC = () => {
...
+  const isWizardOpen = createUpdateModalState !== null;
```
gildub, line 90:
```suggestion
  const isWizardOpen = createUpdateModalState.length > 0;
```

client/src/app/pages/applications/applications-table-analyze/applications-table-analyze.tsx
```diff
@@ -613,9 +616,9 @@ export const ApplicationsTableAnalyze: React.FC = () => {
       </Modal>{" "}
       <AnalysisWizard
         applications={selectedRows}
-        isOpen={isAnalyzeModalOpen}
+        isOpen={isWizardOpen}
         onClose={() => {
-          setAnalyzeModalOpen(false);
+          setCreateUpdateModalState(null);
```
gildub, line 621:
```suggestion
          setCreateUpdateModalState([]);
```

client/src/app/pages/applications/analysis-wizard/analysis-wizard.tsx
```diff
@@ -286,28 +280,37 @@ export const AnalysisWizard: React.FC<IAnalysisWizard> = ({
   };
 
   const isModeValid = applications.every((app) => isModeSupported(app, mode));
+  const resetWizard = () => {
+    setCurrentTaskgroup(null);
+    reset();
+    onClose();
```
gildub, line 286: Why close the wizard here ?

gildub (commented):
Please see inline.

client/src/app/pages/applications/analysis-wizard/analysis-wizard.tsx
```diff
@@ -286,28 +280,37 @@ export const AnalysisWizard: React.FC<IAnalysisWizard> = ({
   };
 
   const isModeValid = applications.every((app) => isModeSupported(app, mode));
+  const resetWizard = () => {
```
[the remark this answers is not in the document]
gitdallas: @ibolton336  - actually the `delete` is more of a cancel right? maybe just `handleCancel` which could delete and then call the close function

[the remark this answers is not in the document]
ibolton336: Updated.

gitdallas (approved):

Issue #1260, opened 2023-08-08
Implement the Admin view Assessment nav item & page


ibolton336:
This issue intends to capture: 
- UI routing
- Page component skeleton for new Admin assessment view
- Placeholder table for questionnaires 
![Image](https://github.com/konveyor/tackle2-ui/assets/11218376/05ef4efa-01ea-435e-80f9-22c0bfef9649)

PR #1268, opened 2023-08-09
:sparkles: Initial Assessment page

gildub:
The hub is going to provide Questionnaire type so the legacy Questionnaire (in model.ts)  is renamed PathfinderQuestionnaire which is likely to become obsolete.

This uses PF5 Dropdown for Kebab menu instead of deprecated Dropdown.

image

Resolves https://github.com/konveyor/tackle2-ui/issues/1260

### DIFF
--- a/client/src/app/api/models.ts
+++ b/client/src/app/api/models.ts
@@ -218,10 +218,10 @@ export interface Assessment {
   status: AssessmentStatus;
   stakeholders?: number[];
   stakeholderGroups?: number[];
-  questionnaire: Questionnaire;
+  questionnaire: PathfinderQuestionnaire;
 }
 
-export interface Questionnaire {
+export interface PathfinderQuestionnaire {
   categories: QuestionnaireCategory[];
 }
 
@@ -704,3 +704,13 @@ export type HubFile = {
   name: string;
   path: string;
 };
+
+export interface Questionnaire {
+  id: number;
+  required: boolean;
+  name: string;
+  questions: number;
+  rating: string;
+  dateImported: string;
+  system: boolean;
+}
--- a/client/src/app/layout/SidebarApp/SidebarApp.tsx
+++ b/client/src/app/layout/SidebarApp/SidebarApp.tsx
@@ -228,10 +228,7 @@ export const SidebarApp: React.FC = () => {
                 </NavExpandable>
               ) : null}
               <NavItem>
-                <NavLink
-                  to={Paths.questionnaires}
-                  activeClassName="pf-m-current"
-                >
+                <NavLink to={Paths.assessment} activeClassName="pf-m-current">
                   Assessment
                 </NavLink>
               </NavItem>
--- a/client/src/app/layout/SidebarApp/SidebarApp.tsx
+++ b/client/src/app/layout/SidebarApp/SidebarApp.tsx
@@ -226,7 +226,15 @@ export const SidebarApp: React.FC = () => {
                     </NavLink>
                   </NavItem>
                 </NavExpandable>
-              ) : null}
+              ) : null}{" "}
+              <NavItem>
+                <NavLink
+                  to={Paths.questionnaires}
+                  activeClassName="pf-m-current"
+                >
+                  Assessment
+                </NavLink>
+              </NavItem>
             </NavList>
           )}
         </Nav>
--- a/client/src/app/layout/SidebarApp/SidebarApp.tsx
+++ b/client/src/app/layout/SidebarApp/SidebarApp.tsx
@@ -226,7 +226,7 @@ export const SidebarApp: React.FC = () => {
                     </NavLink>
                   </NavItem>
                 </NavExpandable>
-              ) : null}{" "}
+              ) : null}
               <NavItem>
                 <NavLink
                   to={Paths.questionnaires}
--- a/client/src/app/pages/questionnaires/Questionnaires.tsx
+++ b/client/src/app/pages/questionnaires/Questionnaires.tsx
@@ -29,6 +29,7 @@ import CubesIcon from "@patternfly/react-icons/dist/esm/icons/cubes-icon";
 import EllipsisVIcon from "@patternfly/react-icons/dist/esm/icons/ellipsis-v-icon";
 
 import {
+  useDeleteQuestionnaireMutation,
   useFetchQuestionnaires,
   useUpdateQuestionnaireMutation,
 } from "@app/queries/questionnaires";
@@ -100,6 +101,21 @@ export const Questionnaires: React.FC = () => {
 
   const { mutationFn: updateQuestionnaire } = useUpdateQuestionnaireMutation(
     onSaveQuestionnaireSuccess,
+    onSaveQuestionnaireError
+  );
+
+  const onDeleteQuestionnaireSuccess = (name: string) => {
+    pushNotification({
+      title: t("toastr.success.deletedWhat", {
+        what: name,
+        type: t("terms.questionnaire"),
+      }),
+      variant: "success",
+    });
+  };
+
+  const { mutationFn: deleteQuestionnaire } = useDeleteQuestionnaireMutation(
+    onDeleteQuestionnaireSuccess,
     onSaveQuestionnaireError
   );
 
@@ -290,6 +306,7 @@ export const Questionnaires: React.FC = () => {
                                     ...questionnaire,
                                     required: !questionnaire.required,
                                   },
+                                  // TODO Remove mock when Hub API is ready
                                   mockQuestionnaires,
                                   setMockQuestionnaires
                                 );
@@ -448,8 +465,12 @@ export const Questionnaires: React.FC = () => {
         onClose={() => setQuestionnaireToDelete(null)}
         onConfirm={() => {
           if (questionnaireToDelete) {
-            // TODO Add API Query
-            // deleteQuestionnaire(questionnaireToDelete);
+            deleteQuestionnaire(
+              questionnaireToDelete,
+              // TODO Remove mock when Hub API is ready
+              mockQuestionnaires,
+              setMockQuestionnaires
+            );
             setQuestionnaireToDelete(null);
           }
         }}
--- a/client/src/app/pages/questionnaires/Questionnaires.tsx
+++ b/client/src/app/pages/questionnaires/Questionnaires.tsx
@@ -173,7 +173,6 @@ export const Questionnaires: React.FC = () => {
     selectionState: { selectedItems },
     propHelpers: {
       toolbarProps,
-      toolbarBulkSelectorProps,
       filterToolbarProps,
       paginationToolbarItemProps,
       paginationProps,
@@ -202,7 +201,6 @@ export const Questionnaires: React.FC = () => {
           >
             <Toolbar {...toolbarProps}>
               <ToolbarContent>
-                <ToolbarBulkSelector {...toolbarBulkSelectorProps} />
                 <FilterToolbar {...filterToolbarProps} />
                 <ToolbarGroup variant="button-group">
                   {/* <RBAC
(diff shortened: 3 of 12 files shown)


client/src/app/layout/SidebarApp/SidebarApp.tsx
```diff
@@ -226,7 +226,15 @@ export const SidebarApp: React.FC = () => {
                     </NavLink>
                   </NavItem>
                 </NavExpandable>
-              ) : null}
+              ) : null}{" "}
```
ibolton336, line 229: Looking great! A few small things but this is great progress. Is this space needed?

client/src/app/pages/questionnaires/questionnaires.tsx
```diff
@@ -0,0 +1,459 @@
+import * as React from "react";
```
ibolton336, line 1: So all components are lower case file names now?

client/src/app/pages/questionnaires/questionnaires.tsx
```diff
@@ -0,0 +1,459 @@
...
+                            {...getTdProps({ columnKey: "required" })}
+                          >
+                            {questionnaire.required}
+                            <Switch
+                              id={`required-switch-${rowIndex.toString()}`}
+                              label="Yes"
+                              labelOff="No"
+                              isChecked={questionnaire.required}
+                              onChange={(_event, _checked) => {
+                                updateQuestionnaire(
```
ibolton336, line 288: Are we hitting the api on every change? I think currently we dont submit the questionnaire until the end of the assessment.

client/src/app/pages/questionnaires/questionnaires.tsx
```diff
@@ -0,0 +1,459 @@
...
+                            width={10}
+                            {...getTdProps({ columnKey: "required" })}
+                          >
+                            {questionnaire.required}
+                            <Switch
+                              id={`required-switch-${rowIndex.toString()}`}
+                              label="Yes"
+                              labelOff="No"
+                              isChecked={questionnaire.required}
+                              onChange={(_event, _checked) => {
```
ibolton336, line 287: I have seen these params defined several different ways. I personally like using the underscore as a single char for params we aren't going to be using.

ibolton336 (changes requested):

client/src/app/pages/questionnaires/questionnaires.tsx
gildub: You're right since it's a TSX file I suppose it should be Questionnaires.tsx.

client/src/app/pages/questionnaires/questionnaires.tsx
gildub: This is not assessment it's questionnaire page, although the menu is dubbed Assessment.
So when the user is switching the `Required` field we need to update the backend.
I'm not sure yet how the final Questionnaire data is going to look like, for now I assume it's part of the Questionnaire data.

client/src/app/pages/questionnaires/questionnaires.tsx
gildub: I agree with you that looks better. 
Meanwhile _ can be used only once otherwise there is a conflict.

client/src/app/queries/questionnaires.ts
```diff
@@ -0,0 +1,108 @@
...
+  setMockQuestionnaires: (questionnaires: Questionnaire[]) => void
+) => {
+  const newMockQuestionnaires = mockQuestionnaires.map((questionnaire) =>
+    questionnaire.id === obj.id ? obj : questionnaire
+  );
+  setMockQuestionnaires(newMockQuestionnaires);
+};
+
+export const useUpdateQuestionnaireMutation = (
+  _onSuccess: () => void,
```
ibolton336, line 46: whats going on here?

client/src/app/queries/questionnaires.ts
```diff
@@ -0,0 +1,108 @@
...
+  setMockQuestionnaires: (questionnaires: Questionnaire[]) => void
+) => {
+  const newMockQuestionnaires = mockQuestionnaires.filter(
+    (questionnaire) => questionnaire.id !== id
+  );
+  setMockQuestionnaires(newMockQuestionnaires);
+};
+
+export const useDeleteQuestionnaireMutation = (
+  _onSuccess: (name: string) => void,
```
ibolton336, line 82: Why add the underscores here?

client/src/app/pages/questionnaires/questionnaires.tsx
ibolton336: Oh ok makes sense.

client/src/app/queries/questionnaires.ts
gildub: This is mock stuff

client/src/app/queries/questionnaires.ts
gildub: This is mock function

client/src/app/pages/questionnaires/questionnaires.tsx
ibolton336: So we aren't using either param?

ibolton336: If we aren't using the params, I think you can just not pass them to the callback function here.

client/src/app/queries/questionnaires.ts
ibolton336: Should probably name is accordingly, then. Looks like a real mutation function we would expect to use. 

Are underscores standard for mock function param names?

client/src/app/api/models.ts
```diff
@@ -704,3 +704,13 @@ export type HubFile = {
   name: string;
   path: string;
 };
+
+export interface Questionnaire {
```
mturley, line 708: Should we name this `CustomQuestionnaire` to distinguish it from `PathfinderQuestionnaire` rather than imply inheritance between the two? or is it intended that all questionnaires will eventually use this type and pathfinder objects will be deprecated?

client/src/app/layout/SidebarApp/SidebarApp.tsx
```diff
@@ -227,6 +227,14 @@ export const SidebarApp: React.FC = () => {
                   </NavItem>
                 </NavExpandable>
               ) : null}
+              <NavItem>
+                <NavLink
+                  to={Paths.questionnaires}
+                  activeClassName="pf-m-current"
+                >
+                  Assessment
```
mturley, line 235: I suppose this is design feedback for Justin but it seems confusing to me that we'll now have two views called Assessment - one here and one as a tab in application inventory. I guess because this is under admin it makes sense, but I wonder if we should propose that the nav item and page title are called "Questionnaires" instead, and a subtitle in the page header could be used to explain that they are for assessment.

client/src/app/pages/questionnaires/Questionnaires.tsx
```diff
@@ -0,0 +1,480 @@
...
+      dateImported: assessment.dateImported || "",
+    }),
+    initialSort: { columnKey: "name", direction: "asc" },
+    hasPagination: true,
+    isLoading: isFetching,
+  });
+  const {
+    currentPageItems,
+    numRenderedColumns,
+    selectionState: { selectedItems },
```
mturley, line 173: You can remove this line that pulls out `selectedItems` since we're not using selection state on this page.

client/src/app/pages/questionnaires/Questionnaires.tsx
```diff
@@ -0,0 +1,480 @@
...
+            style={{
+              backgroundColor: "var(--pf-v5-global--BackgroundColor--100)",
+            }}
+          >
+            <Toolbar {...toolbarProps}>
+              <ToolbarContent>
+                <ToolbarBulkSelector {...toolbarBulkSelectorProps} />
+                <FilterToolbar {...filterToolbarProps} />
+                <ToolbarGroup variant="button-group">
+                  {/* <RBAC
```
mturley, line 206: Side-note: are we going to need to revisit RBAC on all these new views before 7.0 feature freeze? I think we have some commented-out or omitted RBAC stuff on Issues/Dependencies etc as well. Not sure if we're tracking that anywhere

client/src/app/pages/questionnaires/Questionnaires.tsx
```diff
@@ -0,0 +1,480 @@
...
+                <ToolbarItem {...paginationToolbarItemProps}>
+                  <SimplePagination
+                    idPrefix="questionnaire-table"
+                    isTop
+                    paginationProps={paginationProps}
+                  />
+                </ToolbarItem>
+              </ToolbarContent>
+            </Toolbar>
+            <Table {...tableProps} aria-label="Assement table">
```
mturley, line 253: Typo here
```suggestion
            <Table {...tableProps} aria-label="Assesment table">
```
also it seems weird to have the labels and ids/filename inconsistent between the Assessment and Questionnaires titles. Probably fine until we run it by Justin.

client/src/app/pages/questionnaires/Questionnaires.tsx
```diff
@@ -0,0 +1,478 @@
...
+        id="import.modal"
+        title={t("dialog.title.import", {
+          what: t("terms.questionnaire").toLowerCase(),
+        })}
+        variant={ModalVariant.medium}
+        isOpen={isImportModal}
+        onClose={() => setIsImportModal(false)}
+      >
+        <Text>TODO Import questionnaire component</Text>
+      </Modal>{" "}
```
mturley, line 418: The `{" "}`s here and on line 440 can probably be removed since Modal doesn't render anything inline, so there's no need for space characters between them

client/src/app/queries/questionnaires.ts
mturley: +1 if the underscores are there just to remind us this is mock stuff, we may want to instead use the real names and add a TODO comment instead.

mturley (commented):
Looking good, a couple small things.

client/src/app/layout/SidebarApp/SidebarApp.tsx
gildub: @mturley,
My understanding is that the Administrator -> Assessment page is actually displaying the questionnaires but the Required button activate those questionnaires to the default Assessment process.
So the Questionnaires.tsx page might actually not the correct name. I wonder if I should rename it now or wait.
Because this page handles Assessment settings.

client/src/app/pages/questionnaires/Questionnaires.tsx
gildub: @mturley, I don't know.
@ibolton336, what do you think ?

client/src/app/pages/questionnaires/Questionnaires.tsx
gildub: @mturley, let me fix the typo in the meantime ;)

client/src/app/api/models.ts
mturley: if pathfinder will be removed eventually and we'll only have one questionnaire type, i'm fine with leaving the name as-is.

client/src/app/layout/SidebarApp/SidebarApp.tsx
mturley: yeah maybe Assessment Settings is a better name than Assessment then. I just find the current name potentially confusing to users.

mturley (approved):

client/src/app/layout/SidebarApp/SidebarApp.tsx
gildub: @ibolton336, what do you think about renaming that page Assessment.tsx ?
Because the used case is about providing Administrator a way to control assessment, which starts with activating (Required column) questionnaires.

ibolton336: I still find that to be a naming conflict. I'd rather it be something like AssessmentTools/QuestionaireSettings/AssessmentSettings etc.

mturley (approved):

ibolton336 (approved):

PR #1269, opened 2023-08-09
:ghost: Use nullish coalescing instead of awkward ternary in getActiveRowDerivedState

mturley:
Just a small change to incorporate feedback from @sjd78 after merge on #1229 before I forget about it.


gildub (approved):
LGTM

PR #1270, opened 2023-08-09
:seedling: Require node/npm versions as used in ubi9/nodejs-18

sjd78:
Versions pulled from `registry.access.redhat.com/ubi9/nodejs-18:1-59.1690899127`
  - node: 18.14.2
  - npm: 9.5.0

If the node or npm versions don't match the versions in `package.json`, any `npm install` will flag a warning.

The `.npmrc` file setting `engine-strict=true` forces the install version mismatch warnings to become errors.

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,0 +1,1 @@
+engine-strict=true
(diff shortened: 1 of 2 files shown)


sjd78:
Nothing magical with this one.  Just a tweak to help ensure we're all on at least the same nodejs version.

PR #1273, opened 2023-08-09
:bug: Add a fallback for issue title

ibolton336:
- In the event of a blank ruleReport.description, fallback to use the name or a hardcoded string to prevent an app crash

### DIFF
--- a/client/src/app/pages/issues/helpers.ts
+++ b/client/src/app/pages/issues/helpers.ts
@@ -1,5 +1,9 @@
 import { Location, LocationDescriptor } from "history";
-import { AnalysisIssueReport, AnalysisRuleReport } from "@app/api/models";
+import {
+  AnalysisIssue,
+  AnalysisIssueReport,
+  AnalysisRuleReport,
+} from "@app/api/models";
 import {
   FilterCategory,
   FilterType,
@@ -113,11 +117,12 @@ export const getAffectedAppsUrl = ({
     .replace("/:rule/", `/${encodeURIComponent(ruleReport.rule)}/`);
   const prefix = (key: string) =>
     `${TableURLParamKeyPrefix.issuesAffectedApps}:${key}`;
+
   return `${baseUrl}?${trimAndStringifyUrlParams({
     newPrefixedSerializedParams: {
       [prefix("filters")]: serializeFilterUrlParams(toFilterValues).filters,
       [FROM_ISSUES_PARAMS_KEY]: fromIssuesParams,
-      issueTitle: ruleReport.description.split("\n")[0],
+      issueTitle: getIssueTitle(ruleReport),
     },
   })}`;
 };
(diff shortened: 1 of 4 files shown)


client/src/app/pages/issues/issue-detail-drawer/file-incidents-detail-modal/file-incidents-detail-modal.tsx
```diff
@@ -58,7 +58,10 @@ export const FileIncidentsDetailModal: React.FC<
     isFetching ||
     (firstFiveIncidents.length > 0 && activeTabIncidentId === undefined);
 
-  const issueTitle = issue.description.split("\n")[0];
+  const issueTitle =
+    issue.description.split("\n")[0] ||
+    issue.name.split("\n")[0] ||
+    "Issue name";
```
jortel, line 64: Seems like something other than "issue name" would be better.
perhaps:
- unnamed
- anonymous
- not-named

client/src/app/pages/issues/helpers.ts
```diff
@@ -195,3 +200,10 @@ export const parseReportLabels = (
   });
   return { sources, targets, otherLabels };
 };
+
+export const getIssueTitle = (
+  issueReport: AnalysisRuleReport | AnalysisIssue | AnalysisRuleReport
+) =>
+  issueReport?.description?.split("\n")[0] ||
+  issueReport?.name?.split("\n")[0] ||
+  "*Unnamed*";
```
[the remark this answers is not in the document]
ibolton336: Seems that we would still need the optional chaining here to prevent an error if we try to call the split method on something that is not found.

client/src/app/pages/issues/helpers.ts
```diff
@@ -195,3 +200,10 @@ export const parseReportLabels = (
   });
   return { sources, targets, otherLabels };
 };
+
+export const getIssueTitle = (
+  issueReport: AnalysisRuleReport | AnalysisIssue | AnalysisIssueReport
+) =>
+  issueReport?.description?.split("\n")[0] ??
+  issueReport?.name?.split("\n")[0] ??
+  "*Unnamed*";
```
sjd78, line 209: Is empty string a valid issue title?

One of the main differences between `||` and `??` is how empty string (and 0) is handled.  For `??` they're considered valid values, and for `||` they're considered `false`.  Here with `??`, if the `description` is empty string, the title will be empty string.

`"".split("\n")[0] ?? "foo"` is `""`

`"".split("\n")[0] || "foo"` is `"foo"`

ibolton336: Good to know! I think I will revert back to `||`. We don't wan't an empty string title.

gildub (approved):
LGTM

PR #1280, opened 2023-08-10
:bug: Add taskgroup when uploading binary

ibolton336:
- Improve handling of taskgroup state by using contextAPI
- Async create the taskgroup before uploading the custom binary in the single app use case. This is currently broken because we require a task group to exist before navigating to the upload file screen. This regression was caused by the move away from useEffect here https://github.com/konveyor/tackle2-ui/pull/1267
Resolves https://issues.redhat.com/browse/MTA-1175

### DIFF
--- a/client/src/app/pages/applications/analysis-wizard/custom-rules.tsx
+++ b/client/src/app/pages/applications/analysis-wizard/custom-rules.tsx
@@ -54,11 +54,11 @@ import { OptionWithValue, SimpleSelect } from "@app/components/SimpleSelect";
 import { toOptionLike } from "@app/utils/model-utils";
 import { useFetchIdentities } from "@app/queries/identities";
 import useRuleFiles from "@app/hooks/useRuleFiles";
-interface CustomRulesProps {
-  taskgroupID: number | null;
-}
-export const CustomRules: React.FC<CustomRulesProps> = (props) => {
+import { useTaskGroup } from "./components/TaskGroupContext";
+
+export const CustomRules: React.FC = () => {
   const { t } = useTranslation();
+  const { taskGroup, updateTaskGroup } = useTaskGroup();
 
   const { watch, setValue, control, getValues } =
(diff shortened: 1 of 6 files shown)


client/src/app/pages/applications/analysis-wizard/custom-rules.tsx
```diff
@@ -92,7 +92,7 @@ export const CustomRules: React.FC<CustomRulesProps> = (props) => {
     successfullyReadFileCount,
     handleFile,
     removeFiles,
-  } = useRuleFiles(props?.taskgroupID, values.customRulesFiles);
+  } = useRuleFiles(taskGroup?.id, values.customRulesFiles);
```
sjd78, line 95: This is the only bit that looks suspicious to me under brain compile.

What happens if the taskGroup changes?   Maybe something like from `undefined` to `{ id: 5 }`?

I'll run through some smoke tests in a bit.  Maybe that will clear up my confusion.

ibolton336: +1 Fair concern. The taskgroup should always exist by the time the user arrives at the custom-rules step as it is created either 1) when [onMove](https://github.com/konveyor/tackle2-ui/blob/435b73afc0d3f4500805c618756d49acf445c29b/client/src/app/pages/applications/analysis-wizard/analysis-wizard.tsx#L310) to step 2 of the wizard or 2) user [uploads](https://github.com/konveyor/tackle2-ui/pull/1280/files#diff-defb41a56081c3324873ae6267336b4b17a1cfd59b3f626243cb5565c8bba3c6R119) a binary on step 1 of the wizard.

sjd78: https://react.dev/learn/reusing-logic-with-custom-hooks#passing-reactive-values-between-hooks

Ok so I'm not concerned about that anymore.

sjd78 (approved):
LGTM

PR #1286, opened 2023-08-14
:seedling: Use `MockServiceWorker` to provide mock data for questionnaires

sjd78:
Using MockServiceWorker (`msw`) [^1] allows us to provide stub/mock data endpoints for rest services on hub that don't yet exist.  In future, msw can also be used to provide a completely stubbed out hub so the UI can run statically.

The msw layer added in this change will intercept the network requests and use mock data for the `/hub/questionnaires` endpoints.  By providing normal endpoints in this way, the rest fetch functions and the react-query hooks can be built and tested as per normal.

Once the real hub endpoints are in place, the stub layer can be disabled and any handling code can be adjusted to cover the differences between the mocks and the real service.

Specifics of the implementation:
  - All of the `msw` code is held in `client/src/mocks`

  - A small change in `client/src/index.tsx` will dynamically load and start our `msw` layer if building/running in development mode.

  - The `AssessmentSettings.tsx` page and the `questionnaires.ts` queries have been modified to remove any reference to mock data.

  - The `mockServiceWorker.js` file[^2] is created by the command `npx msw init ./public`, and is recommended to keep in the repo so everyone uses the same version.  This file should be upgraded with any `msw` package upgrade.

  - Development build config `webpack.dev.ts` has been adjusted to provide the file `mockServiceWorker.js`.  The file is __not__ provided in a production build.

  - In the `stub-new-work/questionnaires.ts` handlers, no validation on the content of the json/form data is done.  Hub services may do field validation on json binding.  That is not done in the `msw` layer.

[^1]: https://mswjs.io/docs/
[^2]: https://mswjs.io/docs/getting-started/integrate/browser


ibolton336:
Awesome!

ibolton336 (approved):

PR #1291, opened 2023-08-16
:seedling: Setup eslint with prettier

sjd78:
Replace the old and broken eslint configuration in `client/package.json` with a more robust configuration.  `eslint`, `typescript-eslint` and `prettier` are configured to work together to lint javascript and typescript files.

Changes:
  - One eslint config for all packages: `.eslintrc.cjs`

  - Add explicit config files for prettier: `.prettierrc.mjs`, `.prettierignore`, and `.editorconfig`

  - Prettier v2 and v3 have changes to some styles, including the default for tailing commas.  Our configurations are now set to v2 default "es5" trailing commas.

  - A total of 10 eslint rules have been configurated as warnings instead of errors.  There are a lot of these warnings, and fixing them in this PR would create a very large PR.  Setting as warnings will allow us to fix the problems and set them back to errors in future PRs.  1 PR for each warn to error conversion is what I anticipate.

  - A set of errors have been fixed in the PR.  These errors only hit a few times each and were easy to fix.

  - Use `lint-staged` as pre-commit hook:
    - Previously `pretty-quick` was used to run `prettier@^2` in the client package when committing staged files.  Since `pretty-quick` does not currently play well with `prettier@^3`, and since we'd like to also use the eslint setup, the pre-commit hook has be changed over to use `lint-staged`.

    - Each monorepo package has its own `lint-staged` configuration based on the files in each repo.  Both `eslint` and `prettier` are used as appropriate.  If any staged files cannot be auto-fixed by eslint, the commit should fail.


.eslintrc.cjs
```diff
@@ -0,0 +1,87 @@
...
+
+    // TODO: set to "error" when all resolved, but keep the `argsIgnorePattern`
+    "@typescript-eslint/no-unused-vars": ["warn", { argsIgnorePattern: "^_" }],
+
+    // TODO: each one of these can be removed or set to "error" when they're all resolved
+    "unused-imports/no-unused-imports": ["warn"],
+    "@typescript-eslint/ban-types": "warn",
+    "@typescript-eslint/no-explicit-any": "warn",
+    "react/jsx-key": "warn",
+    "react-hooks/rules-of-hooks": "warn",
```
mturley, line 51: IMO the rules-of-hooks rule should be `"error"` since conditionally-called hooks can cause really strange broken behaviors.

sjd78:
Now includes the commit from #1316

.eslintrc.cjs
mturley: Sounds perfect to me.

mturley (approved):

ibolton336:
Lgtm - thanks for the work on this.

ibolton336 (approved):

gitdallas (approved):

PR #1295, opened 2023-08-17
:sparkles: add autocomplete component

### LINKED ISSUE #1266: Autocomplete form field
Create a new component for integrating the new autocomplete drop down field with react-hook-form

The Archetype form will use this new component for three fields as described in the enhancement: 

- Tags: Dropdown, multi selection. All selected options will be displayed as removable labels.
- Stakeholders: Dropdown, multi selection. All selected options will be displayed as removable labels. (Optional)
- Stakeholder group(s): Dropdown, multi selection. All selected options will be displayed as removable labels. (Optional)

All dropdown fields will have an autocomplete behavior. Tags appearing in the Tags field will be arranged by their corresponding categories:

Image

gitdallas:
closes #1266 

it is looking mostly like the images in the story (made assumptions on the dropdown menu).

may need some tweaks depending on needs, but wanted to get a PR up before i head out for the long weekend.

image

image


client/src/app/components/Autocomplete.tsx
```diff
@@ -0,0 +1,288 @@
...
+
+  /** callback for removing a chip from the chip selections */
+  const deleteChip = (chipToDelete: string) => {
+    const newChips = currentChips.filter(
+      (chip) => !Object.is(chip, chipToDelete)
+    );
+    setCurrentChips(newChips);
+  };
+
+  React.useEffect(() => {
```
avivtur, line 66: Can you please explain why is this useEffect hook needed?
useEffect is meant to be used with side effects, this looks like the things running here can be inserted into the handleInputChange method, as it changes on every change of inputValue as it seem on the dependency array at line 116

client/src/app/components/Autocomplete.tsx
```diff
@@ -0,0 +1,288 @@
...
+}
+
+export const Autocomplete: React.FC<IAutocompleteProps> = ({
+  // TODO: data just for testing purposes, should be removed
+  options = ["Cluster", "Kind", "Label", "Name", "Namespace", "Status"],
+  value,
+  labelColor,
+  selections = [],
+}) => {
+  const [inputValue, setInputValue] = React.useState(value || "");
```
avivtur, line 38: You can also assign the value prop with a default value of "" and remove the `|| ""` at the end

client/src/app/components/Autocomplete.tsx
```diff
@@ -0,0 +1,288 @@
...
+    setCurrentChips(newChips);
+  };
+
+  React.useEffect(() => {
+    /** in the menu only show items that include the text in the input */
+    const filteredMenuItems = options
+      .filter(
+        (item: string) =>
+          !inputValue ||
+          item.toLowerCase().includes(inputValue.toString().toLowerCase())
```
avivtur, line 72: toString() is not needed as inputValue is already a string

client/src/app/components/Autocomplete.tsx
```diff
@@ -0,0 +1,288 @@
...
+
+  /** close the menu when a click occurs outside of the menu or text input group */
+  const handleClick = (event?: MouseEvent) => {
+    console.log("click", event);
+    if (!event) {
+      return;
+    }
+    if (searchInputRef.current?.contains(event.target as HTMLElement)) {
+      setMenuIsOpen(true);
+    } else if (
```
avivtur, line 207: else if can be simple if

avivtur (commented):
Hi there :) 
Great work on this! 
I've added some comments

client/src/app/components/Autocomplete.tsx
```diff
@@ -0,0 +1,288 @@
...
+    switch (event.key) {
+      case "Tab":
+      case "Escape":
+        event.preventDefault();
+        focusTextInput();
+        setMenuIsOpen(false);
+        break;
+      case "Enter":
+      case " ":
+        setTimeout(() => setMenuIsOpen(false), 0);
```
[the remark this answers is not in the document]
gitdallas: to allow onSelect to perform as well. although this isn't needed anymore as i've changed to close on selection (because otherwise the user wouldn't even see the label being added).

client/src/app/components/Autocomplete.tsx
```diff
@@ -0,0 +1,288 @@
...
+        break;
+    }
+  };
+
+  const inputGroup = (
+    <div ref={searchInputRef}>
```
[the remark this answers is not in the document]
gitdallas: for whatever reason, if i do this here it will not open the menu

client/src/app/components/Autocomplete.tsx
gitdallas: side effect of both a change in current chips or a change in the search now

client/src/app/components/Autocomplete.tsx
```diff
@@ -0,0 +1,284 @@
...
+  searchString?: string;
+  searchInputAriaLabel?: string;
+  labelColor?: LabelProps["color"];
+  selections?: Set<string>;
+  menuHeader?: string;
+  noResultsMessage?: string;
+}
+
+export const Autocomplete: React.FC<IAutocompleteProps> = ({
+  // TODO: data just for testing purposes, should be removed
```
ibolton336, line 28: Could you implement this in a field for testing purposes as a part of this PR?

gitdallas: @ibolton336 - it's being used for tags field now

gitdallas:
While testing this out on the tags field.. I'm wondering if this is meant to allow "new" options or if the user should be restricted to the available options.

ibolton336:
> 

Looks really good! Great question. I think in the case of the application form, we are limiting to existing tags.

gitdallas:
I went ahead and made it configurable with a flag prop. Also made it so that if a user doesn't pass in a menu header, it will not have one (more similar to how the business service dropdown looks).

client/src/app/components/Autocomplete.tsx
avivtur: If you want to update a component’s state when some props or state change, You shouldn’t need an Effect. Removing unnecessary Effects will make your code easier to follow, faster to run, and less error-prone.
https://react.dev/learn/you-might-not-need-an-effect

client/src/app/components/Autocomplete.tsx
```diff
@@ -0,0 +1,288 @@
...
+    if (!event) {
+      return;
+    }
+    if (searchInputRef.current?.contains(event.target as HTMLElement)) {
+      setMenuIsOpen(true);
+    } else if (
+      menuRef.current &&
+      !menuRef.current.contains(event.target as HTMLElement) &&
+      searchInputRef.current &&
+      !searchInputRef.current.contains(event.target as HTMLElement)
```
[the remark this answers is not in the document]
avivtur: ?

client/src/app/components/Autocomplete.tsx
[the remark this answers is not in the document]
avivtur: I wonder why that happens, I'll try to checkout this PR locally and see if I can figure out what's happening to the ref here

gitdallas:
Enter by keypress will work. If you are not allowing user to input custom labels, it will add a new label with whatever the user has. If you are not allowing custom labels, it will only work if the text matches (case insensitive) an option or if there is a hint given.

gitdallas:
As for the menu changes, I wish I would have known about this when I was starting the story. It looks like it will require quite a bit of change... the checkboxes and all. I'm also unsure how the categories work - what will the data look like?

ibolton336:
> As for the menu changes, I wish I would have known about this when I was starting the story. It looks like it will require quite a bit of change... the checkboxes and all. I'm also unsure how the categories work - what will the data look like? @ibolton336

Screenshot 2023-08-29 at 1 33 58 PM

This is example of a tagCategory with tags as it exists today in the controls panel of the app.

gitdallas:
@ibolton336 - if you don't mind, i would prefer another story. I'd like to focus more on removing deprecated dependencies.

ibolton336 (approved):

ibolton336:
> @ibolton336 - if you don't mind, i would prefer that being moved to another story. I'd like to focus more on removing deprecated dependencies.

Alright fair enough! Opened https://github.com/konveyor/tackle2-ui/issues/1320 up that can be worked on now. The deprecated stuff can wait for out bug fix cycle after feature freeze. We should prioritize custom assessment feature work until then.

PR #1296, opened 2023-08-21
:bug: Fix tag rank input 

ibolton336:
https://issues.redhat.com/browse/MTA-1195


client/src/app/pages/controls/tags/components/tag-category-form.tsx
```diff
@@ -189,7 +189,12 @@ export const TagCategoryForm: React.FC<TagCategoryFormProps> = ({
             onMinus={() => {
               onChange((value || 0) - 1);
             }}
-            onChange={() => onChange}
+            onChange={(event) => {
+              const target = event?.target as HTMLInputElement;
+              if (target) {
+                onChange(target.valueAsNumber);
+              }
+            }}
```
sjd78, line 197:
```suggestion
            onChange={onChange}
```

Would that work?

sjd78: Yeah ok, that's crazy.

Looking at the next component in the file, your original change looks reasonable.

ibolton336: Haha! I think we avoid this with the TextInput and other components interfacing with react-hook-form by creating a wrapper and using some fancy typescript extension of the passed in onChange handler type. @mturley Could probably explain it better than I.

sjd78: Eventually I will get react-hook-form and the wrapper components around PF fields loaded up in my brain.  Until then it'll just be an adventure.

mturley (approved):

gitdallas (approved):

Issue #1292, opened 2023-08-17
Add lock icon for "system" questionnaires 


ibolton336:
System questionnaires need to have a lock icon as shown in the mocks here:

<img width="1243" alt="Screenshot 2023-08-17 at 10 38 13 AM" src="https://github.com/konveyor/tackle2-ui/assets/11218376/f28572cb-3b88-4fa1-93ea-afa523cc3446">

avivtur:
Hi @ibolton336 
Can I get assigned to this issue?

avivtur:
@ibolton336 Hi
The icon seems to be under the `Name` column but is not aligned with the start of the header, is that on purpose?
![lock-icon-not-aligned](https://github.com/konveyor/tackle2-ui/assets/67270715/a3ecab2b-19e5-450d-9564-40eea5bf3254)

ibolton336:
@avivtur Good question. I would think we want the icon aligned with the header. @justinxhale

PR #1302, opened 2023-08-23
:sparkles: Add lock icon for "system" questionnaires

avivtur:
closes https://github.com/konveyor/tackle2-ui/issues/1292

Adding a lock icon to "system" questionnaire.

Before:
lock-icon-b4

After:
lock-icon

### DIFF
--- a/client/src/app/pages/assessment-management/assessment-settings/assessment-settings-page.tsx
+++ b/client/src/app/pages/assessment-management/assessment-settings/assessment-settings-page.tsx
@@ -25,9 +25,8 @@ import {
 } from "@patternfly/react-core";
 import { AxiosError } from "axios";
 import { Table, Tbody, Td, Th, Thead, Tr } from "@patternfly/react-table";
-import { LockIcon } from "@patternfly/react-icons";
-import CubesIcon from "@patternfly/react-icons/dist/esm/icons/cubes-icon";
-import EllipsisVIcon from "@patternfly/react-icons/dist/esm/icons/ellipsis-v-icon";
+import { LockIcon, EllipsisVIcon, CubesIcon } from "@patternfly/react-icons";
+import spacing from "@patternfly/react-styles/css/utilities/Spacing/spacing";
 
 import {
   useDeleteQuestionnaireMutation,
@@ -276,12 +275,7 @@ const AssessmentSettings: React.FC = () => {
                           </Td>
                           <Td width={25} {...getTdProps({ columnKey: "name" })}>
                             {questionnaire.system && (
-                              <LockIcon
-                                style={{
-                                  marginRight:
-                                    "var(--pf-v5-global--spacer--sm)",
-                                }}
-                              />
+                              <LockIcon className={spacing.mrSm} />
                             )}
                             {questionnaire.name}
                           </Td>
--- a/client/src/app/pages/assessment-management/assessment-settings/assessment-settings-page.tsx
+++ b/client/src/app/pages/assessment-management/assessment-settings/assessment-settings-page.tsx
@@ -25,6 +25,7 @@ import {
 } from "@patternfly/react-core";
 import { AxiosError } from "axios";
 import { Table, Tbody, Td, Th, Thead, Tr } from "@patternfly/react-table";
+import { LockIcon } from "@patternfly/react-icons";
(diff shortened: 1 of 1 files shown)


ibolton336 (approved):

client/src/app/pages/assessment-management/assessment-settings/assessment-settings-page.tsx
```diff
@@ -274,6 +275,14 @@ const AssessmentSettings: React.FC = () => {
                             />
                           </Td>
                           <Td width={25} {...getTdProps({ columnKey: "name" })}>
+                            {questionnaire.system && (
+                              <LockIcon
+                                style={{
+                                  marginRight:
+                                    "var(--pf-v5-global--spacer--sm)",
+                                }}
```
sjd78, line 283: Looks like there is a different way to do inline icons (at least with PF5) that don't involve touching the icon style https://www.patternfly.org/components/icon#inline

client/src/app/pages/assessment-management/assessment-settings/assessment-settings-page.tsx
```diff
@@ -274,6 +275,14 @@ const AssessmentSettings: React.FC = () => {
                             />
                           </Td>
                           <Td width={25} {...getTdProps({ columnKey: "name" })}>
+                            {questionnaire.system && (
```
sjd78, line 278: To think about:

There is a good case to be made to use a ternary instead of logical `&&` to control the falsey output.  See https://kentcdodds.com/blog/use-ternaries-rather-than-and-and-in-jsx

sjd78 (changes requested):
I think the PF5 example inline style would be good to follow.

client/src/app/pages/assessment-management/assessment-settings/assessment-settings-page.tsx
avivtur: I'm a bit confused about you're suggestion, all I meant here is to give a proper spacing from the icon to the text so they won't be too close.
I tried to do as suggested in the example to something like, which didn't gave me the margin I was looking for:
```
<Td width={25} {...getTdProps({ columnKey: "name" })}>
  <TextContent>
    <Text component="p">
      {questionnaire.system && (
        <Icon isInline>
          <LockIcon />
        </Icon>
      )}
      {questionnaire.name}
    </Text>
  </TextContent>
</Td>
```

client/src/app/pages/assessment-management/assessment-settings/assessment-settings-page.tsx
avivtur: I see what you mean, however, I don't agree that it means we should always use ternary operators, because I believe they are less readable. 
We can either do !!questionnaire.system OR Boolean(questionnaire.system) to go around it, only if they var is not already a boolean.
Since in this specific system is already guaranteed to be boolean (as Questionnaire type suggests) we can do:
`questionnaire.system && ...` safely as it will turn out to either: `false && ...` which won't render the icon (it will not render false to the screen) or `true && ...` which will render only the icon to the screen

client/src/app/pages/assessment-management/assessment-settings/assessment-settings-page.tsx
ibolton336: Also, you can use the "spacing" utility class import to handle any custom spacing. 

```suggestion
import spacing from "@patternfly/react-styles/css/utilities/Spacing/spacing";
        className={spacing.mrSm}
```

avivtur: Sounds like what I was looking for, I'll check that thanks :+1:

client/src/app/pages/assessment-management/assessment-settings/assessment-settings-page.tsx
sjd78: More or less readable is a bit subjective.

In this particular case, the logical && does make sense.  And careful with anything in javascript being guaranteed to be any specific kind of type.  JS just doesn't work that way, and TS only can do static type checking.

sjd78 (approved):
The helper class for spacing looks better.

mturley (approved):

Issue #1276, opened 2023-08-10
Delete Questionnaire


ibolton336:
Questionnaire can be deleted by selecting the Delete option from the kebab menu in the questionnaire row. This should open a modal window displaying the following text:

Deleting a questionnaire will cascade into the deletion of all answered questionnaires associated to applications and/or archetypes. This action cannot be undone. Please enter the name of the questionnaire below to confirm:

Below that, a text field your be displayed for users to enter the name of the questionnaire to be deleted. The Delete button in the modal won't be enabled until the user enters the right name.

PR #1303, opened 2023-08-23
:sparkles: Questionnaire delete confirmation

avivtur:
closes https://github.com/konveyor/tackle2-ui/issues/1276

Adding an explanation of what happens if the questionnaire is deleted and forcing the user to write the questionnaire's name to deletion.

Note: the input is case-sensitive to the name of the questionnaire.

Before:
delete-questionnaire-dialog-b4

After:
delete-questionnaire-dialog2
delete-questionnaire-dialog


client/src/app/pages/assessment-management/assessment-settings/assessment-settings-page.tsx
```diff
@@ -99,6 +100,9 @@ const AssessmentSettings: React.FC = () => {
   const [questionnaireToDelete, setQuestionnaireToDelete] =
     React.useState<Questionnaire | null>();
 
+  const [questionnaireNameToDelete, setQuestionnaireNameToDelete] =
+    React.useState<string>("");
```
ibolton336, line 104: Don't we need this ```<string>``` type var here since this is a simple string value and can be inferred from usage. Looking good to me so far, though.

ibolton336 (changes requested):

client/src/app/pages/assessment-management/assessment-settings/assessment-settings-page.tsx
```diff
@@ -414,18 +423,24 @@ const AssessmentSettings: React.FC = () => {
...
+            questionnaireNameToDelete={questionnaireNameToDelete}
+            setQuestionnaireNameToDelete={setQuestionnaireNameToDelete}
+          />
+        }
         confirmBtnVariant={ButtonVariant.danger}
         confirmBtnLabel={t("actions.delete")}
         cancelBtnLabel={t("actions.cancel")}
-        onCancel={() => setQuestionnaireToDelete(null)}
-        onClose={() => setQuestionnaireToDelete(null)}
+        onCancel={() => handleDeleteQuestionnaireClose()}
```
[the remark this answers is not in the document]
avivtur: all we did before was `setQuestionnaireToDelete(null)` now we run `setQuestionnaireNameToDelete("")` as well. Since I need it in 3 different places in this file I wrote a handler function to avoid code duplications

ibolton336 (approved):

sjd78 (changes requested):
Issues I found:

 - The Cancel button on the delete confirm modal is never enabled, but the X icon and the Esc key do work.

 - In other "type the name of the thing to delete" interactions I've used, the name of the thing is provided on the modal.  Right now, if I don't remember the exact full name of the assessment questionnaire, I need to close the confirm box copy the name, then go back in, then paste it.  Adding the name questionnaire being deleted to the confirm modal would help.

avivtur:
Hi @sjd78 
Thank you for taking the time to review this PR! :smile: 

> Issues I found:
> 
> * The Cancel button on the delete confirm modal is never enabled, but the X icon and the Esc key do work.

That's not entirely accurate, the Cancel button is enabled only when submit is enabled as well, but I agree that that is not a wanted behavior, I should expose a new prop the ConfirmModal component to disable the submit only.

> * In other "type the name of the thing to delete" interactions I've used, the name of the thing is provided on the modal.  Right now, if I don't remember the exact full name of the assessment questionnaire, I need to close the confirm box copy the name, then go back in, then paste it.  Adding the name questionnaire being deleted to the confirm modal would help.

Agreed, will follow the example from Console "Delete project" modal:
delete-project

mturley:
@sjd78 I agree, I think there's a cleaner abstraction to be found here. I wonder if we are going to ever want another entity with type-the-name-to-confirm-deletion in this way? If so, maybe we should make a `ConfirmDeleteModal` or `ConfirmDeleteWithNameModal` that either simply wraps `ConfirmDialog` with the state/logic that is currently in the page here, or uses its own `Modal` instead of `ConfirmDialog`. Then we can have a `ConfirmDeleteQuestionnaireModal` that is a trivial usage of it, or just inline that usage in the page (since it would encapsulate all the state).

avivtur:
@mturley @sjd78 @ibolton336 Hi there
Can you please review it again?

client/src/app/components/ConfirmDeleteCatalog/ConfirmDeleteCatalog.tsx
```diff
@@ -0,0 +1,102 @@
...
+    </Button>
+  );
+  return (
+    <Modal
+      id="confirm-delete-dialog"
+      variant={ModalVariant.small}
+      titleIconVariant={titleIconVariant}
+      isOpen={isOpen}
+      onClose={handleClose}
+      aria-label={t("Confirm delete dialog")}
```
sjd78, line 80: Either this key needs to be fixed to something like `"dialog.aria.confirmDelete" or just use the English string.  I've not seen any other externalized aria labels.

avivtur: Right, habit from CNV, will change

client/src/app/components/ConfirmDialog.tsx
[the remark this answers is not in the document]
avivtur: Right, will revert changes

sjd78 (approved):
LGTM

PR #1312, opened 2023-08-24
:sparkles: Add component for downloading the static report

ibolton336:
- Adds a download button to handle downloading the analysis report with custom headers driven by props passed in
- Uses the Fetch API to allow setting custom headers. Converts the received data to a blob & then creates an object URL from that blob. Then we can use that URL with a temporary anchor `<a>` element to trigger the download with a localized loading state. 
- This approach allows us to bin any custom headers we were adding in the proxy and just use the fetch api to set the custom headers. 

image

### DIFF
--- a/client/src/app/pages/applications/components/application-detail-drawer/components/download-button.tsx
+++ b/client/src/app/pages/applications/components/application-detail-drawer/components/download-button.tsx
@@ -1,0 +1,56 @@
+import React, { useState } from "react";
+import { Alert, Button } from "@patternfly/react-core";
+import spacing from "@patternfly/react-styles/css/utilities/Spacing/spacing";
+import { Application } from "@app/api/models";
+import { Spinner } from "@patternfly/react-core";
+import { useDownloadStaticReport } from "@app/queries/download";
+
+export enum MimeType {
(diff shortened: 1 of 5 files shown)


client/src/app/pages/applications/components/application-detail-drawer/components/download-button.tsx
```diff
@@ -0,0 +1,84 @@
...
+
+      const blob = await response.blob();
+      const url = window.URL.createObjectURL(blob);
+      const a = document.createElement("a");
+      a.style.display = "none";
+      a.href = url;
+      a.download = `analysis-${application.name}.` + mimeType;
+
+      document.body.appendChild(a);
+      a.click();
```
mturley, line 56: Oh wow, this is very clever but I'm surprised it's necessary. I think we've used the file-saver library in the past to achieve the same thing, but tbh this might be more elegant.

I wonder if you even need to do the fetch in JS... could we just have the href of this anchor element be the URL we're fetching and let the click event do all the work? If that works that would mean the browser handles the whole download and can show progress natively instead of having it appear as a delayed instant-download.

mturley (approved):

Issue #1262, opened 2023-08-08
Export questionnaires


ibolton336:
The module should be able to export existing questionnaires to YAML as well.

Existing questionnaires will be exportable as YAML files. For that, the user will have to click on the Export option from the kebab menu available on the row that represents the existing questionnaire the user wants to export. After clicking on Export, the system will generate the YAML file and initiate the download in the browser.

PR #1322, opened 2023-08-30
:sparkles: Fetch export questionnaire and application static report with refactor.

gildub:
Use `file-saver` package similarly as the download approach for the application static report.

Refactor the application static report download hook : 
- Move the `useFetchStaticReport` (former  `useDownloadStaticReport`) within `queries/applications.ts` because it's tapping into applications ressource.
- As the download is only fetching (GET) the report data this is not a mutation.

Resolves https://github.com/konveyor/tackle2-ui/issues/1262


client/src/app/pages/applications/components/application-detail-drawer/components/download-button.tsx
```diff
@@ -1,32 +1,19 @@
...
-}) {
-  const {
-    mutate: downloadFile,
-    isLoading,
-    isError,
-  } = useDownloadStaticReport();
+import { MimeType } from "@app/api/models";
+import { useFetchStaticReport } from "@app/queries/applications";
+
+function DownloadButton({ id, mimeType }: { id: number; mimeType: MimeType }) {
```
ibolton336, line 9: Could extract this to a defined interface or type:

```suggestion
function DownloadButton({ id, mimeType }: DownloadButtonProps) {
```

client/src/app/pages/assessment-management/assessment-settings/ExportQuestionnaireDropdownItem.tsx
```diff
@@ -0,0 +1,29 @@
...
+export const ExportQuestionnaireDropdownItem: React.FC<
+  IExportQuestionnaireDropdownItemProps
+> = ({ id }) => {
+  const { t } = useTranslation();
+  const { questionnaire } = useFetchYamlQuestionnaire(id);
+
+  const exportQuestionnaire = () => {
+    if (questionnaire) {
+      const file = new Blob([questionnaire]);
+      saveAs(file, `questionnaire-${id}.yaml`);
```
ibolton336, line 20:
```suggestion
      saveAs(new Blob([questionnaire]), `questionnaire-${id}.yaml`);
```

client/src/app/api/rest.ts
```diff
@@ -748,3 +750,37 @@ export const updateQuestionnaire = (
...
+      if (response.status !== 200) {
+        throw new Error("Network response was not ok when downloading file.");
+      }
+      response.data;
+    })
+    .catch((error) =>
+      console.error("There was an error downloading the file:", error)
+    );
+
+export const getYamlQuestionnaire = (id: number): Promise<void> =>
```
ibolton336, line 775: Changing the name to indicate that this is a download endpoint would be better I think. And removing yaml from the name of the functions would probably be best just incase we need to support other types.

ibolton336 (changes requested):

client/src/app/pages/applications/components/application-detail-drawer/components/download-button.tsx
mturley: Why `function` syntax here and not a const using the `React.FC` type?

mturley: Ah I realized that's preexisting. Maybe we should take this opportunity to make it consistent with our other components

mturley (commented):
+1 to Ian's feedback, LGTM otherwise.

Edit: Mea culpa, I reviewed without testing and didn't catch the below issues. Will review again.

ibolton336:
Something isn't looking right in the drawer 
Screenshot 2023-08-30 at 1 37 18 PM

client/src/app/pages/applications/components/application-detail-drawer/components/download-button.tsx
```diff
@@ -1,32 +1,26 @@
...
-  application: Application;
-  mimeType: MimeType;
-}) {
-  const {
-    mutate: downloadFile,
-    isLoading,
-    isError,
-  } = useDownloadStaticReport();
+}) => {
+  const { report, isLoading, isError } = useFetchStaticReport(id, mimeType);
```
ibolton336, line 18: Looks like this is causing the report to be fetched on load rather than on click when the button is implemented in the drawer. This causes the loading state to be triggered and the request ultimately fails due to nothing being returned from the downloadApplicationAnalysisReport function.

client/src/app/api/rest.ts
```diff
@@ -748,3 +750,37 @@ export const updateQuestionnaire = (
...
+      responseType: "blob",
+      headers: {
+        Accept: `application/x-${type}`,
+      },
+    })
+    .then((response) => {
+      if (response.status !== 200) {
+        throw new Error("Network response was not ok when downloading file.");
+      }
+      response.data;
```
ibolton336, line 769: This is never returned.

ibolton336 (dismissed):

gildub:
@ibolton336, we're getting the "Error downloading report" as well on main branch when clicking on the YAML or Report links. So is that approach is solid enough ?

client/src/app/api/rest.ts
```diff
@@ -314,6 +315,22 @@ export const getApplications = (): Promise<Application[]> =>
...
+  applicationId: number,
+  type: MimeType
+): Promise<void> =>
+  axios.get(
+    `${APPLICATIONS}/${String(applicationId)}/analysis${
+      type === MimeType.TAR ? "/report" : ""
+    }`,
+    {
+      responseType: "blob",
+      headers: {
```
ibolton336, line 328: No accept headers are required for the /report request any longer as I mentioned https://github.com/konveyor/tackle2-ui/pull/1329 . I am not really following the intent of this PR. I think effort would be better spent on one of our open kanban issues. Thx.

client/src/app/pages/applications/components/application-detail-drawer/components/download-button.tsx
```diff
@@ -1,56 +1,57 @@
...
-      )}
-    </>
+    <Button
+      key={mimeType}
+      onClick={handleDownload}
+      id={`download-${mimeType}-button`}
+      variant="link"
+      className={spacing.pXs}
+    >
+      {mimeType === MimeType.YAML ? "YAML" : "Report"}
```
sjd78, line 52: Nit-pick: This text could just be passed in as a prop and set to whatever needed by the consuming component.

sjd78 (approved):
The change is looking nice and clean now.  Only a single nit-pick, but not a blocker.

client/src/app/pages/applications/components/application-detail-drawer/components/download-button.tsx
gildub: @sjd78 , what do you mean ?
The text is determined from the mimeType prop already passed to the DownloadButton.
The text is then wrapped by the Button component corresponds to the Children prop.

sjd78: I mean you could put the regular text here:
https://github.com/konveyor/tackle2-ui/blob/c44218dd54457bd09de662c9580a56474f42f193/client/src/app/pages/applications/components/application-detail-drawer/application-detail-drawer-analysis.tsx#L137-L157

But then the button could look like:
```tsx
                      <DownloadButton
                        application={application}
                        mimeType={MimeType.TAR}
                      >
                        tar
                      </DownloadButton>
```

sjd78: I'm ok with the way it is now.  It's just a possible alternative.

gildub: @sjd78 ,
Oh I see now, thanks for the example.
I don't see the gain, it would add extra lines of code for the same result.

gitdallas (approved):

client/src/app/pages/applications/components/application-detail-drawer/components/download-button.tsx
```diff
@@ -1,56 +1,55 @@
...
-}: {
-  application: Application;
-  mimeType: MimeType;
-}) {
-  const {
-    mutate: downloadFile,
-    isLoading,
-    isError,
-  } = useDownloadStaticReport();
+}) => {
```
sjd78, line 20: You'll need to pull in the `children` as a prop to have the text (or react node) to render...

```jsx

export const DownloadButton: React.FC<IDownloadButtonProps> = ({
  id,
  mimeType,
  children
}) => {
```

and throw `children: React.ReactNode` in the props interface

client/src/app/pages/applications/components/application-detail-drawer/components/download-button.tsx
```diff
@@ -1,56 +1,55 @@
...
-        </>
-      )}
-    </>
+    <Button
+      key={mimeType}
+      onClick={handleDownload}
+      id={`download-${mimeType}-button`}
+      variant="link"
+      className={spacing.pXs}
+    />
```
sjd78, line 51: Still need to render the text here....
```tsx
<Button ...>
  {children}
</Button>
```

sjd78 (commented):
I think one more step to get the button text (or react node) to render is still needed.

client/src/app/pages/applications/components/application-detail-drawer/components/download-button.tsx
gildub: @sjd78, Yeah I realized that while testing latest, thanks.

sjd78 (approved):
LGTM

ibolton336:
Looks like this broke the tar download functionality and the loading spinner in the app analysis detail drawer.

PR #1331, opened 2023-09-06
:seedling: eslint - add react-query checks

sjd78:
Looking at [1], add eslint rules for react-query.  The rules are currently set to `error` in order to highlight what "best practices" will be enforced with this change.

Fixed the reported errors for the eslint plugins:

  - `"@tanstack/query/prefer-query-object-syntax`: fix the `useQuery()`
    to use the query object syntax throughout the code base.

  - `@tanstack/query/exhaustive-deps`:

    - Fix the `queryKey` params to include all of the variables used
      in the `queryFn`.  This is similar behavior to the `useMemo()`
      lint rule.

    - Where a `params` prop was being transformed to a string in the
      `queryKey`, the transform was removed.  The bare `params` object
      is allowed to be used as part of the queryKey.

  - A number of warnings in the files that were touched were also resolved (some prettier, some no any type, etc).

[1] - https://tanstack.com/query/latest/docs/react/eslint/eslint-plugin-query

### DIFF
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -33,6 +33,7 @@ module.exports = {
     "@typescript-eslint",
     "react",
     "react-hooks",
+    "@tanstack/query",
   ],
 
   // NOTE: Tweak the rules as needed when bulk fixes get merged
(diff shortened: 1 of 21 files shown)


sjd78:
No urgency on this PR, I just wanted to post it while I was thinking about it reading through react-query docs yet again.

mturley (approved):

mturley (approved):
LGTM!

PR #1337, opened 2023-09-07
:bug: Regex to avoid substring matching on route paths

### LINKED ISSUE #1335: `formatPath` helper breaks if some params are substrings of others
[The `formatPath` helper found here](https://github.com/konveyor/tackle2-ui/blob/main/client/src/app/Paths.ts#L1), used to interpolate path parameters (`:foo`) in URL strings, has a bug:

```ts
export const formatPath = (path: Paths, data: any) => {
  let url = path as string;

  for (const k of Object.keys(data)) {
    url = url.replace(":" + k, data[k]);
  }

  return url;
};
```

If the URL contains multiple path params that are substrings of each other (like "ruleset" and "rule" in `/issues/all/:ruleset/:rule/affected-applications`), replacing e.g. `:rule` will replace the first half of `:ruleset` as well and result in a malformed URL like `/issues/all/1set/1/affected-applications` . We may want to come back to this and make it use a regex instead (only replace if the placeholder is either followed by a / or is at the end of the string).

[In the Issues page I worked around it](https://github.com/konveyor/tackle2-ui/blob/main/client/src/app/pages/issues/helpers.ts#L116) by including the slashes in my replace string (replace("/:rule/", foo)) but it would be nice to fix `formatPath` and use it there.

ibolton336:
This updates the formatPath function to use a regex pattern that matches :k followed by either a / or the end of the string ($). This way, it won't inadvertently replace parts of other placeholders. 

Resolves #1335

### DIFF
--- a/client/src/app/Paths.ts
+++ b/client/src/app/Paths.ts
@@ -2,7 +2,8 @@ export const formatPath = (path: Paths, data: any) => {
   let url = path as string;
 
   for (const k of Object.keys(data)) {
-    url = url.replace(":" + k, data[k]);
+    const regex = new RegExp(`:${k}(/|$)`, "g");
+    url = url.replace(regex, data[k] + "$1");
   }
 
   return url;
--- a/client/src/app/Paths.ts
+++ b/client/src/app/Paths.ts
@@ -1,14 +1,3 @@
-export const formatPath = (path: Paths, data: any) => {
-  let url = path as string;
-
-  for (const k of Object.keys(data)) {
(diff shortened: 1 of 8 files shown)


mturley (approved):
Awesome! Thanks

sjd78:
No unit test?! :wink:

ibolton336:
> No unit test?! 😉

Added some and moved the util to the utils file.

sjd78 (approved):
LGTM

PR #1343, opened 2023-09-08
:sparkles: Archetype form for Create and Edit

### LINKED ISSUE #1265: Archetypes form
Image

New archetypes can be created by clicking on the Create new archetype button at the top menu. This will open a modal window with the following fields:

- Name: String
- Description: String (Optional)
- Tags: Dropdown, multi selection. All selected options will be displayed as removable labels.
- Stakeholders: Dropdown, multi selection. All selected options will be displayed as removable labels. (Optional)
- Stakeholder group(s): Dropdown, multi selection. All selected options will be displayed as removable labels. (Optional)

sjd78:
Add the Archetype form along with Create and Edit actions from the archetype table.

MSW stubs for the hub archetype api are enabled by default for now.

Resolves #1265

## Screenshots
Create new (empty):
Screenshot from 2023-09-11 19-48-12

Tag selection field open with typeahead active:
Screenshot from 2023-09-11 19-49-19

Create fully filled in:
Screenshot from 2023-09-11 19-49-35

Create in error with an archetype name that already exists:
Screenshot from 2023-09-11 19-49-51

Edit existing:
Screenshot from 2023-09-11 19-52-39

### DIFF
--- a/client/src/app/pages/archetypes/archetypes-page.tsx
+++ b/client/src/app/pages/archetypes/archetypes-page.tsx
@@ -63,7 +63,9 @@ const Archetypes: React.FC = () => {
   const [openCreateArchetype, setOpenCreateArchetype] =
     useState<boolean>(false);
 
-  const [archetypeToEdit, setArchetypeToEdit] = useState<Archetype>();
+  const [archetypeToEdit, setArchetypeToEdit] = useState<Archetype | null>(
+    null
+  );
 
   const { archetypes, isFetching, error: fetchError } = useFetchArchetypes();
 
@@ -292,12 +294,12 @@ const Archetypes: React.FC = () => {
         title={t("dialog.title.updateArchetype")}
         variant="medium"
         isOpen={!!archetypeToEdit}
-        onClose={() => setArchetypeToEdit(undefined)}
+        onClose={() => setArchetypeToEdit(null)}
       >
         <ArchetypeForm
           key={archetypeToEdit?.id ?? -1}
           toEdit={archetypeToEdit}
-          onClose={() => setArchetypeToEdit(undefined)}
+          onClose={() => setArchetypeToEdit(null)}
         />
       </Modal>
 
--- a/client/src/app/pages/archetypes/archetypes-page.tsx
+++ b/client/src/app/pages/archetypes/archetypes-page.tsx
@@ -1,4 +1,4 @@
-import React from "react";
+import React, { useState } from "react";
 import { useTranslation } from "react-i18next";
 import { useHistory } from "react-router-dom";
 import {
@@ -9,6 +9,7 @@ import {
   EmptyStateFooter,
   EmptyStateHeader,
   EmptyStateIcon,
+  Modal,
   PageSection,
   PageSectionVariants,
   Text,
@@ -57,6 +58,13 @@ const Archetypes: React.FC = () => {
   const { t } = useTranslation();
   const history = useHistory();
   const { pushNotification } = React.useContext(NotificationsContext);
+
+  const [openCreateArchetype, setOpenCreateArchetype] =
+    useState<boolean>(false);
+
+  const [archetypeToEdit, setArchetypeToEdit] = useState<Archetype | null>(
+    null
+  );
 
   const { archetypes, isFetching, error: fetchError } = useFetchArchetypes();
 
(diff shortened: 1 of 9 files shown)


client/src/app/pages/archetypes/components/archetype-form/archetype-form.tsx
```diff
@@ -0,0 +1,352 @@
...
+  useUpdateArchetypeMutation,
+} from "@app/queries/archetypes";
+import { duplicateNameCheck, getAxiosErrorMessage } from "@app/utils/utils";
+import { useFetchTagCategories } from "@app/queries/tags";
+
+interface ArchetypeFormValues {
+  name: string;
+  description?: string;
+  comments?: string;
+  criteriaTags: string[]; // TODO: string[] only works if tags are uniquely named globally
```
ibolton336, line 36: I was wrong about this field needing to be surface on UI level. This is an internal hub construct. We only need to pass tags as they are passed on applications today.

client/src/app/pages/archetypes/archetypes-page.tsx
```diff
@@ -270,8 +277,33 @@ const Archetypes: React.FC = () => {
...
+      >
+        <ArchetypeForm onClose={() => setOpenCreateArchetype(false)} />
+      </Modal>
+
+      {/* Edit modal */}
+      <Modal
+        title={t("dialog.title.updateArchetype")}
+        variant="medium"
+        isOpen={!!archetypeToEdit}
+        onClose={() => setArchetypeToEdit(undefined)}
```
ibolton336, line 295: Wondering if we should be using null here. This goes against my understanding of undefined usage vs null within our project. I was thinking we should use null when intentionally setting something as empty/not available & reserve undefined for something unintentionally empty / loading state from restful request or otherwise unexpected cases.

I think its acceptable and perfectly valid to use undefined in this case - just want to be clear on what we decide to use across the project for cases like this.

client/src/app/pages/archetypes/components/archetype-form/archetype-form.tsx
```diff
@@ -0,0 +1,352 @@
...
+  tags: string[]; // TODO: string[] only works if tags are uniquely named globally
+  stakeholders?: object[];
```
ibolton336, line 38: We have types for these guys already in our models file.

ibolton336 (approved):
I think this is in great shape and would be happy to merge it at this stage if we want to push the remaining open tasks to later PRs.

client/src/app/pages/archetypes/archetypes-page.tsx
sjd78: I can swap it to `null`, it is just adding and extra `| null` to the useState type param and a few adjustments.

client/src/app/pages/archetypes/components/archetype-form/archetype-form.tsx
sjd78: Humm -- I still think that the 2 tag files are needed on UI...
 1. Criteria, so people can choose the tags that go in to the archetype-application matching algo
 2. "Static" archetype, so people can choose tags that should be associated with applications just by being matched to the archetype

There is a difference between what types the form values have and what is in the data model.  Currently, the `Autocomplete` component only supports a single list of strings for everything.  That doesn't line up nicely with an array of Tags or similar.  Thankfully when I tested creating Tags from curl, the names did need to be globally unique (regardless of what tag category tags belong to), so the current impl of Autocomplete is sufficient.

client/src/app/pages/archetypes/components/archetype-form/archetype-form.tsx
sjd78: I saw them.  Those should be used in the model itself, but here in the form field type, I'm planning on switching it to `string[]` just like for the tags.

sjd78:
> I think this is in great shape and would be happy to merge it at this stage if we want to push the remaining open tasks to later PRs.

I'd like to throw the Stakeholders in there with the autocomplete and then move forward.  That shouldn't take me much more time today after sorting out how to use Autocomplete for tags.

client/src/app/pages/archetypes/components/archetype-form/archetype-form.tsx
ibolton336: @sjd78 I think the hub is taking care of that logic for us. It should work similarly to how the application form works. Can just pass tags as they are. The hub should know which tag belongs to what category & what tags are used in the matching logic. 
Screenshot 2023-09-11 at 12 21 30 PM


As far as the autocomplete component goes, I think we should just show the list of tags with no category for this first pass. And address the missing visual component of showing the associated category after feature freeze.

client/src/app/pages/archetypes/components/archetype-form/archetype-form.tsx
ibolton336: Oh right! Carry on.

client/src/app/pages/archetypes/components/archetype-form/archetype-form.tsx
sjd78: Sure.  The form will display whatever data is provided for an edit and just empty values for a new archetype.

The form interface uses a `string[]` instead of a `TagRef[]` so the `Autocomplete` component is easier to use (via the `ItemsSelect` wrapper component).

As for the "criteria tags" and "archetype tags" fields, they are both still required inputs on the enhancement: https://github.com/konveyor/enhancements/blob/90b827b68cc367284a66bf66f087d5c263487e05/enhancements/assessment-module/README.md#creating-new-archetypes

Are you talking about swapping the actual Archetype API model from `Tag[]` to `TagRef[]`?

PR #1345, opened 2023-09-11
:sparkles: Selectors to uniquely identify dropdowns

gildub:
@ibragins explained QE needs : 
- A consistent selector across the UI to identify kebab in toolbar, for instance id="toolbar-kebab"
- A consistent selector across the UI to distinct, on a given table row between, between a "pencil" button and the "actions" kebab. There is no need to uniquely identify those selector by row, QE already has selectors for that purpose.

We can see those IDs in action in current PR in applications-table-assessment.tsx.

Meanwhile that cannot be easliy added to the legacy tables (PF4 or PF5 deprecated table) because we don't have access to the wrapper component. Therefore will have to wait for legacy table migration to be finished first.  

This applies to every page having a top kebab (Toolbar) and rows with several actions button/kebab: 
- [x] applications-table-assessment.tsx
- [x] applications-table-analysis.tsx
- [x] migration-waves.tsx

On the MigrationWaves page, the deprecated dropdown has been replaced with PF5 version.

https://github.com/konveyor/tackle2-ui/issues/443

### DIFF
--- a/client/src/app/pages/applications/applications-table-analyze/applications-table-analyze.tsx
+++ b/client/src/app/pages/applications/applications-table-analyze/applications-table-analyze.tsx
@@ -514,7 +514,7 @@ export const ApplicationsTableAnalyze: React.FC = () => {
                 </RBAC>
               </ToolbarItem>
               {dropdownItems.length ? (
-                <ToolbarItem>
+                <ToolbarItem id="toolbar-kebab">
                   <Dropdown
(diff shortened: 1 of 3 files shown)


ibolton336 (approved):

ibolton336:
@istein1

sshveta:
Very helpful PR @ibolton336 . Thanks for these changes.

PR #1349, opened 2023-09-12
:sparkles: Add stakeholders and stakeholder group fields to assessment wizard

ibolton336:
- Uses new Autocomplete field wrapper in the assessment wizard. 
- Moves the ItemsSelect component to the root level of the app for code sharing


sjd78 (approved):
Nice! LGTM

Issue #1332, opened 2023-09-06
Modify assessment based on tags that are present in the associated application


ibolton336:
- When creating the initial assessment, we need to mock the skip, add, remove, etc tag functionality

Issue #1324, opened 2023-08-31
Assessment status in app-assessments-table needs updating 


ibolton336:
- With the change to multiple questionnaires per assessment, we need a way to convey this in the ui. [Sketch
](https://www.sketch.com/s/af50e991-8061-4b69-84c1-c11241e274f2/a/mPOgDzJ)
![image](https://github.com/konveyor/tackle2-ui/assets/11218376/bf77c9f3-7209-4eaf-bb0e-1fcefca37254)

- Update - there will be an assessed boolean on the application struct that we can leverage for this.

Issue #1278, opened 2023-08-10
Assess Archetypes flow


ibolton336:
Archetypes will be assessable by clicking on the Assess option from the kebab menu from a row in the table from the [archetypes view](https://github.com/konveyor/enhancements/blob/46eb949447f43a39d04b4db92886a01e8f0f227d/enhancements/assessment-module/README.md#archetypes-main-view). Doing that will lead to the questionnaire list view and kickstart the assessment process in the exact same way described before for applications.

Note: Conditional questions on a questionnaire belonging to the assessment of an archetype will only take into account tags directly associated with the archetype, and will ignore tags coming from applications associated to that archetype.

Note: Tags in the assessment source of an application coming from the assessment of an associated archetype are transitive. If the application stops being associated to the archetype or the archetype is deleted, the tags will be removed from the application.

PR #1368, opened 2023-09-15
:sparkles: Fix missing archetype flow pieces

### LINKED ISSUE #1298: Check existing archetypes when starting an app assessment
- Wire up check for existing assessments from archetypes when starting an app assessment 
- Add a view assessments action to the assessment modal as shown in the wireframe. It is unclear if accept & view assessments take the user to the same assessment actions page with different state.  TODO: [View assessments page for existing assessments](https://github.com/konveyor/tackle2-ui/issues/1301)

Image

ibolton336:
Closes #1298 #1278 #1332 #1324 
- Adjusts routes for archetype assessment flow vs application assessment flow
- Creates a hook for looking up app or archetype based on their ID to avoid conditional hook calls. 
- Wire up create assessment override check 
Screenshot 2023-09-18 at 12 33 20 PM
-Drive assessment wizard from assessment rather than questionnaire

- With this PR, assessment flow is fully working for apps and archetypes against the hub outstanding PR.


client/src/app/components/questionnaire-summary/questionnaire-summary.tsx
```diff
@@ -36,18 +36,16 @@ interface QuestionnaireSummaryProps {
...
 const QuestionnaireSummary: React.FC<QuestionnaireSummaryProps> = ({
   summaryData,
   summaryType,
   isFetching = false,
   fetchError = null,
-  isArchetype,
 }) => {
   const { t } = useTranslation();
-
+  const isArchetype = location.pathname.includes("/archetypes/");
```
mturley, line 48: Hm, what's the reason for changing this from a prop? If we ever needed to show the questionnaire content for an archetype outside the archetypes page (as a modal or something) this would be an issue. Not that we ever need to be able to do that necessarily, it's probably fine, just curious

client/src/app/hooks/useFetchArchetypeOrApplicationByAssessment.ts
```diff
@@ -0,0 +1,42 @@
...
+  );
+
+  return {
+    data,
+    isLoading,
+    isError,
+    error,
+    isArchetype,
+  };
+}
```
mturley, line 42: So, IMO this coupling of the two types of data being fetched into one useQuery isn't necessary. We could keep the two hooks separate and call both, but just pass `enabled: false` to the one that isn't relevant (expose that as an argument on the hook).

But I understand that would mean undoing a chunk of your work so I'm fine with leaving it as-is in the interest of time if necessary.

client/src/app/pages/assessment/components/assessment-actions/assessment-actions-page.tsx
```diff
@@ -11,19 +11,21 @@ import { Link, useParams } from "react-router-dom";
...
 
-  const { application } = useFetchApplicationByID(applicationId || "");
-  const { archetype } = useFetchArchetypeById(archetypeId || "");
+  const { data } = useFetchArchetypeOrApplicationByAssessment(
+    isArchetype,
+    isArchetype ? archetypeId : applicationId
+  );
 
-  console.log("archetype", archetype);
+  console.log("archetype or app data", data);
```
mturley, line 28: Do we still need the console.log here?

client/src/app/pages/assessment/components/assessment-actions/assessment-actions-page.tsx
```diff
@@ -47,18 +49,18 @@ const AssessmentActions: React.FC = () => {
       </PageSection>
       <PageSection>
         {isArchetype ? (
-          <ConditionalRender when={!archetype} then={<AppPlaceholder />}>
+          <ConditionalRender when={!data} then={<AppPlaceholder />}>
             <TextContent>
-              {archetype ? (
-                <AssessmentActionsTable archetype={archetype} />
+              {data ? (
+                <AssessmentActionsTable archetype={data as Archetype} />
```
mturley, line 55: If we do separate the two query hooks back out and only have one enabled, we can just pass the data from the relevant query hook here on lines 55 and 63 instead of using the `as` assertions.

mturley (commented):
A couple of nitpicks here that may be a pain to address. I'm fine leaving them if you want, but wanted to express them anyway to see if you agree.

client/src/app/components/questionnaire-summary/questionnaire-summary.tsx
ibolton336: I changed everything to be driven off of the location path because I was seeing some undefined props on hard navigation to some of the routes. Maybe that would be better handled by redirects. Some more testing is probably warranted here.

client/src/app/hooks/useFetchArchetypeOrApplicationByAssessment.ts
ibolton336: Not a big deal to change. I think that is probably a good call.

client/src/app/pages/assessment/components/assessment-actions/assessment-actions-page.tsx
ibolton336: Makes sense - thanks for the review!

client/src/app/components/questionnaire-summary/questionnaire-summary.tsx
sjd78: If you made the `isArchetype` prop required and provided a default value like `isArchetype = false`, then you would not see an undefined value.

Alternatively, you could use the `useLocation` or `useRouteMatch` hook from react-router to be a bit more robust in the `isArchetype` check.

client/src/app/components/questionnaire-summary/questionnaire-summary.tsx
```diff
@@ -90,13 +88,23 @@ const QuestionnaireSummary: React.FC<QuestionnaireSummaryProps> = ({
...
       <Breadcrumb>
         <BreadcrumbItem>
-          <Link
-            to={formatPath(Paths.applicationAssessmentActions, {
-              applicationId: (summaryData as Assessment)?.application?.id,
-            })}
-          >
-            Assessment
-          </Link>
+          {isArchetype ? (
```
sjd78, line 91: This _could_ be push into the `to` prop instead of containing the full tag

client/src/app/pages/assessment/components/assessment-page-header.tsx
```diff
@@ -19,38 +20,38 @@ export const AssessmentPageHeader: React.FC<AssessmentPageHeaderProps> = ({
 }) => {
   const { t } = useTranslation();
   const history = useHistory();
+  const isArchetype = location.pathname.includes("/archetypes/");
```
sjd78, line 23: This has been repeated enough times that it is a good candidate to create a hook for....

```js
const isArchetype = useIsArchetypeFromLocation();
```

sjd78 (commented):
Another way to organize the component would be to have an `ApplicationAssessmentSummary` and an `ArchetypeAssessmentSummary`.  Then have those just wrap `AssessmentSummary` with a bunch of constants (isArchetype, title, urlPath1, urlPath2, etc).  The child components would need to have access to those props (just passing them down is ok), but they wouldn't need to know how to select them.

For things like:
```jsx
          <Text component="p">
            {isArchetype ? archetype?.name : application?.name}
          </Text>
```
The name can just come from a `EntityWithName` kind of type:
```jsx
  ... // like in `items-select.tsx`:
  EntityWithName extends { name?: string };

  ...
  entityWithName: EntityWithName

  ...
          <Text component="p">{entityWithName?.name}</Text>
```

Then any `archetype` or `application` could be given as the `entity`.

None of that is a blocker though.

client/src/app/pages/assessment/components/assessment-page-header.tsx
ibolton336: All good ideas. Will work on refactoring.

ibolton336:
> Another way to organize the component would be to have an `ApplicationAssessmentSummary` and an `ArchetypeAssessmentSummary`. Then have those just wrap `AssessmentSummary` with a bunch of constants (isArchetype, title, urlPath1, urlPath2, etc). The child components would need to have access to those props (just passing them down is ok), but they wouldn't need to know how to select them.
> 
> For things like:
> 
> ```js
>           <Text component="p">
>             {isArchetype ? archetype?.name : application?.name}
>           </Text>
> ```
> 
> The name can just come from a `EntityWithName` kind of type:
> 
> ```js
>   ... // like in `items-select.tsx`:
>   EntityWithName extends { name?: string };
> 
>   ...
>   entityWithName: EntityWithName
> 
>   ...
>           <Text component="p">{entityWithName?.name}</Text>
> ```
> 
> Then any `archetype` or `application` could be given as the `entity`.
> 
> None of that is a blocker though.

I like this pattern. I think it is a candidate for a refactor post feature freeze. I am hesitant to attempt this with their being a third usage for this summary component in the assessment-management-settings page that is archetype/application agnostic.

client/src/app/hooks/useIsArchetype.ts
```diff
@@ -0,0 +1,17 @@
...
+
+const useIsArchetype = () => {
+  const location = useLocation();
+  const [isArchetype, setIsArchetype] = useState(
+    location.pathname.includes("/archetypes/")
+  );
+
+  useEffect(() => {
+    setIsArchetype(location.pathname.includes("/archetypes/"));
+  }, [location.pathname]);
```
mturley, line 12: You don't need the state or the effect here because all they are doing is duplicating the new value from `useLocation()` when it causes a re-render. If that `useEffect` is being triggered, it means we already have an updated `location.pathname` value in scope (because `useLocation` watches for it to change and re-renders), and we can just return that directly instead of putting it in state. So I think this whole hook can be rewritten as:

```ts
const useIsArchetype = () => useLocation().pathname.includes("/archetypes/");
```

mturley (changes requested):
One last nitpick here @ibolton336 . I do think there is some more room for cleanup here but I don't think it needs to block feature freeze. We should maybe open a tech debt issue to make sure we follow up.

client/src/app/hooks/useIsArchetype.ts
ibolton336: Updated - thanks!

client/src/app/components/ConfirmDialog.tsx
```diff
@@ -68,6 +76,28 @@ export const ConfirmDialog: React.FC<ConfirmDialogProps> = ({
...
+
+  const actions = [confirmBtn];
+
+  if (cancelBtn) {
+    actions.push(cancelBtn);
+  }
+
+  if (customActionBtn) {
+    actions.splice(1, 0, customActionBtn);
+  }
```
sjd78, line 100:
```jsx
const actions = [
  confirmBtn,
  customActionBtn,
  cancelBtn,
].filter(Boolean)
```

Then any of the vars that are undefined/null just get filtered out.

client/src/app/pages/assessment/components/assessment-actions/components/dynamic-assessment-actions-row.tsx
```diff
@@ -129,18 +137,59 @@ const DynamicAssessmentActionsRow: FunctionComponent<
...
+      for (const archetypeRef of application.archetypes) {
+        try {
+          const assessments = await getAssessmentsByItemId(
+            true,
+            archetypeRef.id
+          );
+
+          if (assessments && assessments.length > 0) {
+            setArchetypeRefToOverride(archetypeRef);
+            return;
```
sjd78, line 158: is this `return` needed or is it supposed to be a `break`?

sjd78 (changes requested):
Just blocked on the return on the for loop, otherwise looks good.

client/src/app/pages/assessment/components/assessment-actions/components/dynamic-assessment-actions-row.tsx
ibolton336: Whoops. Thanks good catch.

mturley (approved):

sjd78 (approved):
LGTM

Issue #1264, opened 2023-08-08
Archetypes Table


ibolton336:
Each row will represent an existing archetype, containing the following information:

- Name: Name of the archetype
- Description: Brief description of the archetype
- Tags: The list of tags that define the archetype.
- Maintainers: List of stakeholders and stakeholder groups that maintain the archetype in the organization. This is metainformation that could be useful to navigate the organization and reach out to key people when more information about an archetype is needed.
- Applications: List of applications that are associated with the archetype. Association happens automatically via a match of the tags that define with the archetype with the ones that are associated with the application.

The header of the archetypes table will include filters that include Name, Description, Tags and Maintainers as criteria. Alongside the filtering, a button will be displayed to create a new archetype.

Each row will have a menu with the following options:

Assess: Launches the assessment for a given archetype.
Review: Navigates to the review view for the archetype.
Edit: Opens a modal window to edit an existing archetype.
Delete: Deletes the archetype.

sjd78:
After #1308, here is a list of things that need to be added to the table:

- [ ] Table action kebab and confirm modals
- [ ] Pagination (IDK if it would be client or server side)
- [ ] Table column components need to be cleaned up:
  - [ ] Description column: truncated text with tooltip for full description as needed)
  - [ ] Tags column: 
    - [ ] How to combine the criteria and archetype tags?
    - [ ] How to sort?
    - [ ] Do the tag categories matter here?
    - [ ] Reuse of the `ApplicationTagLabel` component for `Label` rendering?
  - [ ] Maintainers column:
    - [ ] Don't show the full name, generate initials
    - [ ] Sort individual stakeholders with stakeholder groups?
    - [ ] Add tooltips for each Label with the full name?
  - [ ] Application column:
    - [x] Basic "No Applications" message when no matching applications (as determined by hub response)
    - [x] Basic "Application count" message (as determined by hub response)
    - [ ] When count >0 render a link to navigate to the application inventory assessment page with filters set to show the applications for the archetype? This is not explicitly in the mockups or enhancement doc, so probably best to kick it off to a new issue.

mturley:
I'm assuming pagination will be client side here, at least until hub folks tell us there's a scalability issue with that (I'm not sure if this endpoint supports the pagination args).

sjd78:
With further changes detailed in JIRA tickets, this issue can be considered complete.

PR #1370, opened 2023-09-15
:sparkles: Add client-side pagination to Archetypes table

mturley:
Part of #1264, see [checklist here](https://github.com/konveyor/tackle2-ui/issues/1264#issuecomment-1702928566).


sjd78 (approved):
Ok, that's just far too straight forward to add client side pagination like that.

mturley:
Thanks @sjd78 ! I honestly believe most table stuff can be that straightforward with the right abstraction. I've seen so many tables in products become a mess of repeated and confusing boilerplate, and the success so far we've had with the table-controls stuff gives me hope.

PR #1371, opened 2023-09-15
:sparkles: Add Duplicate Archetype action

mturley:
Part of #1264, see [checklist here](https://github.com/konveyor/tackle2-ui/issues/1264#issuecomment-1702928566).

The Duplicate action is a cross between create and edit: a "create new archetype" form is opened, but its fields are all prefilled as if we are editing an existing archetype. The prefilled name has an appended " (duplicate)" suffix on it.

Just for fun, in case a user duplicates the same archetype multiple times without touching the default name, we check to see if the new default name is taken and use " (duplicate N)" as the suffix, with N incrementing until we get a name that isn't taken. This is probably unnecessary, but I found it annoying that the "name must be unique" error didn't appear until I touched the field, so this makes sure that case never comes up.

Because ArchetypeForm already had the logic for prefilling the form for editing, all we needed to do was pass in an `archetype` and an `isDuplicating` boolean which slightly changes the behavior:
* When validating that the name is unique, we don't treat the archetype being duplicated as the "current item", so it is not an exception to that validation like it needs to be when editing (if you input the same name of the item you're editing that's valid, but not so for duplicating).
* We have the default name behavior with the suffix as described above.
* Normally the submit button is disabled if the form isn't dirty, but the prefilled form with a generated name is valid for submission, so we allow a non-dirty form to be submitted when `isDuplicating` is true. This way you can rapid-fire duplicating things without touching the form if you so desire (useful for testing the pagination in https://github.com/konveyor/tackle2-ui/pull/1370 😄)
* The submit button says "Create", and on submit we create instead of updating.

### DIFF
--- a/client/src/app/pages/archetypes/archetypes-page.tsx
+++ b/client/src/app/pages/archetypes/archetypes-page.tsx
@@ -68,6 +68,9 @@ const Archetypes: React.FC = () => {
     null
   );
 
+  const [archetypeToDuplicate, setArchetypeToDuplicate] =
+    useState<Archetype | null>(null);
+
   const { archetypes, isFetching, error: fetchError } = useFetchArchetypes();
 
   const onError = (error: AxiosError) => {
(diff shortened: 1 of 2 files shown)


sjd78 (approved):
LGTM.

The `getDefaultName()` function was a bit jarring but after reading it slowly makes sense.

Issue #1305, opened 2023-08-23
Reports page api updates


ibolton336:
The reports page needs to be driven by all available questionnaire data. There will also be a dropdown to select which archetype to drive the reports data from at the top of the page with the default being all existing questionnaires in the database with a "required" flag set to true

https://www.sketch.com/s/af50e991-8061-4b69-84c1-c11241e274f2/a/VEvWnDo

https://www.sketch.com/s/af50e991-8061-4b69-84c1-c11241e274f2/a/zyq8Y1M

To do:
  - Current Landscape:
    - [x] Add dropdown to select a questionnaire: See #1374
    - [x] Adjust the donut charts to display aggregated risk when "All questionnaires" is selected: See #1395
    - [x] Adjust the donut charts to display aggregated risk, with `riskMessages` from the questionnaire displayed as a sub-title, when a single questionnaire is selected: See See #1395
  - Identified risks:
    - [ ] rearrange to accommodate Rationale and Mitigation strategies data as expandable rows

sjd78:
Enhancement document section: https://github.com/konveyor/enhancements/blob/90b827b68cc367284a66bf66f087d5c263487e05/enhancements/assessment-module/README.md#changes-in-the-application-reports-view

sjd78:
With the details on this issue a bit old now, and direction changing slightly in advance of the next set of changes, this issue can be closed.

PR #1374, opened 2023-09-19
:sparkles: Reports: select Questionnaires

gildub:
Shows selection of possible questionnaires of current assessments
Pass to `<Landscape>`:
- If "All questionnaires" option has been selected then the whole assessments list is passed
- When a specific questionnaire is selected then only the assessments bound to that questionnaire are passed.

The consolidation of the risks is left to be handled by the Landscape component in following PR.

https://github.com/konveyor/tackle2-ui/issues/1305

image


client/src/app/pages/reports/reports.tsx
```diff
@@ -31,11 +37,20 @@ import { Landscape } from "./components/landscape";
...
   // i18
   const { t } = useTranslation();
 
+  const [isQuestionnaireSelectOpen, setIsQuestionnaireSelectOpen] =
+    React.useState<boolean>(false);
+
+  const [selectedQuestionnaire, setSelectedQuestionnaire] =
+    React.useState<string>("All questionnaires");
+
+  const { questionnaires } = useFetchQuestionnaires();
```
ibolton336, line 52: Needs to be driven from taken assessments. Otherwise we won't have any calculated risk fields.

ibolton336: So this page shouldn't need to use any questionnaire data at all. It is showing a report based on all assessed questionnaires in the database in the form of "assessments". We should be doing a getAssessments() fetch at the root  of the page and driving the data from that.

client/src/app/pages/reports/reports.tsx
```diff
@@ -31,11 +37,20 @@ import { Landscape } from "./components/landscape";
...
   // i18
   const { t } = useTranslation();
 
+  const [isQuestionnaireSelectOpen, setIsQuestionnaireSelectOpen] =
+    React.useState<boolean>(false);
+
+  const [selectedQuestionnaire, setSelectedQuestionnaire] =
+    React.useState<string>("All questionnaires");
+
+  const { assessments } = useFetchAssessments();
```
ibolton336, line 52: +1

client/src/app/pages/reports/reports.tsx
```diff
@@ -66,6 +81,34 @@ export const Reports: React.FC = () => {
...
+
+  const onSelectQuestionnaire = (
+    _event: React.MouseEvent<Element, MouseEvent> | undefined,
+    value: string | number | undefined
+  ) => {
+    setSelectedQuestionnaire(value as string);
+    setIsQuestionnaireSelectOpen(false);
+  };
+
+  const questionnaires = assessments.reduce((result: string[], item) => {
```
ibolton336, line 105: Shouldn't need to do this IMO. The data you need is all available from the assessment object itself I think.

gildub: @ibolton336, we need to reduce the assessments list because the same questionnaire can be used in several assessment and in order to not have repetition of the questionnaires in the select.

client/src/app/pages/reports/reports.tsx
```diff
@@ -78,13 +122,53 @@ export const Reports: React.FC = () => {
...
+                            </SelectOption>
+                          ))}
+                        </Select>
+                      </FlexItem>
+                    </Flex>
                   </CardHeader>
                   <CardBody>
                     <Bullseye>
-                      <Landscape />
+                      <Landscape
```
ibolton336, line 162: Need to be passing the assessment here.

ibolton336 (changes requested):
Data still being driven from questionnaires.

client/src/app/pages/reports/reports.tsx
ibolton336: We don't care about questionnaires at all. There could be 200 assessments all driven from the same questionnaire. We just care about the assessment data. I think you can safely remove every reference to questionnaires in this entire file.

client/src/app/pages/reports/reports.tsx
gildub: @ibolton336, yes, I've adjusted.

client/src/app/pages/reports/reports.tsx
gildub: @ibolton336,

With following data set (extracted from latest) which comes directly from the assessments we need to reduce the redundant questionnaire.

```ts
[
  {
    "id": 1,
    "application": {
      "id": 1,
      "name": "App1"
    },
    "questionnaire": {
      "id": 1,
      "name": "Test questionnaire"
    },
   ..../....
 },
  {
    "id": 2,
    "application": {
      "id": 2,
      "name": "App2"
    },
    "questionnaire": {
      "id": 2,
      "name": "Test questionnaire 2"
    },
   ..../....
 },
{
    "id": 3,
    "application": {
      "id": 3,
      "name": "App3"
    },
    "questionnaire": {
      "id": 1,
      "name": "Test questionnaire"
    },
   ..../....
 },
 ```
 
Please try by yourself by creating 3 applications in which 2 of them use the same questionnaire.

Comment (ibolton336) [reply client/src/app/pages/reports/reports.tsx]: Ah Ok.

**Comment by ibolton336** [comment]
Screenshot 2023-09-19 at 6 28 48 PM

**Review by ibolton336 — CHANGES_REQUESTED** [review]
Reports page crashes on load when navigating from applications section

**Comment by gildub** [comment]
@ibolton336, loading has been addressed.

File: client/src/app/pages/reports/components/landscape/landscape.tsx [hunk 5]
```diff
@@ -91,36 +95,36 @@ export const Landscape: React.FC = () => {
             <Donut
               value={landscapeData.low}
               total={applications.length}
-              color={RISK_LIST["green"].hexColor}
-              riskLabel={t("terms.lowRisk")}
-              riskDescription={t("terms.cloudNativeReady")}
+              color={RISK_LIST["red"].hexColor}
+              riskLabel={t("colors.red")}
+              // riskDescription={}
             />
```
sjd78, line 101: If this one is 'red', should the value be switched to `landscapeData.high`?

client/src/app/pages/reports/components/landscape/landscape.tsx
```diff
@@ -91,36 +95,36 @@ export const Landscape: React.FC = () => {
...
             <Donut
               value={landscapeData.high}
               total={applications.length}
-              color={RISK_LIST["red"].hexColor}
-              riskLabel={t("terms.highRisk")}
-              riskDescription={t("terms.unsuitableForContainers")}
+              color={RISK_LIST["green"].hexColor}
+              riskLabel={t("colors.green")}
+              // riskDescription={}
             />
```
sjd78, line 119: If this one is 'green', should the value be switched to `landscapeData.low`?

client/src/app/pages/reports/reports.tsx
```diff
@@ -78,13 +129,54 @@ export const Reports: React.FC = () => {
...
+                          isOpen={isQuestionnaireSelectOpen}
+                          selected={selectedQuestionnaire}
+                          onSelect={onSelectQuestionnaire}
+                          onOpenChange={(_isOpen) =>
+                            setIsQuestionnaireSelectOpen(false)
+                          }
+                          toggle={toggleQuestionnaire}
+                          shouldFocusToggleOnSelect
+                        >
+                          <SelectOption key={0} value="All questionnaires">
```
sjd78, line 152: I suggest using a constant value here to make sure it is the same when you need to test it lower...

```jsx
  const ALL_SELECTED = "all-questionnaires";
  ...

  value={ALL_SELECTED}
```

client/src/app/pages/reports/reports.tsx
```diff
@@ -78,13 +129,54 @@ export const Reports: React.FC = () => {
...
+                        </Select>
+                      </FlexItem>
+                    </Flex>
                   </CardHeader>
                   <CardBody>
                     <Bullseye>
-                      <Landscape />
+                      <Landscape
+                        assessments={
+                          selectedQuestionnaire === "All questionnaires"
```
sjd78, line 171: Swap out the string for a constant for safety

client/src/app/pages/reports/reports.tsx
```diff
@@ -78,13 +129,54 @@ export const Reports: React.FC = () => {
...
+                          toggle={toggleQuestionnaire}
+                          shouldFocusToggleOnSelect
+                        >
+                          <SelectOption key={0} value="All questionnaires">
+                            All questionnaires
+                          </SelectOption>
+                          {questionnaires.map((questionnaire, index) => (
+                            <SelectOption
+                              key={index}
+                              value={questionnaire.name}
```
sjd78, line 158: Just curious... Why use `name` here instead of just `id`?

client/src/app/pages/reports/reports.tsx
```diff
@@ -66,6 +86,37 @@ export const Reports: React.FC = () => {
...
+
+  const onSelectQuestionnaire = (
+    _event: React.MouseEvent<Element, MouseEvent> | undefined,
+    value: string | number | undefined
+  ) => {
+    setSelectedQuestionnaire(value as string);
+    setIsQuestionnaireSelectOpen(false);
+  };
+
+  const questionnaires: Ref[] =
```
sjd78, line 110: So this is really a list of questionnaires that have been answered in at least one assessment right?

Maybe name it something like `answeredQuestionnaires` or similar?

sjd78 (changes requested):
I don't actually see where the `Landscape` component is sensitive to a selected questionnaire.  What am I missing?

client/src/app/pages/reports/components/landscape/landscape.tsx
```diff
@@ -48,7 +48,11 @@ const extractLandscapeData = (
   return { low, medium, high, unassessed };
 };
 
-export const Landscape: React.FC = () => {
+interface ILandscapeProps {
+  assessments: Assessment[];
+}
+
+export const Landscape: React.FC<ILandscapeProps> = ({ assessments }) => {
```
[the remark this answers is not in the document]
gildub: @sjd78, that's right, sorry for not being explicit enough about that in the description.
The risk management data needs to be available for the LandscapeProps to exploit it.
So we can leave that for following PR.

client/src/app/pages/reports/reports.tsx
gildub: @sjd78, +1 for using a const meanwhile with a name value (please read below for explanation).

client/src/app/pages/reports/reports.tsx
gildub: @sjd78, normally yes it would make sense to use questionnaire.id for selected value.
Since the questionnaire.name is a unique key constraint there is no reason to go back and forth (see note below)  with the questionnaire id to uniquely identify the questionnaire and pass it along.

Note: Because PF select doesn't allow in the OnSelect() to handle an object, such as Ref, (onSelect) we'd have to get the name of the questionnaire to match the selected value.

client/src/app/pages/reports/reports.tsx
gildub: @sjd78 , that's right. Thanks for suggestion

ibolton336 (approved):

sjd78 (approved):

PR #1376, opened 2023-09-20
:sparkles: Review archetypes

### LINKED ISSUE #1279: Review Archetypes flow
The review flow will stay mostly the same, but now architects should be able to review both archetypes from the Archetypes view or individual applications from the *Application inventory view. As with assessments, applications associated to an archetype that has been reviewed will inherit the review, and it will be possible to override it with a dedicated review as well.

ibolton336:
- Review Archetypes 
Closes #1279

### DIFF
--- a/client/src/app/pages/archetypes/archetypes-page.tsx
+++ b/client/src/app/pages/archetypes/archetypes-page.tsx
@@ -69,6 +69,7 @@ const Archetypes: React.FC = () => {
   const [archetypeToEdit, setArchetypeToEdit] = useState<Archetype | null>(
     null
   );
+  const [reviewToEdit, setReviewToEdit] = React.useState<number | null>(null);
 
   const [archetypeToDuplicate, setArchetypeToDuplicate] =
     useState<Archetype | null>(null);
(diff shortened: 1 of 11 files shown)


client/src/app/queries/reviews.ts
```diff
@@ -89,39 +92,55 @@ export const useDeleteReviewMutation = (
   });
 };
 
-export function useGetReviewByAppId(applicationId: string | number) {
+export function useGetReviewByItemId(
+  id?: string | number,
+  isArchetype?: boolean
```
sjd78, line 97: Doesn't look like this param is used

client/src/app/pages/archetypes/archetypes-page.tsx
```diff
@@ -68,6 +68,7 @@ const Archetypes: React.FC = () => {
   const [archetypeToEdit, setArchetypeToEdit] = useState<Archetype | null>(
     null
   );
+  const [reviewToEdit, setReviewToEdit] = React.useState<number | null>(null);
```
sjd78, line 71: `reviewToEdit` doesn't seem to be used anywhere.  Would it be enough to just do the history push?

client/src/app/pages/review/components/review-form/review-form.tsx
```diff
@@ -107,39 +112,11 @@ export const ReviewForm: React.FC<IReviewFormProps> = ({
...
-  //       onSaved(response);
-  //     })
-  //     .catch((error) => {});
-  // };
   const queryClient = useQueryClient();
   const onHandleUpdateReviewSuccess = () => {
     queryClient.invalidateQueries([
       reviewsByItemIdQueryKey,
-      application.review?.id,
+      application?.review?.id,
```
sjd78, line 119: Would it make more sense to push the query invalidation to the `updateReviewMutation` itself instead of leaving it here?

sjd78 (commented):
Just a few minor comments.  No real blockers.

Also, humm...interesting that the review data for an archetype isn't shown anywhere else other than the review view of that archetype.  I don't see it anywhere in the [enhancement doc](https://github.com/konveyor/enhancements/blob/90b827b68cc367284a66bf66f087d5c263487e05/enhancements/assessment-module/README.md), but I'd think it'll probably need a set of views added similar to how the applications show reviews...

ibolton336:
WIP - pending hub change

sjd78 (approved):
LGTM

Issue #1366, opened 2023-09-15
Archetype drawer


sjd78:
From the archetypes main view, clicking on a row will open a side drawer:
![image](https://github.com/konveyor/tackle2-ui/assets/3985964/00f46e52-db7d-4639-8e94-441df0606eb4)

Reference: https://github.com/konveyor/enhancements/blob/90b827b68cc367284a66bf66f087d5c263487e05/enhancements/assessment-module/README.md#archetypes-main-view

PR #1377, opened 2023-09-20
:sparkles: Archetype drawer

sjd78:
Enable row selection on the Archetype table such that clicking an archetype row will display a details drawer with the archetype's details.

`PageDrawerContent` in `PageDrawerContext.tsx` extended to have separate drawer header and content components.  This allows the header and the content to have appropriate patternfly styling.

Resolves: #1366

Screen shot:
image


mturley (approved):
I'm late here, but looks great :)

PR #1382, opened 2023-09-20
:bug: Fix translation keys for import questionnaire form

sjd78:
The import form had incorrect translation keys.  Keys are adjusted to use the secondary grouping "dialog.message" etc.


sjd78:
Before:
image


After:
image

ibolton336 (approved):

PR #1384, opened 2023-09-21
:seedling: Add `hacks/` to keep sample data and scripts to help developers

sjd78:
Modeled after tackle2-hub's `hacks/` folder, add a `hacks/` folder where sample import files (and in future helpful scripts) can be centrally held.  The goal is to have a common place to hold things that are useful for developers.


ibolton336 (approved):

mturley (approved):
Great idea

PR #1386, opened 2023-09-21
:bug: Update query when assessment is deleted

### LINKED ISSUE #1383: Retake assessment behavior polish
Currently, the retake functionality is not showing the existing answers for an assessment previously taken
Also, when an assessment is taken then the cancel button is pressed, the assessment needs to be deleted on the api level if no questions have been answered.
The retake/continue logic fails in this case where the assessment exists but isn't yet "started" because no answers have been filled out yet. So we are showing retake for these empty assessments when they should be deleted on wizard close.

ibolton336:
- Empty assessments should be automatically discarded when closing the wizard as to reduce confusion when computing status for assessment.
- Partially resolves #1383


sjd78 (approved):
LGTM

PR #1387, opened 2023-09-21
:bug: Fully unique ids for answer selections

ibolton336:
- Avoids bugs when similar questions are entered in the form with the same text causing the wrong answer to be selected. Adds unique IDs for selectable answers and questions.


sjd78 (approved):
LGTM

PR #1388, opened 2023-09-21
:bug: Don't allow archived questionnaires to be assessed

ibolton336:
- Restrict archived questionnaires to read-only


sjd78 (approved):
LGTM

PR #1392, opened 2023-09-22
:sparkles: Add useLocalTableControlsWithUrlParams hook and update archetypes table to use it

mturley:
When using the table-control hooks for a server-paginated table, you can use either `useTableControlState` or `useTableControlUrlParams` to provide state, then you take the returned object from that and use its properties to fetch API data, then pass the object with additional API-dependent properties to `useTableControlProps`.

For a client-paginated table, there is no need to split this into two steps because the API request is not dependent on the table state (there's no need to take pagination/sort/filter state and include it in the API request), so the shorthand `useLocalTableControls` hook exists which is simply `useTableControlProps(useLocalTableControlState(args))` (`useLocalTableControlState` is similar to `useTableControlState` but performs client-side logic with the `getLocal[Feature]DerivedState` helpers in between calling each feature hook). However, `useLocalTableControls` only allows the use of React state as the source of truth rather than URL params. This PR adds an equivalent `useLocalTableControlsWithUrlParams` shorthand hook that is a drop-in replacement for `useLocalTableControls` which uses URL params instead of React state.

The only additional argument available when switching to this new hook is `urlParamKeyPrefix`, which is optional and is used to distinguish the params for multiple tables on the same page. Even though the archetypes table is the only table on its page, I used it to be consistent in case we do add additional tables in modals or drawers, etc. like we did for Issues / Affected apps.

With the archetypes table now using URL params for its filter state, we should now easily be able to navigate to the archetypes table filtered by tags (as discussed with @ibolton336) by using an approach similar to `getAffectedAppsUrl` on the issues page ([usage here](https://github.com/konveyor/tackle2-ui/blob/main/client/src/app/pages/issues/affected-apps-link.tsx#L22C1-L28), [implementation here](https://github.com/konveyor/tackle2-ui/blob/main/client/src/app/pages/issues/helpers.ts#L100)), which uses the `trimAndStringifyUrlParams` and `serializeFilterUrlParams` helpers. This implementation might look something like:

```ts
export const getArchetypesUrlFilteredByTags = (tags: string[]) => {
  const baseUrl = Paths.archetypes;
  return `${baseUrl}?${trimAndStringifyUrlParams({
    newPrefixedSerializedParams: {
      [`${TableURLParamKeyPrefix.archetypes}:filters`]: serializeFilterUrlParams({ tags }).filters,
    },
  })}`;
};
```

Note (see the code comment in this diff): Because of the way we implemented `useUrlParams`, there is no way to conditionally use URL params or React state within the same hook based on an argument. This is why all the separate `use[Feature]State` and `use[Feature]UrlParams` hooks exist, and it also results in a moderate amount of duplication of code in this PR because of the logic required to feed the values returned from these feature hooks into each other. In the future we should refactor `useUrlParams` into something like `useStateOrUrlParams` to eliminate this issue and the duplication it causes (I also mentioned this in the initial docs in #1355). That will come in a future PR.


ibolton336 (approved):

client/src/app/hooks/table-controls/useLocalTableControlState.ts
```diff
@@ -77,3 +87,75 @@ export const useLocalTableControlState = <
...
+  });
+
+  const selectionState = useSelectionState({
+    items: filteredItems,
+    isEqual: (a, b) => a[idProperty] === b[idProperty],
+    initialSelected,
+    isItemSelectable,
+  });
+
+  const sortState = useSortUrlParams({ ...args, sortableColumns, initialSort });
```
sjd78, line 130: You _could_ use a bit of trickery to get this as a single call .....

```jsx
  const isUrlParamBased = !!urlParamKeyPrefix;

  const sortState = (isUrlParamBased ? useSortUrlParams : useSortState)({
    ...args, 
    sortableColumns,
    initialSort,
  });
```

sjd78 (commented):
Looks pretty good to me.  I just have a suggestion to help with the code duplication....

gitdallas (approved):

client/src/app/hooks/table-controls/useLocalTableControlState.ts
mturley: I considered this, but the eslint rules-of-hooks rule doesn't like it I think (if you re-render with a different value for `isUrlParamBased` it would switch up which hooks are called and screw up the render, not that we'd do that). If we added an eslint-ignore there I think it would be too easy to accidentally break it in a future change.

What I'd like to do eventually (soon perhaps) is refactor `useUrlParams` (which is used inside `useSortUrlParams` and each of the other `use[Feature]UrlParams` hooks) so that it itself has a `isUrlParamBased` condition, and has its own `useState` internally that it either uses or ignores in favor of URL params depending on config. Then we could get rid of all the `use[Feature]UrlParams` hooks and instead have the URL params be an option in `use[Feature]State`, all the way up to `useTableControlState`.

PR #1399, opened 2023-09-25
:bug: Don't allow an archetype to be assessed if empty

ibolton336:
Closes https://issues.redhat.com/browse/MTA-1314


sjd78 (commented):
The code change is fine, but I don't understand why the issue was raised. Nothing in the enhancement doc prevents reviewing or assessing an empty archetype.

PR #1402, opened 2023-09-26
:ghost: Remove shared app filter and unused menuActions

ibolton336:
- Removes some old code I noticed after the dropdown update & analysis table PRs were merged
- Menu actions are no longer used in the page header component and have been moved to the toolbar
- Removes old shared app filter hook that is no longer in use


ibolton336:
> i noticed menuActions was only given empty arrays as well (had to put some fake data in to test it locally), but left it in assuming that there may be a need for it coming up. if not, lgtm.

Appreciate it!

sjd78 (approved):
LGTM

Issue #1321, opened 2023-08-29
Add category grouping option for autocomplete tag dropdown implementation 


ibolton336:
https://github.com/konveyor/tackle2-ui/pull/1295#issuecomment-1697653629 

Address the new grouping option for tag categories. We want to be able to see associated categories that tags belong to in the dropdown as shown in the mockups.

sjd78:
@rszwajko - Did one of your recent changes cover this issue already?

rszwajko:
The grouping feature for autocomplete has been added in #1799 but it's not enabled yet. However it should look the same as for the stakeholders here -> https://github.com/konveyor/tackle2-ui/blob/56e556f23d54b7ecb656e1345e63ddf65695c757/client/src/app/pages/assessment/components/assessment-stakeholders-form/assessment-stakeholders-form.tsx#L47

dymurray:
Closing due to inactivity or age, please reopen if you feel this was done in error

PR #1403, opened 2023-09-26
:bug: fix(autocomplete): tags not generated for preselected options

gitdallas:
closes jira 1315


ibolton336:
Just noticed that the app can have multiple tags with the same name from different sources. This is causing duplicate tags to render in the UI 
Screenshot 2023-09-26 at 4 45 07 PM




Screenshot 2023-09-26 at 4 45 52 PM


Wondering if this enhancement would fall under https://github.com/konveyor/tackle2-ui/issues/1321 or if we would need to be able to display tag by category AND source in this dropdown.

gitdallas:
@ibolton336 - how would this work in practice? should there be tags with the same id/name and a different source? how would the labels/tags look in this case?

ibolton336:
> @ibolton336 - how would this work in practice? should there be tags with the same id/name and a different source? how would the labels/tags look in this case?

@justinxhale

ibolton336:
@gitdallas I am thinking that any tag that is not of source type "" (empty string aka "manual") should be hidden from this view. Maybe there is a read only field that might display tags associated to an app by archetype / assessment / analysis. IIRC this screen only lets you select "manual" source tags so in turn I think we should only be displaying "manual" source tags.

gitdallas:
@ibolton336 - the `tags` value in `application-form` doesn't seem to have a source type:

image

ibolton336:
Sounds good @gitdallas - Looks like we at least have a source optional property on the TagRef so at least we are covered there.

client/src/app/pages/applications/components/application-form/application-form.tsx
```diff
@@ -445,78 +447,36 @@ export const ApplicationForm: React.FC<ApplicationFormProps> = ({
...
+            renderInput={({ field: { value, name, onChange } }) => {
+              const selections = value
+                .map(
+                  (formTag) =>
+                    tags?.find((tagRef) => tagRef.name === formTag.name)
+                )
+                .map((matchingTag) =>
+                  matchingTag ? matchingTag.name : undefined
+                )
+                .filter((e) => e !== undefined) as string[];
```
ibolton336, line 459:
```suggestion
              const tagNames = new Set(tags?.map((tag) => tag.name));

              const selections = value.flatMap((formTag) =>
                formTag.source === "" && tagNames.has(formTag.name)
                  ? [formTag.name]
                  : []
              );

```

ibolton336: Played around with this a bit and borrowed Mikes suggestion of using a Set/has here instead of creating a map and using includes since set narrows the tags to unique values & boosts performance. Explicitly checking for the empty string on the form tag on lookup should give us what we want since empty string here means "manually selected tag" for a tag on an application object (a formTag in this case) .

gitdallas: I ended up making `tagOptions` a set of strings (tag names) from the beginning and using that. Also as the onChange event maps the form values to `tags` (which seems to have undefined source instead of ""), I merge a `{source:""}` object when returning from `getTagRef`. Without this, after making any changes to the value we'd filter them out if they didn't have a source prop in the `tags` array.

ibolton336: Noticing that onSubmit, any tags that are filtered out from the selections on line 448 are not being added back in to the payload.

ibolton336: Sam confirmed that there is no way for us to overwrite those non-manual tags by updating the application object directly so we should be good on what we are passing to the put request here. There seems to be some ongoing discussion/weirdness around this since the non-manual tags are "virtual" tags being inherited from other membership criteria etc. 



We do need to make sure we are filtering out any pre-existing, non-manual tags in the selectable tag list so that we can't accidentally create a new "manual" tag for a tag that already exists on the application from a different source.

gitdallas: I'm unsure if there is a change you're asking for or if you are saying to wait.

Right now it is using `tags` (filtered to only include undefined or "" source value) as the available options and not letting the user add new tags that don't exist outside the options.

ibolton336: Understood - My concern was that after we are filtering out the tags on an app with a source other than "", we never add them back in to the tags payload on submit so we'd be overwriting those tags.

ibolton336: >Right now it is using tags (filtered to only include undefined or "" source value) as the available options and not letting the user add new tags that don't exist outside the options.

The change I think we might temporarily need is to check to see if any tags in the "tagOptions" list you've created match up with any other tags on the application (archetype source or other non-manual tags) and remove any matching available tag from the list. Modifying the tags api fetch response won't include any tags with other sources . Those sources are only on tags that have been associated with an application or archetype. We would have to cross check the response from the tags fetch api with what exists on the application in this case.

ibolton336 (approved):

sjd78:
> Sounds good @gitdallas - Looks like we have a "source" optional property on the TagRef so at least we are covered there. The Tags from the tags api won't have a source since they aren't associated with an app/archetype yet.

I'm looking at this PR today while rebasing #1408 on this change.

A few things are quite confusing.

First, the list of tags generated from the tagCategories are just `Tag` and not `TagRef`.  The way the `tags` array is being kept in state via an effect isn't the best approach.  The way `tagOptions` set is generated is confusing, and breaks if you have the `tags` just be a `Tag[]`.  It's much easier to do this to get a full unique list of tags:
```jsx
const tags = (tagCategories || []).flatMap((tc) => tc.tags || []);
const tagOptions = new Set(tags.map((t) => t.name));
```

Second, the `Application.tags` field is, in fact, a `TagRef[]`.  This means that we need to do something to convert the selected tag names to a `TagRef[]` for the rest api payload object.  Right now looks like the Autocomplete onChange handler is doing the transforms.  Maybe moving to use `ItemsSelect` may be helpful (client/src/app/components/items-select/items-select.tsx).  Assuming the update api endpoint will just deal with updating `source: ""` fields, that should be all we need to send in the payload...

Any further filtering of the set of tags to display etc should probably just be done in the same general area as where `tags` and `tagOptions` are handled...
```jsx
  const applicationManualTags =
    new Set(application?.tags?.filter((t) => t.source === "").map((t) => t.name) ?? []);

  const applicationSourcedTags =
    new Set(application?.tags?.filter((t) => t.source !== "").map((t) => t.name) ?? []);

  const manualTagOptions = new Set(
    tags.map((t) => t.name).filter((name) => !applicationSourcedTags.has(name))
  );
```

Then drive the select box with `manualTagOptions` and adjust the payload generation accordingly.

ibolton336:
Agreed that this could use another pass @sjd78. This issue was blocking QE so wanted to get something in for them so that they can fix their automated tests. You raise some good points.

PR #1406, opened 2023-09-27
:bug: Add a fallback image if the image path is broken

ibolton336:
- Closes https://issues.redhat.com/browse/MTA-1249


sjd78:
Looks like this should be closed in favor of #1420

PR #1409, opened 2023-09-27
:bug: Shows client side loading state for jira tracker

ibolton336:
Closes [False 'not connected' status for new Jira instance](https://issues.redhat.com/browse/MTA-1019)


client/src/app/pages/external/jira/trackers.tsx
```diff
@@ -259,11 +266,15 @@ export const JiraTrackers: React.FC = () => {
                           width={10}
                           {...getTdProps({ columnKey: "connection" })}
                         >
-                          <TrackerStatus
-                            name={tracker.name}
-                            connected={tracker.connected}
-                            message={tracker.message}
-                          />
+                          {updatingTrackerId === tracker.id ? (
```
sjd78, line 269: This will be "the most recently updated/created tracker".  Might want to track an array of tracker ids just to be sure that you get a spinner on multiple trackers if they're both edited very close together.  

Maybe a Set would work better?  But then since a Set is mutable, getting `useState()` to track changes would be a bit of a challenge.  Some stackoverflow discussion: https://stackoverflow.com/questions/58806883/how-to-use-set-with-reacts-usestate

client/src/app/pages/external/jira/tracker-form.tsx
```diff
@@ -57,49 +57,70 @@ interface FormValues {
...
+    setUpdatingTrackerId(_.data.id);
 
-  const { mutate: createTracker } = useCreateTrackerMutation(
-    onCreateTrackerSuccess,
-    onCreateUpdatetrackerError
-  );
+    setTimeout(() => {
+      setUpdatingTrackerId(null);
+    }, 5000);
+  };
```
sjd78, line 94: Is this still needed with how the hook is structured?

client/src/app/pages/external/jira/tracker-form.tsx
```diff
@@ -57,49 +57,70 @@ interface FormValues {
...
         type: t("terms.instance"),
       }),
       variant: "success",
     });
 
+    setUpdatingTrackerId(tracker.id);
+
+    setTimeout(() => {
+      setUpdatingTrackerId(null);
+    }, 5000);
```
sjd78, line 111: Is this still needed with how the hook is structured?

client/src/app/pages/external/jira/tracker-form.tsx
```diff
@@ -57,49 +57,70 @@ interface FormValues {
...
+  };
+
+  const onCreateUpdatetrackerError = (error: AxiosError) => {
+    setAxiosError(error);
+    setUpdatingTrackerId(null);
```
sjd78, line 116: I'm not sure this is needed.  If there is an error, the success callback won't set the tracker id.

sjd78 (changes requested):
I don't mind the artificial spinner after kicking off the create or update.

The hook should be able to cancel the timers that are set.  So maybe change the `setUpdatingTrackerId()` to `addUpdatingTrackerId()` and adjust a few calls?

client/src/app/pages/external/jira/trackers.tsx
ibolton336: Converted to use a set internal to the hook but exported the readonly state as an array. Interesting on the mutable workaround. Updated the hook to use the approach you linked.

client/src/app/pages/external/jira/useUpdatingTrackerIds.ts
```diff
@@ -0,0 +1,33 @@
...
+      timerId = window.setTimeout(() => {
+        setUpdatingTrackerIds(new Set());
+      }, delay);
+    }
+
+    return () => {
+      if (timerId !== null) {
+        window.clearTimeout(timerId);
+      }
+    };
```
sjd78, line 23: Does this work if you updated 2 trackers within the 10 second interval?

The `addUpdatingTrackerId` function looks right, but the removal of the ids seems off since I don't see something like `new Set(prevSet.remove(expiredId))`.

sjd78 (commented):
I haven't tested this yet, but I'm not sure updating 2 trackers will show 2 spinners.

client/src/app/pages/external/jira/useUpdatingTrackerIds.ts
ibolton336: It is actually working because on refetch this hook is remounted / cleaned up. Only works because we are routinely polling for trackers.

ibolton336: Ah I see what you mean. The entire set is wiped when one timeout expires. Fixing.

client/src/app/pages/external/jira/trackers.tsx
```diff
@@ -263,6 +266,12 @@ export const JiraTrackers: React.FC = () => {
                             name={tracker.name}
                             connected={tracker.connected}
                             message={tracker.message}
+                            // isTrackerUpdating={updatingTrackerIds.has(
+                            //   tracker.id
+                            // )}
```
sjd78, line 271: Can remove this comment!

sjd78 (approved):
The hook is a bit confusing but looks like it'll be ok.  Everything else LGTM.

PR #1416, opened 2023-09-28
:seedling: Setup TS to know about `.filter(Boolean)`

sjd78:
With `array-filter-Boolean.ts`, TS will select a better Array.filter() override so this kind of thing will work automatically without an explicit type casting:

```js
const a: Array<string | undefined | null> = ["A", "B", undefined, null, "C", ""];

const b: string[] = a.filter(Boolean);
// b = ["A", "B", "C"]
```

See https://www.karltarvas.com/typescript-array-filter-boolean.html
See https://stackoverflow.com/a/51390763/1470607


ibolton336 (approved):
Nice find!

PR #1418, opened 2023-09-29
:bug: Business services: Fix create/edit when owner is included

sjd78:
The `owner` field on the `BusinessService` payload needs to be a pure `Ref` object or it will be rejected by the REST API call.  Adopt the same set of data transforms used in the application-form to handle getting the correct set of data.

Related changes:
  - Business services related REST API functions updated to have the correct response types

  - Business services queries updated to pass REST API response and input values to `onSuccess()` and `onError()` handlers

  - `BusinessServiceForm` updated to use mutation response data to display the name of the business service in success messages

  - Refactored `business-service-form.tsx` to move all data access/mutation code to hook `useApplicationFormData() to logically divide concerns (data access v. UI handling)

Resolves: https://issues.redhat.com/browse/MTA-1346

### DIFF
--- a/client/src/app/api/rest.ts
+++ b/client/src/app/api/rest.ts
@@ -618,27 +618,27 @@ export const updateStakeholderGroup = (
 ): Promise<StakeholderGroup> =>
   axios.put(`${STAKEHOLDER_GROUPS}/${obj.id}`, obj);
 
+// ---------------------------------------
 // Business services
+//
+export const getBusinessServices = () =>
+  axios
+    .get<BusinessService[]>(BUSINESS_SERVICES)
+    .then((response) => response.data);
 
-export const getBusinessServices = (): Promise<BusinessService[]> =>
-  axios.get(BUSINESS_SERVICES).then((response) => response.data);
+export const getBusinessServiceById = (id: number | string) =>
+  axios
+    .get<BusinessService>(`${BUSINESS_SERVICES}/${id}`)
+    .then((response) => response.data);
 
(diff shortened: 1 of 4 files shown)


ibolton336 (approved):

client/src/app/api/rest.ts
```diff
@@ -618,27 +618,27 @@ export const updateStakeholderGroup = (
...
-  obj: New<BusinessService>
-): Promise<BusinessService> => axios.post(BUSINESS_SERVICES, obj);
+export const createBusinessService = (obj: New<BusinessService>) =>
+  axios.post<BusinessService>(BUSINESS_SERVICES, obj);
 
-export const updateBusinessService = (
-  obj: BusinessService
-): Promise<BusinessService> => axios.put(`${BUSINESS_SERVICES}/${obj.id}`, obj);
+export const updateBusinessService = (obj: BusinessService) =>
+  axios.put<void>(`${BUSINESS_SERVICES}/${obj.id}`, obj);
```
ibolton336, line 638: I like the void callout here. Not sure why we were pretending to get something useful back before.

ibolton336:
Tested and LGTM!

client/src/app/api/rest.ts
sjd78: I'll be revisiting all of the rest functions soon in a similar way.

ibolton336:
https://issues.redhat.com/browse/MTA-1375?filter=-1

PR #1420, opened 2023-09-29
:bug: Fetch target images in auth envs

ibolton336:
Closes https://issues.redhat.com/browse/MTA-1249?filter=-1

- Image was not being fetched in envs with auth enabled. Image tags () in HTML do not follow the same request mechanism as AJAX or fetch requests. They do not use the XMLHttpRequest or axios API, so they were not intercepted by axios middleware interceptor we have set up for API requests.


client/src/app/components/TargetCard.tsx
```diff
@@ -68,6 +69,8 @@ export const TargetCard: React.FC<TargetCardProps> = ({
   const { t } = useTranslation();
   const [isCardSelected, setCardSelected] = React.useState(cardSelected);
 
+  const [imageDataUrl, setImageDataUrl] = React.useState<string | null>(null);
```
sjd78, line 72: It's not really a url anymore, it's just the image data itself, so.... s/imageDataUrl/imageData/

client/src/app/components/TargetCard.tsx
```diff
@@ -105,23 +108,23 @@ export const TargetCard: React.FC<TargetCardProps> = ({
...
-      );
+  const fetchImageAndSetDataUrl = async (imagePath: string) => {
+    try {
+      const imageDataResponse = await axios.get(imagePath, {
+        headers: {
+          Accept: "application/octet-stream",
+        },
+      });
+      const encodedSvg = encodeURIComponent(imageDataResponse.data);
+      setImageDataUrl(`data:image/svg+xml,${encodedSvg}`);
```
sjd78, line 119: This doesn't work when I create a custom target where the image is a png.

Maybe look at the response's content type (if that is correct at least)?

```js
setImageDataUrl(`data:${imageDataResponse.headers["content-type"]},${encodedSvg}`);
```

client/src/app/components/TargetCard.tsx
```diff
@@ -105,23 +108,23 @@ export const TargetCard: React.FC<TargetCardProps> = ({
...
+    } catch (error) {
+      console.error("There was an issue fetching the image:", error);
     }
-
-    return result;
   };
+  const imagePath = target?.image?.id
+    ? `/hub/files/${target?.image.id}`
+    : DefaultRulesetIcon;
+  fetchImageAndSetDataUrl(imagePath);
```
sjd78, line 127: So I think this can be swapped out for a `useEffect()` so it doesn't run too often...

```jsx
  useEffect(() => {
    const imagePath = target?.image?.id
      ? `/hub/files/${target?.image.id}`
      : DefaultRulesetIcon;

    // sneaky self executing async function since useEffect() needs a sync function
    (async () => {
      try {
        const imageDataResponse = await axios.get(imagePath, {
          headers: {
            Accept: "application/octet-stream",
          },
        });
        const encodedSvg = encodeURIComponent(imageDataResponse.data);
        setImageDataUrl(`data:${imageDataResponse.headers["content-type"]},${encodedSvg}`);
      } catch (error) {
        console.error("There was an issue fetching the image:", error);
      }
    })();
  }, [target, setImageDataUrl]);
```

sjd78 (changes requested):
Great catch to see the 302 response is the request missing auth headers!

A few suggestions to tighten things up.

The biggest issue is getting the content correct in the data url.  Try creating your own custom target with a png or jpg image upload..

client/src/app/components/TargetCard.tsx
```diff
@@ -105,23 +108,23 @@ export const TargetCard: React.FC<TargetCardProps> = ({
...
-        />
-      );
+  const fetchImageAndSetDataUrl = async (imagePath: string) => {
+    try {
+      const imageDataResponse = await axios.get(imagePath, {
+        headers: {
+          Accept: "application/octet-stream",
+        },
+      });
+      const encodedSvg = encodeURIComponent(imageDataResponse.data);
```
sjd78, line 118: This will garble regular image files.  Maybe something like this: https://stackoverflow.com/a/44058739

client/src/app/components/TargetCard.tsx
sjd78: Ok, so reading it again today (with less blurry eyes and more coffee) and it really is a data-url.  So the name does make sense!

ibolton336: +1 Updated

sjd78 (approved):
Doing all the work in a hook is a great idea!

I tested with custom targets that have png and jpg icons.  It all looks good!

image

PR #1421, opened 2023-10-02
:seedling: fix childpaths to work with nested nodemodules

gitdallas:
this came up while testing prerelease pf versions

@sjd78 came up with these changes to be more flexible with where stylepaths can be


sjd78 (approved):
This change does leave the possibility that multiple versions of a file could be included in the bundled output.  In reality that should only be a problem if using versions like `^5.2.27-prerelease.57` since that won't semver match a `^5.2.1` dependency version in another module.

To see what files are included in a build, we can check the source maps with `jq`.

This will show all of the css files and what bundles they are packaged in:
```sh
> npm run build
> cd client/dist
> jq '{ file, sources }' *.css.map css/*.css.map
```

This will show all of the css files across all chunks that come from a dependency:
```sh
> npm run build
> cd client/dist
> jq -r '.sources[]' *.css.map css/*.css.map | sort | uniq | grep 'node_modules'
```

From either of those lists (but probably easier from the second) you can spot if the same file appears in the bundles but from different paths.  If that happens, a bit of build debugging and dependency version tuning is recommended!

PR #1425, opened 2023-10-02
:ghost: Convert sh & shg to refs on waves to match api

ibolton336:
Resolves:
  - [MTA-1343](https://issues.redhat.com/browse/MTA-1343)
  - [MTA-1337](https://issues.redhat.com/browse/MTA-1337)
  - [MTA-1281](https://issues.redhat.com/browse/MTA-1281)


client/src/app/pages/migration-waves/migration-waves.tsx
```diff
@@ -160,10 +161,20 @@ export const MigrationWaves: React.FC = () => {
...
-    updateMigrationWave(migrationWave);
+    const payload: MigrationWave = {
+      id: migrationWave.id,
+      name: migrationWave.name,
+      startDate: migrationWave.startDate,
+      endDate: migrationWave.endDate,
+      stakeholderGroups: migrationWave.stakeholderGroups,
+      stakeholders: migrationWave.stakeholders,
+      applications: applicationRefs || [],
+    };
```
sjd78, line 175: If most things are the same, then spreading the object and only specifying the keys to be overridden is a bit more compact:
```js
    const payload: MigrationWave = {
      ...migrationWave,
      applications: applicationRefs || [],
    };
```

or

```js
  updateMigrationWave({ ...payload, applications: applicationRefs || [] });
```

sjd78 (changes requested):
It looks good and works, but a few style change suggestions to reduce some complexity.

client/src/app/pages/migration-waves/migration-waves.tsx
ibolton336: I tried this, but there are some things included inthe migrationWave we have stored from the api which break the put request when submitting. So I explicitly defined what was passed here.

sjd78: So weird.  I wonder how many more json parse type errors we're going to hit this bug round.

sjd78 (approved):
LGTM

PR #1428, opened 2023-10-03
:bug: fix issue with select menu getting cut by edge of modal

gitdallas:
image

becomes

image


client/src/app/components/SimpleSelect.tsx
```diff
@@ -40,6 +40,7 @@ export const SimpleSelect: React.FC<ISimpleSelectProps> = ({
 
   return (
     <Select
+      menuAppendTo="parent" // prevent menu from being clipped by modal edges
```
ibolton336, line 43: Adding this overflow seems to be a much better experience, but not sure if it is the PF friendly way to accomplish this. 
```suggestion
      menuAppendTo="parent" // prevent menu from being clipped by modal edges
      style={{ overflowY: "auto", maxHeight: "200px" }}

```

ibolton336 (changes requested):

client/src/app/components/SimpleSelect.tsx
ibolton336: Screenshot 2023-10-03 at 7 54 02 PM

gitdallas: i added a maxHeight of 200

gitdallas:
@ibolton336 overflow is automatic

client/src/app/components/SimpleSelect.tsx
ibolton336: I see the overflow now with the maxHeight set. LGTM other than the console.log

ibolton336 (changes requested):

client/src/app/components/SimpleSelect.tsx
```diff
@@ -37,9 +37,11 @@ export const SimpleSelect: React.FC<ISimpleSelectProps> = ({
   ...props
 }) => {
   const [isOpen, setIsOpen] = useState(false);
-
+  console.log({ ...props });
```
ibolton336, line 40: This guy snuck in. Otherwise LGTM.

ibolton336 (approved):

PR #1429, opened 2023-10-03
:ghost: Adds missing dependency link to app drawer

ibolton336:
Screenshot 2023-10-03 at 7 41 14 PM
 - Adds a dependency link to the application detail drawer

### DIFF
--- a/client/src/app/pages/applications/components/application-detail-drawer/application-detail-drawer.tsx
+++ b/client/src/app/pages/applications/components/application-detail-drawer/application-detail-drawer.tsx
@@ -11,7 +11,8 @@ import {
   TabTitleText,
   Spinner,
   Bullseye,
-  Button,
+  List,
+  ListItem,
 } from "@patternfly/react-core";
 import spacing from "@patternfly/react-styles/css/utilities/Spacing/spacing";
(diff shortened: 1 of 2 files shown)


client/src/app/pages/applications/components/application-detail-drawer/application-detail-drawer.tsx
```diff
@@ -84,10 +86,21 @@ export const ApplicationDetailDrawer: React.FC<
...
               </Link>
             ) : null}
+            {application ? (
+              <Link
+                to={getDependenciesUrlFilteredByAppName(application?.name)}
+                style={{ display: "block" }}
+              >
+                Dependencies
+              </Link>
+            ) : null}
```
sjd78, line 103: I'd suggest to use a [plain List](https://www.patternfly.org/components/list#plain) here instead of a `style={{ display: "block" }}` to align vertically.

You could put the `List` with both links within the same `{application ? ...List... : null}` block.

sjd78 (changes requested):
The compose URL is good and works well.

The Link rendering does work, but I'd prefer a different style for laying out the components.

PR #1438, opened 2023-10-05
:bug: Update archetype model, form and detail drawer to current hub spec

sjd78:
### Summary

Resolves: #1436

With hub updates to tag handling and api field name changes related to archetypes, adjustments needed to be made to the UI.

### Archetype model
  - Update `models.ts` to the current state of `TagRef` and `Archetype` in terms of field names and data types.

  - MSW stub data updated to use api model changes.

### Archetype Table 
  - **Note**: The tags column on the archetype table is currently displaying ALL tags returned with no source differentiation.

### Archetype Add/EditForm
  - Update `ArchetypeForm` to use `criteria` instead of `criteriaTags`  as per the api model changes.

  - With hub changes to combine the archetype and non-archetype tags into a single field, the form processing now ignored non-archetype fields but still sends them on an update.

  - Note: This change mirrors the behavior of manual / non-manual tags on applications and the application form.

  - Stakeholder and stakeholder groups handling is updated to use the `*ToRef()` helper function style introduced in the application form change #1408.

### Archetype Detail Drawer

  - Since all tags for an archetype are provided in the same field `tags`, filter them into "archetype tags" (i.e. an empty source) and "assessment tags" (i.e. have a non-empty source) to display them separately.

  - Note: The "archetype tags" are the tags manually attached to the archetype in the new/edit form.

PR #1446, opened 2023-10-06
:bug: Application form handle manual tags w/o `source` attribute

sjd78:
Follow up: #1438

Api changed and if a tag's source is empty, the field is not sent. Change the tag filtering test to consider `source === undefined` and `source === ""` as manual tags.


client/src/app/pages/applications/components/application-form/application-form.tsx
```diff
@@ -103,9 +103,11 @@ export const ApplicationForm: React.FC<ApplicationFormProps> = ({
...
-  const manualTags = application?.tags?.filter((t) => t.source === "") ?? [];
+  const manualTags =
+    application?.tags?.filter((t) => t?.source ?? "" === "") ?? [];
 
-  const nonManualTags = application?.tags?.filter((t) => t.source !== "") ?? [];
+  const nonManualTags =
+    application?.tags?.filter((t) => t?.source ?? "" !== "") ?? [];
 
   // Allow all tags to be selected manually, even if they are included from another source
   const allowedManualTags = tags;
```
ibolton336, line 113:
```suggestion
  const allowedManualTags = tags.filter((tag) => !manualTags.includes(tag));
```

Thoughts on adding this to handle the dupe tag scenario discovered? Can totally be another PR. Just posting here before I forget.

ibolton336 (approved):

client/src/app/pages/applications/components/application-form/application-form.tsx
sjd78: Is the dupe tag scenario still a problem?

IIRC, all tags should be selectable as a manual tag even if the same tag shows up with two different sources.  Is there yet another dupe tag scenario?

ibolton336: Yeah - when importing an app with a bunch of tags already applied ( potentially duplicate tags), this was necessary. Open to revisiting this. I think the hub is adding additional duplication logic.

sjd78: So there can be tags on an application that don't exist as a tag?   That's a weird import behavior...

Issue #1244, opened 2023-08-01
Replace deprecated PF Wizard


gitdallas:
Will require refactor. Instances of `Wizard`, `WizardStepFunctionType`, `WizardStep`, `WizardContextConsumer`, `WizardFooter`

https://www.patternfly.org/components/wizard

PR #1473, opened 2023-10-12
:seedling: update to new wizard for assessment-wizard and custom-wizard-footer

gitdallas:
Would like some thorough testing on this.

#1244

### DIFF
--- a/client/src/app/pages/assessment/components/assessment-wizard/assessment-wizard.tsx
+++ b/client/src/app/pages/assessment/components/assessment-wizard/assessment-wizard.tsx
@@ -124,7 +124,7 @@ export const AssessmentWizard: React.FC<AssessmentWizardProps> = ({
       assessment.sections
         .flatMap((f) => f.questions)
         .forEach((question) => {
-          const existingAnswer = assessment.sections
+          const existingAnswer = assessment?.sections
             ?.flatMap((section) => section.questions)
             .find((q) => q.text === question.text)
             ?.answers.find((a) => a.selected === true);
@@ -552,6 +552,7 @@ export const AssessmentWizard: React.FC<AssessmentWizardProps> = ({
         }
       }
       setAssessmentToCancel(null);
+      assessment = undefined;
     }
   };
 
@@ -592,7 +593,7 @@ export const AssessmentWizard: React.FC<AssessmentWizardProps> = ({
 
   return (
     <>
-      {isOpen && (
+      {isOpen && assessment && (
         <FormProvider {...methods}>
           <h1>
             Questionnaire:{" "}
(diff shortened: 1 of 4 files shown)


ibolton336:
@gitdallas haha sorry about that! You can create a stakeholder in the "controls" section of the app.

ibolton336 (changes requested):

client/src/app/pages/assessment/components/assessment-wizard/assessment-wizard.tsx
```diff
@@ -506,62 +503,6 @@ export const AssessmentWizard: React.FC<AssessmentWizardProps> = ({
...
-      onSaveAsDraft={() => {
-        methods.setValue(SAVE_ACTION_KEY, SAVE_ACTION_VALUE.SAVE_AS_DRAFT);
-        methods.handleSubmit(onSubmit)();
-      }}
-    />
-  );
-
   useEffect(() => {
     const unlisten = history.listen((newLocation, action) => {
       if (action === "PUSH" && assessment) {
```
ibolton336, line 508: Looks like something with the way the new wizard works is causing this to trigger on next button click. Need to figure out how to clear the assessment when the user navigates away from the wizard without explicitly saving/cancelling the wizard so that we don't end up with an orphaned assessment.

ibolton336 (changes requested):

client/src/app/pages/assessment/components/assessment-wizard/assessment-wizard.tsx
```diff
@@ -610,22 +551,81 @@ export const AssessmentWizard: React.FC<AssessmentWizardProps> = ({
...
+          methods.handleSubmit(onSubmit)();
+        }}
+      />
+    );
+  };
+
   return (
     <>
-      {isOpen && (
+      {isOpen && !!sortedSections.length && (
```
ibolton336, line 591: I don't think we still need this change after updating the sortedSections constant. We should probably add some loading state now that the initial values are async. Maybe we could add a conditional render wrapper within this component.

gitdallas: This is actually needed in combination with the update. The update sets sortedSections to [] when loading so that this will clear the Wizard render.

ibolton336: Ah I see - noticed that when i remove the useMemo from sortedSections, I no longer need this. Seems like the memoization there is giving us more trouble than it is worth & probably not saving too many re-renders since the assessment prop is cached and managed by react-query. Take it or leave it!

ibolton336:
Looking good - just a few small comments re: leftover function & useMemo usage.

ibolton336 (approved):

mturley (approved):
Haven't tested but code LGTM

PR #1470, opened 2023-10-12
:bug: Add lists of associated apps & archetypes

ibolton336:
https://issues.redhat.com/browse/MTA-1384?filter=12422675

PR #1481, opened 2023-10-16
:ghost: Adds colors to drawer tags & removes boilerplate

ibolton336:
Follow up to: https://github.com/konveyor/tackle2-ui/pull/1470 

- Added a reusable component for drawer labels. 
- Adds random colors to the tags just for differentiation between the tags when viewing in the drawer.

### DIFF
--- a/client/src/app/components/labels-from-items/labels-from-items.tsx
+++ b/client/src/app/components/labels-from-items/labels-from-items.tsx
@@ -1,5 +1,5 @@
 import React from "react";
-import { LabelProps } from "@patternfly/react-core";
+import { LabelGroup, LabelProps } from "@patternfly/react-core";
 import { LabelCustomColor } from "../LabelCustomColor";
 import { useTranslation } from "react-i18next";
 
@@ -8,14 +8,20 @@ interface RandomColorLabelProps extends LabelProps {}
 export const RandomColorLabel: React.FC<RandomColorLabelProps> = ({
   ...props
 }) => {
-  const getRandomColor = () => {
-    const letters = "0123456789ABCDEF";
-    let color = "#";
-    for (let i = 0; i < 6; i++) {
-      color += letters[Math.floor(Math.random() * 16)];
-    }
-    return color;
-  };
+  const getRandomColor = (() => {
+    "use strict";
+
+    const randomInt = (min: number, max: number) => {
+      return Math.floor(Math.random() * (max - min + 1)) + min;
+    };
+
+    return () => {
+      const h = randomInt(0, 360);
+      const s = randomInt(42, 98);
+      const l = randomInt(40, 90);
+      return `hsl(${h},${s}%,${l}%)`;
+    };
+  })();
 
   const randomColor = getRandomColor();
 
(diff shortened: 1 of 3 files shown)


client/src/app/components/labels-from-items/labels-from-items.tsx
```diff
@@ -0,0 +1,45 @@
...
+  ...props
+}) => {
+  const getRandomColor = () => {
+    const letters = "0123456789ABCDEF";
+    let color = "#";
+    for (let i = 0; i < 6; i++) {
+      color += letters[Math.floor(Math.random() * 16)];
+    }
+    return color;
+  };
```
sjd78, line 18: There is an alternative that looks like it can generate some better colors (like keeping super dark and super light colors out):  https://gist.github.com/bendc/76c48ce53299e6078a76

client/src/app/pages/applications/components/application-detail-drawer/application-detail-drawer-assessment.tsx
```diff
@@ -164,17 +164,7 @@ export const ApplicationDetailDrawerAssessment: React.FC<
...
-      {(archetypeRefs as Ref[])
-        .sort((a, b) => a.name.localeCompare(b.name))
-        .map((ref) => (
-          <Label color="grey" key={ref.id}>
-            {ref.name}
-          </Label>
-        ))}
-    </LabelGroup>
-  );
+}) => <LabelsFromItems items={archetypeRefs as Ref[]} />;
```
sjd78, line 170: Is the `as Ref[]` required?

client/src/app/components/labels-from-items/labels-from-items.tsx
```diff
@@ -0,0 +1,45 @@
...
+    return <div>{noneMessage || t("terms.none")}</div>;
+  }
+
+  return (
+    <div>
+      {items.map((item, index) => (
+        <RandomColorLabel key={index}>{item.name}</RandomColorLabel>
+      ))}
+    </div>
+  );
```
sjd78, line 44: Looks like previously the Labels were wrapped in a `LabelGroup` instead of the `<div>`.

client/src/app/components/labels-from-items/labels-from-items.tsx
```diff
@@ -0,0 +1,45 @@
...
+  const randomColor = getRandomColor();
+
+  return <LabelCustomColor color={randomColor} {...props} />;
+};
+
+export function LabelsFromItems<T extends { name: string }>({
+  items,
+  noneMessage,
+}: {
+  items: T[];
```
sjd78, line 29: If this prop is made optional, the "nonMessage" test can be adjusted, and consumers of the component can also have the source of the items be optional.

sjd78 (changes requested):
I like the structure of the change.  Only a few comments.

client/src/app/components/labels-from-items/labels-from-items.tsx
```diff
@@ -0,0 +1,51 @@
...
+export function LabelsFromItems<T extends { name: string }>({
+  items,
+  noneMessage,
+}: {
+  items?: T[];
+  noneMessage?: string;
+}): JSX.Element {
+  const { t } = useTranslation();
+
+  if (items?.length === 0) {
```
sjd78, line 40: If `items` is undefined, this will reduce to `undefined === 0` which is `false`.

Try `if (items?.length ?? 0 === 0) {` to force a 0 if `items` is undefined.

client/src/app/pages/archetypes/components/archetype-detail-drawer.tsx
```diff
@@ -170,56 +169,18 @@ const ArchetypeDetailDrawer: React.FC<IArchetypeDetailDrawerProps> = ({
...
-        .sort((a, b) => a.name.localeCompare(b.name))
-        .map((sh) => (
-          <Label color="grey" key={sh.id}>
-            {sh.name}
-          </Label>
-        ))}
-    </LabelGroup>
-  );
+const TagLabels: React.FC<{ tags?: Tag[] }> = ({ tags }) => (
+  <LabelsFromItems items={tags as Tag[]} />
```
sjd78, line 175: nit-pick: Should not need the type cast since `items` is set to be optional

sjd78 (commented):
Looks great, just one last change and one nit-pick

sjd78 (approved):

PR #1485, opened 2023-10-20
:bug: Don't auto select source labels when setting a target

ibolton336:
- Only include explicitly selected source labels from a selected target


galt-tr (approved):
VISIACK

dymurray (approved):

PR #1486, opened 2023-10-23
:bug: Convert assessment wizard to modal

ibolton336:
- https://issues.redhat.com/browse/MTA-1437 
- https://issues.redhat.com/browse/MTA-1467
Resolves loading state for slow connections & moves the assessment wizard to a modal to simplify code & improve UX. This was approved by Justin from UXD.


mturley (commented):
@ibolton336 this mostly LGTM, but with the wizard in a modal it should have a `WizardHeader` at the top passed in via the `header` prop of Wizard ([example here](https://www.patternfly.org/components/wizard#within-modal))

gitdallas:
I was going to comment about the wizard header as well.

Also, I was wondering why you hid the button like that instead of using `showClose={false}` on the modal.. but when i attempt to do that, I get an error about a focus trap without a tabbable node. Perhaps this would change with the wizard header in place.

sjd78 (commented):
Generally looks ok, but another issue -- on the Assessment Actions screen, the archived questionnaires View buttons have a spinner to the left:
image

gitdallas (changes requested):
I'm still seeing the focus trap without tabbable node error, also I think this should be reworded now that it isn't a page:

image

perhaps just replace "page" with "assessment" or "questionnare"?

client/src/app/pages/assessment/components/assessment-wizard/assessment-wizard-modal.tsx
```diff
@@ -0,0 +1,44 @@
...
+  const { assessment, isFetching, fetchError } =
+    useFetchAssessmentById(assessmentId);
+
+  return (
+    <>
+      {isOpen && (
+        <Modal
+          isOpen={isOpen}
+          onClose={onRequestClose}
+          showClose={false}
```
gitdallas, line 26:
```suggestion
          showClose={isFetching || !!fetchError}
```

this seems to fix the focus trap issue. it is an issue when you have an error/loader rather than the wizard.

client/src/app/pages/assessment/components/assessment-wizard/assessment-wizard.tsx
```diff
@@ -583,13 +583,13 @@ export const AssessmentWizard: React.FC<AssessmentWizardProps> = ({
             <ConfirmDialog
               title={t("dialog.title.leavePage")}
```
gitdallas, line 584: title should probably be changed too. "Close Assessment"? or if you want to use existing message... maybe `dialog.title.discard` with Assessment for the `what`?

gitdallas (commented):
Mostly lgtm, one verbiage comment

sjd78 (commented):
Code changes LGTM.
The assessment actions page and the wizard all work well.

As already noted, the title for the confirm cancel modal is confusing since the assessment is a wizard and not a page anymore.

client/src/app/pages/assessment/components/assessment-wizard/assessment-wizard.tsx
sjd78: Other possible confirm modal titles:

  - "Leave assessment"
  - "Confirm cancel"
  - "Close wizard"

gitdallas (approved):

sjd78 (approved):
LGTM with a nice confirm title and everything!

PR #1487, opened 2023-10-24
:bug: Accurately represent upload state in binary upload step

ibolton336:
- Address https://issues.redhat.com/browse/MTA-1420 
- Removes bg polling while wizard is open 
- Adds a check for ongoing mutations while the file upload operation is underway.


sjd78 (commented):
LGTM

PR #1491, opened 2023-10-26
:ghost: Restore reports page

ibolton336:
- Adds reports page back to side drawer
- Refactors adoption candidate table to use new table format. 
- Started refactoring identified risks, but it doesn't make sense with our current model to list all questions for a given questionnaire with a "risk" value and an associated applications column. It just simply doesn't map to the new model & needs more design input. 

https://issues.redhat.com/browse/MTA-1345?filter=-1


ibolton336:
image

client/src/app/pages/reports/components/adoption-candidate-table/adoption-candidate-table.tsx
```diff
@@ -1,252 +1,141 @@
...
+        numRenderedColumns={numRenderedColumns}
+        noDataEmptyState={
+          <div>
+            <NoDataEmptyState title="No data available." />
+          </div>
+        }
+      >
+        <Tbody>
+          {currentPageItems?.map((item, rowIndex) => {
+            return (
```
sjd78, line 106: Nit-pick -- skip the open `,rowIndex) => { return (` and just go with `, rowIndex) => (`

sjd78 (changes requested):
Looking decent.  A few comments inline.  The blocking comments right now are the ones on if a display value of `0` is acceptable and therefore should use a `??` instead of a `||`.

More generally, I strongly urge you to move the data access & transform logic to a hook.  Not only does it keep the data transforms separate from the render and event code, but it will allow a **much** easier unit test to be setup.  I get the feeling that verifying the transforms with a few unit tests on a data hook will come in handy.

Issue #1242, opened 2023-08-01
Replace deprecated PF Select menus


gitdallas:
Will require some refactoring. Several instances of `Select`, `SelectOption`, `SelectVariant`, `SelectOptionObject`, `SelectProps`, `SelectOptionProps`

https://www.patternfly.org/components/menus/select

dymurray:
Closing due to inactivity or age, please reopen if you feel this was done in error

PR #1496, opened 2023-10-27
:ghost: use pf5 select for context selector

gitdallas:
progress towards #1242 .. using new select for context selector


sjd78 (commented):
Code looks good.  The styling is off from previous.  Is that expected?

Previous:
image

With PR:
image

The borders are different.

The bottom margin/padding on the new one is a bit taller.

gitdallas:
The padding seemed a bit arbitrary, idk.

The border with blue on the bottom is active, the old one does that as well. There are some differences in the border with the new version of the menu toggle as well though. I just let the PF styles be as far as that goes.

I also pulled the context selector component outside of the nav component (it shouldn't have been there)

sjd78:
> The padding seemed a bit arbitrary, idk.
> 

Can the padding around the context selector at least be equal so it is centered in its space? 

> The border with blue on the bottom is active, the old one does that as well. There are some differences in the border with the new version of the menu toggle as well though. I just let the PF styles be as far as that goes.
> 

That's a reasonable approach.

> I also pulled the context selector component outside of the nav component (it shouldn't have been there)

:+1:

PR #1500, opened 2023-10-31
:sparkles: Add provider dropdown for targets

ibolton336:
[MTA-1536](https://issues.redhat.com/browse/MTA-1536)


sjd78 (approved):
LGTM as long as the user does not need to see targets across all languages at the same time.

sjd78 (approved):
Still LGTM as long as a "Show Targets For All Languages" option is not needed.

PR #1502, opened 2023-11-01
:ghost: Default to production for NODE_ENV

ibolton336:
Fixes https://issues.redhat.com/browse/MTA-1599

### DIFF
--- a/common/src/environment.ts
+++ b/common/src/environment.ts
@@ -58,10 +58,12 @@ export const SERVER_ENV_KEYS = [
  * Create a `KonveyorEnv` from a partial `KonveyorEnv` with a set of default values.
  */
 export const buildKonveyorEnv = ({
-  NODE_ENV = "development",
+  NODE_ENV = "production",
+  PORT,
   VERSION = "99.0.0",
   MOCK = "off",
 
+  KEYCLOAK_SERVER_URL,
   AUTH_REQUIRED = "false",
   KEYCLOAK_REALM = "tackle",
   KEYCLOAK_CLIENT_ID = "tackle-ui",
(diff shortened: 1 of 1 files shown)


sjd78 (approved):
Yeah, the destructuring in `buildKonveyorEnv` to convert process.env to the internal ENV was filtering out optional values that I hadn't included.

ibolton336:
> Yeah, the destructuring in `buildKonveyorEnv` to convert process.env to the internal ENV was filtering out optional values that I hadn't included.

Didn't appear to impact the issue at hand, but is nice to have the defaults explicitly included to enhance the dev experience. Seems that the result of buildKonveyorEnv includes all of the env vars available from process.env at runtime even if the type definition constrains the expected properties. TypeScript type checking occurs at compile-time, and it doesn't prevent you from accessing properties not explicitly defined in the type. 

If we wanted to ensure that only the expected properties are included and to have stricter type checking, we could explicitly specify the properties we want from process.env 
```
const konveyorEnv = buildKonveyorEnv({
  NODE_ENV: process.env.NODE_ENV,
  VERSION: process.env.VERSION,
  MOCK: process.env.MOCK,
  AUTH_REQUIRED: process.env.AUTH_REQUIRED,
  KEYCLOAK_REALM: process.env.KEYCLOAK_REALM,
  KEYCLOAK_CLIENT_ID: process.env.KEYCLOAK_CLIENT_ID,
  PROFILE: process.env.PROFILE,
  UI_INGRESS_PROXY_BODY_SIZE: process.env.UI_INGRESS_PROXY_BODY_SIZE,
  RWX_SUPPORTED: process.env.RWX_SUPPORTED,
});
```

PR #1503, opened 2023-11-01
:bug: Add widths to col headers

ibolton336:
https://issues.redhat.com/browse/MTA-1605

### DIFF
--- a/client/src/app/pages/applications/applications-table-analyze/applications-table-analyze.tsx
+++ b/client/src/app/pages/applications/applications-table-analyze/applications-table-analyze.tsx
@@ -570,12 +570,15 @@ export const ApplicationsTableAnalyze: React.FC = () => {
           <Thead>
             <Tr>
               <TableHeaderContentWithControls {...tableControls}>
-                <Th {...getThProps({ columnKey: "name" })} />
-                <Th {...getThProps({ columnKey: "description" })} />
-                <Th {...getThProps({ columnKey: "businessService" })} />
-                <Th {...getThProps({ columnKey: "analysis" })} />
-                <Th {...getThProps({ columnKey: "tags" })} />
-                <Th {...getThProps({ columnKey: "effort" })} />
+                <Th {...getThProps({ columnKey: "name" })} width={20} />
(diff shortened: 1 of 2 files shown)


client/src/app/pages/applications/applications-table-analyze/applications-table-analyze.tsx
```diff
@@ -570,12 +570,15 @@ export const ApplicationsTableAnalyze: React.FC = () => {
...
+                <Th {...getThProps({ columnKey: "name" })} width={20} />
+                <Th {...getThProps({ columnKey: "description" })} width={25} />
+                <Th
+                  {...getThProps({ columnKey: "businessService" })}
+                  width={10}
+                />
+                <Th {...getThProps({ columnKey: "analysis" })} width={10} />
+                <Th {...getThProps({ columnKey: "tags" })} width={10} />
+                <Th {...getThProps({ columnKey: "effort" })} width={10} />
                 <Th />
```
gitdallas, line 582: mentioned in slack but these numbers appear to be % that should add up to 100, also we should probably add to actions column as well

gitdallas (changes requested):

sjd78 (commented):
**TL;DR** - Widths are in percents.  BUT, the `width` prop on `Th` and `Td` is mapped through CSS class names to those percentages.  Therefore, the percentages allowed in the width prop are exactly: `10 | 15 | 20 | 25 | 30 | 35 | 40 | 45 | 50 | 60 | 70 | 80 | 90 | 100`.  Any other values are discarded.

`Th` and `Td` widths are converted to css react-style class names by this function:  https://github.com/patternfly/patternfly-react/blob/main/packages/react-table/src/components/Table/utils/decorators/cellWidth.ts#L22

And then maps through react-styles objects to these css classes: https://github.com/patternfly/patternfly/blob/73685ad0b050492e08309c3ba840ba93b1c7161f/src/patternfly/components/Table/table.scss#L1379C20-L1436

It would be so great if the patternfly docs had Th.width and Td.width defined as such.

PR #1506, opened 2023-11-01
:sparkles: Add an unlink button for removing underlying tickets

ibolton336:
https://issues.redhat.com/browse/MTA-1539


sjd78 (commented):
The jira issue has the link on the table and in the application drawer / details tab.  This change only adds the link to the table.

Does it make sense to skip a "confirm unlink" modal?

PR #1507, opened 2023-11-01
:sparkles: Add archetype filter for applications

ibolton336:
https://issues.redhat.com/browse/MTA-1541


sjd78 (approved):
LGTM as long as the set of archetypes in the list to select from really can be just the archetypes attached to applications and not the full list of archetypes.

PR #1516, opened 2023-11-06
:bug: Add provider to custom target form

ibolton336:
Resolves https://issues.redhat.com/browse/MTA-1627 as discussed in the UXD call 11-7-23
Screenshot 2023-11-07 at 3 50 28 PM

### DIFF
--- a/client/src/app/pages/migration-targets/components/custom-target-form.tsx
+++ b/client/src/app/pages/migration-targets/components/custom-target-form.tsx
@@ -28,7 +28,14 @@ import {
 } from "@app/components/HookFormPFFields";
 import { getAxiosErrorMessage } from "@app/utils/utils";
 import { useCreateFileMutation } from "@app/queries/targets";
-import { IReadFile, New, Rule, Target, TargetLabel } from "@app/api/models";
+import {
+  IReadFile,
+  New,
+  ProviderType,
+  Rule,
+  Target,
+  TargetLabel,
+} from "@app/api/models";
 import { getParsedLabel, parseRules } from "@app/utils/rules-utils";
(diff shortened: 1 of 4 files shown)


client/src/app/pages/migration-targets/components/custom-target-form.tsx
```diff
@@ -391,6 +405,28 @@ export const CustomTargetForm: React.FC<CustomTargetFormProps> = ({
...
   return (
     <Form onSubmit={handleSubmit(onSubmit)}>
+      <HookFormPFGroupController
+        control={control}
+        name="provider"
+        label="Provider "
+        fieldId="provider-select"
+        isRequired
+        renderInput={({ field: { value, name, onChange } }) => (
+          <SimpleSelect
```
sjd78, line 415: Why not a `SimpleSelectBasic` instead to use the newer component?

client/src/app/pages/migration-targets/components/custom-target-form.tsx
```diff
@@ -105,6 +117,7 @@ export const CustomTargetForm: React.FC<CustomTargetFormProps> = ({
     .object()
     .shape({
       id: yup.number().defined(),
+      provider: yup.string().defined(),
```
sjd78, line 120: More informational... I'm pretty sure something like this would work:
```suggestion
      provider: yup.string().oneOf(["Java", "Go"]).defined(),
```

sjd78 (commented):
Looks ok to me.  Just 1 question.

sjd78:
And the solution is to just show the selected provider type from the targets page right?  We're not allowing a change in the provider type on this form.

PR #1518, opened 2023-11-06
:bug: Fix layout bug on row click (intermittent)

ibolton336:
- Addresses an issue with the drawer click functionality which occurs when clicking in-between the columns on a table. The entire layout is left in an unrecoverable state.

https://github.com/konveyor/tackle2-ui/assets/11218376/a3425889-f5f9-4beb-99b1-4c768df0bea4


gitdallas (approved):
nit: maybe worth commenting out the entire useEffect as it won't be doing anything... not that anyone would notice a performance hit i guess..

PR #1521, opened 2023-11-06
:bug: Identified risk refactor

ibolton336:
- Drives the table from assessment data as requested
Resolves https://issues.redhat.com/browse/MTA-1434?


client/src/app/pages/reports/components/identified-risks-table/identified-risks-table.tsx
```diff
@@ -1,161 +1,197 @@
...
-      ],
+  assessments.forEach((assessment) => {
+    assessment.sections.forEach((section) => {
+      section.questions.forEach((question) => {
+        question.answers.forEach((answer) => {
+          if (answer.selected) {
+            const existingItemIndex = tableData.findIndex(
+              (item) =>
+                item.question === question.text &&
+                item.answer === answer.text &&
```
sjd78, line 47: So this will have the same problem of if an assessment has repeated questions / answers right?

Maybe building up a key based on the order fields would be better?

```js
itemId = [
  assessment.id,
  section.order,
  question.order,
  answer.order,
].join(".");
```

sjd78 (commented):
I couldn't test the table render with the one questionnaire I have locally -- the dreaded `hack/import-questionnaire/cloud-native.yaml`!  Seems like this report table has some issues with questions that repeat across sections.

The code looks good, I just couldn't test it.

ibolton336:
> I couldn't test the table render with the one questionnaire I have locally -- the dreaded `hack/import-questionnaire/cloud-native.yaml`! Seems like this report table has some issues with questions that repeat across sections.
> 
> The code looks good, I just couldn't test it.

Addressed this bug here: https://github.com/konveyor/tackle2-ui/pull/1524

PR #1524, opened 2023-11-07
:bug: Fix duplicate question bug when same text exists in multiple sections

ibolton336:


client/src/app/queries/assessments.ts
```diff
@@ -23,8 +27,11 @@ export const useFetchAssessments = () => {
     queryFn: getAssessments,
     onError: (error: AxiosError) => console.log("error, ", error),
   });
+  const assessmentsWithOrder: AssessmentWithSectionOrder[] =
+    data?.map(addSectionOrderToQuestions) || [];
```
sjd78, line 31: The heart of the whole thing right there!

I don't know if `data` would be stable across fetches that return the same data, but if it is, wrapping this transform in a `useMemo` could help.  It's ok as it is, especially if `data` will always be different across fetches even if its content is the same.

sjd78 (approved):
LGTM

It is nice to see a few more query hooks do a bit of useful data transforms that can be nicely reused.

PR #1526, opened 2023-11-08
:ghost: use pf5 select for set-targets

gitdallas:
Also pulled the select out of the text-content wrapper which was throwing off some styles.


client/src/app/pages/applications/analysis-wizard/set-targets.tsx
```diff
@@ -135,20 +135,27 @@ export const SetTargets: React.FC = () => {
...
-            setProvider(selection as string);
-          }}
-        />
       </TextContent>
+      <SimpleSelectTypeahead
+        width={200}
+        value={provider}
+        toggleAriaLabel="Action select dropdown toggle"
+        toggleId="action-select-toggle"
+        id="action-select"
```
ibolton336, line 144: Is there a way we can add a prob to disable the clear functionality?
Screenshot 2023-11-13 at 11 31 52 AM

ibolton336 (changes requested):

client/src/app/pages/applications/analysis-wizard/set-targets.tsx
```diff
@@ -135,20 +135,28 @@ export const SetTargets: React.FC = () => {
...
-          toggleAriaLabel="Action select dropdown toggle"
-          aria-label={"Select provider"}
-          value={provider}
-          options={["Java", "Go"]}
-          onChange={(selection) => {
-            setProvider(selection as string);
-          }}
-        />
       </TextContent>
+      <SimpleSelectTypeahead
```
sjd78, line 139: Should the `SimpleSelectTypeahead` actually be contained within the above `TextContent` to match the location of the previous `SimpleSelect`?

sjd78: ... Nope.  If the new component is inside the `TextContent`, the drop down menu renders weird:
image


outside of the `TextContent` it is fine.

sjd78 (changes requested):
The typeahead isn't working quite right with the search string.  Looks like the search string can disappear but still be applied:

[screencast-localhost_9000-2023.11.20-14_14_42.webm](https://github.com/konveyor/tackle2-ui/assets/3985964/84d76355-f3a0-43c5-87f5-f2952171c918)

client/src/app/pages/applications/analysis-wizard/set-targets.tsx
gitdallas: Yep, this is related to my comment at the top of this PR

sjd78 (commented):
The typeahead filter is now working as expect.

There is one minor annoyance now.  When blurring, the filter gets cleared, the select menu gets re-rendered to show all values and then it closes:

[screencast-localhost_9000-2023.11.20-15_16_40.webm](https://github.com/konveyor/tackle2-ui/assets/3985964/b0d4442f-bc64-492f-ade9-837d7b3d34c9)

Any way to close the menu and then reset the filter so the menu doesn't flash all values before closing?

gitdallas:
interesting... i don't have it running to test and have been rolled onto another project for now. i'll see if i can look tomorrow.. maybe just clearing on open will be good instead... hopefully don't need to restort to setTimeout

ibolton336 (approved):
Looks good! Thanks @gitdallas

sjd78 (approved):
Latest change makes the dropdown much better behaved.

PR #1535, opened 2023-11-09
:sparkles: Add Tag Categories to Tag filters

sjd78:
Resolves: https://issues.redhat.com/browse/MTA-1534

## Summary
On the following tables, the Tag filter has been enhanced to show the Tag with Tag Category:
  - Applications
  - Issues / All Issues
  - Dependencies
  - Dependency Details Drawer / Application Table

Tables that use Tag filters now use a set of options in the format "Tag Category Name / Tag Name".  The filter render component `MultiselectFilterControl` has been updated to have special handling of selected values where the selection chips display the Tag Name and the Tag Category Name is displayed as a tooltip.

Existing item filtering and filter values are still operating as a `string` or `string[]`, but use a string that can uniquely identify each tag.  The net effect means that if a tag of the same name exists in more than one category, they are unique.  Selecting one of those Tags will only select the tag in that specific category.

Additional change details:
  - Type `OptionPropsWithKey` renamed `FilterSelectOptionProps` for more consistent type/component naming
  - Added some jsdoc to `FilterToolbar`

## Screenshots
**Tag chips displayed for selected tags**:
image

**Tag filter select**:
image

### DIFF
--- a/client/src/app/components/FilterToolbar/FilterToolbar.tsx
+++ b/client/src/app/components/FilterToolbar/FilterToolbar.tsx
@@ -47,7 +47,7 @@ export interface IBasicFilterCategory<
    * For server-side filtering, return the search value for currently selected filter items.
    * Defaults to using the UI state's value if omitted.
    */
-  getServerFilterValue?: (filterValue: FilterValue) => FilterValue;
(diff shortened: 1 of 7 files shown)


ibolton336 (approved):
LGTM!

PR #1538, opened 2023-11-13
:bug: Allow save as draft on first page if answers exist

ibolton336:
Resolves https://issues.redhat.com/browse/MTA-1676 
- QE raised the point that we should be able to save as draft on the first page of the wizard if the user has navigated back after selecting answers.


sjd78 (approved):
LGTM

Issue #1339, opened 2023-09-08
Missing column "Management" in the Dependencies side drawer


rromannissen:
### Environment

- Fedora 36
- minikube v1.31.2 (--driver=kvm2 --memory=10g --cpus=3)
- Konveyor 0.3 alpha-2 with additional images:
  - ui_image_fqin: quay.io/ibolton/tackle2-ui:demo
  - hub_image_fqin: quay.io/jortel/tackle2-hub:demo
  - analyzer_fqin: quay.io/jortel/tackle2-addon-analyzer:demo
- [tackle-testapp](https://github.com/konveyor/tackle-testapp) application
- Analysis configuration: 
  - Source + Dependencies
  - cloud-readiness, linux
  - Application and internal dependencies only
  - No custom rules
  - No additional advanced options

### Description

The _Management_ column that defines if a dependency is managed or embedded is missing from the side drawer in the _Dependencies_ view.

### How to reproduce

- Click on the _Dependencies_ option in the left menu.
- Click on any dependency from the main table.

### Expected result

![Screenshot from 2023-09-08 13-28-55](https://github.com/konveyor/tackle2-ui/assets/15966456/3d0ec720-42e6-4fae-a814-5fcaf5beb75c)

### Actual result

![Screenshot from 2023-09-08 13-47-18](https://github.com/konveyor/tackle2-ui/assets/15966456/4568e228-3729-48f9-a39a-177b8f9134d7)

rromannissen:
@dymurray Reproduced in 0.3-alpha5

PR #1540, opened 2023-11-14
:bug: Render dependency versions as links

sjd78:
Resolves: #1340

On the Dependencies page, detail drawer, applications table: render the version text as a link to the maven central repository.  The link uses the dependency's sha as the key in the maven central search.

Add component `ExternalLink` to standardize rendering links outside of the app opening in a new tab.

Note: The resolution to #1338 will require a change to the application table's query filter and response object.  This will be done in PR #1536.  See https://github.com/konveyor/tackle2-hub/issues/557 for the response object change.

Screenshot:
image

PR #1541, opened 2023-11-14
:sparkles: Add management column for application dependencies table

ibolton336:
- Add management column for application dependencies table
- Add parsed label conversion for dependency labels
https://github.com/konveyor/tackle2-ui/issues/1339

Depends on https://github.com/konveyor/tackle2-addon-analyzer/pull/62

### DIFF
--- a/client/src/app/pages/dependencies/dependency-apps-table.tsx
+++ b/client/src/app/pages/dependencies/dependency-apps-table.tsx
@@ -1,10 +1,15 @@
 import * as React from "react";
 import { useTranslation } from "react-i18next";
-import { Toolbar, ToolbarContent, ToolbarItem } from "@patternfly/react-core";
+import {
+  TextContent,
+  Toolbar,
+  ToolbarContent,
+  ToolbarItem,
+} from "@patternfly/react-core";
 import { Table, Tbody, Td, Th, Thead, Tr } from "@patternfly/react-table";
 import spacing from "@patternfly/react-styles/css/utilities/Spacing/spacing";
 import { useSelectionState } from "@migtools/lib-ui";
-import { AnalysisDependency } from "@app/api/models";
+import { AnalysisAppDependency, AnalysisDependency } from "@app/api/models";
 import {
   useTableControlState,
   useTableControlProps,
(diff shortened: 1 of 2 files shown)


client/src/app/pages/dependencies/dependency-apps-table.tsx
```diff
@@ -139,6 +140,15 @@ export const DependencyAppsTable: React.FC<IDependencyAppsTableProps> = ({
     },
   } = tableControls;
 
+  const getManagementValue = (appDependency: AnalysisAppDependency) => {
```
sjd78, line 143: What about `s/getManagementValue/isDependencyManaged/` ?

client/src/app/pages/dependencies/dependency-apps-table.tsx
```diff
@@ -139,6 +140,15 @@ export const DependencyAppsTable: React.FC<IDependencyAppsTableProps> = ({
...
+    const hasJavaLabel = appDependency.dependency.labels.some((label) => {
+      const labelValue = getParsedLabel(label).labelValue;
+      return labelValue === "java";
+    });
+    const isJavaFile = dependency.name.endsWith(".jar");
```
sjd78, line 148: 1. Probably better to use `appDependency.dependency.name` just to keep the sourced data in the same place for the function.  Or move the function and the TD render to a column render helper.  Those helpers can be in the same file.  That's what I'm doing in #1540.

2. I don't see any dependency names ending with ".jar".  All the names I see are `mavenGroupId.mavenArtifactId`.  I assume the `appDependency.dependency.provider` string is supposed to be the language, but I've only seen empty string values for my analyses.

sjd78 (changes requested):
Structure looks good.  2 comments.

client/src/app/pages/dependencies/dependency-apps-table.tsx
ibolton336: @sjd78 
@pranavgaikwad mentioned that we only see ".jar" deps when we are using a binary file as the input.

sjd78: Oh that ***is*** an interesting distinction.  

So only "java" things where sources are not available will have a name with a `.jar` suffix.

sjd78 (approved):
LGTM

I'm ok with moving along with the current calculation and adjusting with another PR in future if needed.

sjd78 (approved):
As long as the column `isJavaDependency` tests are good, LGTM

PR #1542, opened 2023-11-15
:bug: Fix header issue causing tabs to be hidden

ibolton336:
Resolves https://issues.redhat.com/browse/MTA-1651

Wrapping the Tabs component inside a div helps manage the overflow issue seen when testing responsiveness of drawer.


sjd78 (commented):
I know it is stupid, but just wrapping the `Tabs` component in a div will make them show up.  Then the header can remain above the scrolling section.

sjd78 (approved):
LGTM

PR #1552, opened 2023-11-16
:ghost: Refactor Dependencies details drawer layout and unique row handling

sjd78:
Changes:
  - Refactor the header and layout of tabs for the dependencies detail drawer following the changes in #1542.

  - Show the dependency name and provider in the header since those two fields together are the row unique key (see #1554).

  - Update the filter for the dependency applications to include both the name and the provider to match the row unique key (see #1554).


ibolton336 (approved):

sjd78:
Screenshot before:
screenshot-localhost_9000-2023 11 16-15_47_44

After:
screenshot-localhost_9000-2023 11 16-15_48_56

PR #1553, opened 2023-11-16
:ghost: Refactor archetype table app column text

sjd78:
On the archetype table, applications column, update the translation key used to be plural aware via i18next:
  - When `count == 1`, use key `archetypeApplicationCount`
  - When `count != 1`, use key `archetypeApplicationCount_plural`

Net result is "1 application" and "3 applications" will be rendered. The "(s)" suffix is no longer required.

Note: The app is currently using `i18next` v19.  In current versions, and therefore current docs, the count suffixes are different.

### DIFF
--- a/client/public/locales/en/translation.json
+++ b/client/public/locales/en/translation.json
@@ -156,7 +156,8 @@
     "small": "Small"
   },
   "message": {
-    "archetypeApplicationCount": "{{count}} applications",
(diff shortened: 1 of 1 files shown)


sjd78:
Screenshot:
screenshot-localhost_9000-2023 11 16-13_51_40

ibolton336 (approved):

Issue #1509, opened 2023-11-02
[BUG] Dependencies Table: When there are deps with duplicate names, the row selection always selects the first one


pranavgaikwad:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Konveyor version

0.3.0.beta-1.2

### Priority

Critical

### Current Behavior

When there are dependencies with duplicate names (this is a valid case, where dependencies can have same name but differ in versions, sha etc), the row selection does not work as expected:

![deps_dups](https://github.com/konveyor/tackle2-ui/assets/9839757/63cc1747-e339-46f8-aa26-7e45de6ec33a)

### Expected Behavior

The row selection should work even when there are duplicate dep names. The uniqueness of a dep is defined by not just the name but combination of name, sha, and version.

### How Reproducible

Always (Default)

### Steps To Reproduce

1. Add following apps to inventory:
  * https://github.com/konveyor/example-applications/tree/main/book-server
  * https://github.com/konveyor/tackle-testapp
  * https://github.com/konveyor/example-applications/tree/main/daytrader
2. Run analysis for all these apps using `Source + Deps` mode. The targets don't really matter, as no matter what targets, the dependencies for an app don't change.
3. Go to dependencies view and notice:

![deps_dups](https://github.com/konveyor/tackle2-ui/assets/9839757/b2c822a5-d527-45d5-89d6-2143a3ae7de6)

### Environment

```markdown
- OS:
```

PR #1554, opened 2023-11-16
:bug: Refactor dependencies, select row by name and provider

sjd78:
Resolves: #1509

The Dependencies table was setup to use the `AnalysisDependency.name` as the selected row key.  Since the `.name` field is not unique across all rows multiple rows would be erroneously selected at the same time.  To solve the problem, a UI only unique id is generated after fetching and used for the table selection itemId.

Change details:
  - Dependencies fetch query updated to inject a UI uid and return `WithUiId<AnalysisDependency>[]`.  This transform is done outside the react-query `useQuery()` handling.

  - Updated the table to use the `WithUiId<>` key `_ui_unique_id` for selection.

  - Since the button rendered in the "Found in" column had no function, the button was removed.

  - "Found in" column moved to the last column to match the placement of matching columns on the Archetypes and Issues tables.

  - Use i18next plural aware keys for the "Found in" text to avoid the "(s)" suffix.

  - Fixed the `id` and `aria-label` on the table.

  - Added `key` to the label column `Label`s.

### DIFF
--- a/client/src/app/queries/dependencies.ts
+++ b/client/src/app/queries/dependencies.ts
@@ -1,14 +1,16 @@
+import { useMemo } from "react";
 import { useQuery } from "@tanstack/react-query";
 import {
   AnalysisAppDependency,
   AnalysisDependency,
   HubPaginatedResult,
   HubRequestParams,
+  WithUiId,
 } from "@app/api/models";
 import { getAppDependencies, getDependencies } from "@app/api/rest";
 
 export interface IDependenciesFetchState {
-  result: HubPaginatedResult<AnalysisDependency>;
+  result: HubPaginatedResult<WithUiId<AnalysisDependency>>;
   isFetching: boolean;
   fetchError: unknown;
   refetch: () => void;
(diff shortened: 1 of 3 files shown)


sjd78:
Before:
screenshot-localhost_9000-2023 11 16-15_22_18

After:
screenshot-localhost_9000-2023 11 16-15_22_53

client/src/app/queries/dependencies.ts
```diff
@@ -32,8 +34,28 @@ export const useFetchDependencies = (
...
+        _ui_unique_id: `${dep.name}/${dep.provider}`,
+      })
+    );
+
+    return {
+      data: syntheticData,
+      total: data.total,
+      params: data.params,
+    };
+  }, [data, params]);
```
mturley, line 55: I wonder if we can abstract out something like:

```ts
useWithUiId((dep) => `${dep.name}/${dep.provider}`);
```

This indicates to me that maybe internally the table-controls/batteries/helpers could just take a `getItemId` callback in the first place if you don't have a usable `idProperty`. Will factor that into thinking on current refactors over there.

LGTM as-is though if you want to merge.

mturley (approved):
LGTM but an abstraction jumped out at me, take it or leave it.

client/src/app/queries/dependencies.ts
sjd78: `idProperty: string | (item: TItem) -> string`?

;-)

PR #1561, opened 2023-11-21
:sparkles: Add question mark icon for archetype assessed indicator

ibolton336:
Resolves https://issues.redhat.com/browse/MTA-1712?filter=12422675


client/src/app/pages/applications/applications-table/applications-table.tsx
```diff
@@ -869,22 +873,45 @@ export const ApplicationsTable: React.FC = () => {
...
-                      <Td
+                      {/* <Td
                         width={10}
                         modifier="truncate"
                         {...getTdProps({ columnKey: "assessment" })}
                       >
                         <ApplicationAssessmentStatus
                           application={application}
                         />
+                      </Td> */}
```
sjd78, line 884: Looks like you can just remove this comment block now.

sjd78 (changes requested):
Something is a bit off with the layout on the infotip icon:
screenshot-localhost_9000-2023 11 21-17_24_21

client/src/app/pages/applications/applications-table/applications-table.tsx
```diff
@@ -869,22 +873,45 @@ export const ApplicationsTable: React.FC = () => {
...
+                        </Grid>
                       </Td>
                       <Td
                         width={10}
                         modifier="truncate"
                         {...getTdProps({ columnKey: "review" })}
                       >
                         <Grid>
-                          <GridItem span={10}>
+                          <GridItem span={8}>
```
sjd78, line 914: Plus the grid is 12 columns, so you're leaving 3 columns to the right empty

sjd78 (approved):
Code looks ok.

The layout still overlaps when the browser window width is less than 1750px:
Screenshot from 2023-11-22 10-08-51

This change is ok, but I do think there are more effective ways to handling displaying the "Not Started" vs "Completed via inheritance from an archetype" vs "Completed assemment/review direct against the application".  Maybe text values of "Not Started", "Inherited", "Completed" or "Not Started", "From Archetype", "Completed"?  That can be a followup issue.

PR #1564, opened 2023-11-21
:bug: Fix reset pw field on focus

ibolton336:
Resolves https://issues.redhat.com/browse/MTA-1266


sjd78 (commented):
The change makes sense, but the ticket doesn't make sense.  Clearing the password field onFocus is a pretty standard experience.

@JustinXHale - I didn't see PF explicitly say anything about this.  WDYT?

JustinXHale:
> https://issues.redhat.com/browse/MTA-1266

The password to persist even if the user clicks away. It doesn't explicitly state it in PF guidelines however the password persisting aligns across RH

PR #1565, opened 2023-11-21
:bug: Add link to associated ticket

ibolton336:
Resolves https://issues.redhat.com/browse/MTA-701
Resolves https://issues.redhat.com/browse/MTA-1539

### DIFF
--- a/client/src/app/pages/migration-waves/components/ticket-issue.tsx
+++ b/client/src/app/pages/migration-waves/components/ticket-issue.tsx
@@ -1,10 +1,9 @@
 import React from "react";
-import { Button, Text, TextVariants } from "@patternfly/react-core";
+import { Text, TextVariants } from "@patternfly/react-core";
 
 import { Ticket } from "@app/api/models";
 import { useTranslation } from "react-i18next";
-import { UnlinkIcon } from "@patternfly/react-icons";
-import { useDeleteTicketMutation } from "@app/queries/migration-waves";
+import ExternalLink from "@app/components/ExternalLink";
 
 export interface ITicketIssueProps {
   ticket?: Ticket;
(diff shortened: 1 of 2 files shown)


client/src/app/pages/migration-waves/components/ticket-issue.tsx
```diff
@@ -1,25 +1,40 @@
...
+  return (
+    <>
+      <Text
+        component={TextVariants.p}
+        className="pf-v5-u-color-200 pf-v5-u-font-weight-light"
+      >
+        {ticket ? (
+          <a href={ticket.link} target="_">
+            {ticket?.link}
+          </a>
```
sjd78, line 26: Use `ExternalLink` for a consistent "link to outside of the app that will open in another tab" view.

The text should probably remain the same as before instead of the actual link.

client/src/app/pages/migration-waves/components/ticket-issue.tsx
```diff
@@ -1,25 +1,40 @@
...
+        ) : (
+          t("terms.unassigned")
+        )}
+        {ticket?.id && (
+          <Button
+            variant="link"
+            icon={<UnlinkIcon />}
+            onClick={() => deleteTicket(ticket.id)}
+          />
+        )}
```
sjd78, line 36: Would it be better to put the unlink icon button in the wave-status-table by the trash icon button?

sjd78 (changes requested):
The only blocking item is the link text used.  The other comments are strong suggestions.

client/src/app/pages/migration-waves/components/ticket-issue.tsx
ibolton336: Not sure only because this is how it was mocked up. I think I do like the idea of moving to the actions column, though.

sjd78: Looks like the mock-up is sufficiently ambiguous on this one:
image

ibolton336: 100%!

client/src/app/pages/migration-waves/components/ticket-issue.tsx
```diff
@@ -1,25 +1,26 @@
...
-    if (type) return type.name;
-    return "";
-  };
+  const { t } = useTranslation();
 
-  const ticketIssue = useTicketIssue();
-
-  return <Text>{ticketIssue}</Text>;
+  return (
+    <>
```
sjd78, line 16: The extra `<></>` wrapper isn't needed.

Looks like previously the column text would be the name of the type of the issue tracker -- so maybe something like "Jira1" or whatever the instance name of a tracker would be.  Probably a good idea to keep that as the URL/Link could be long like "https://issues.redhat.com/browse/MTA-701" vs "Red Hat Jira".

sjd78 (approved):
The link text is the only thing bugging me still.  I'm ok with leaving it and seeing if a ticket comes back with "use the jira instance name please" or similar.

The rest of the code LGTM

PR #1588, opened 2023-12-05
:bug: display assessment delete error only once for bulk action

ibolton336:
Resolves https://issues.redhat.com/browse/MTA-1810


sjd78 (commented):
:+1:

Pushed toast notification responsibility up to the caller of the set of mutations.  Makes sense.

PR #1593, opened 2023-12-06
:bug: Sort Questionnaire items on fetch

sjd78:
A questionnaire's sections, questions and answers all have a order key.  The key needs to be present in the defining yaml file, but does not need to be in sorted order.  To ensure that the UI always has questionnaire items in proper order order, sort everything in the fetch hooks.

Note: Hub returns the questionnaire in the same order as the original yaml file.  We need to make sure things are sorted properly.


ibolton336 (approved):
Works great! LGTM

PR #1594, opened 2023-12-06
:ghost: Add some missing `key` props to fragments

sjd78:
Since using the nice looking version of `React.Fragment` of `<>...</>` does not allow for adding a key prop when a fragment is used in an array/map(), switch to the verbose version and add the key.

Therefore `<>...</>` becomes `<React.Fragment key={...}>...</React.Fragment>`.


ibolton336 (approved):
LGTM

PR #1600, opened 2023-12-07
:bug: Add useEffect for action state sync

ibolton336:
Resolves https://issues.redhat.com/browse/MTA-1826

- Uses a useEffect hook to synchronize the state of the action based on the assessment object. This way, whenever assessment changes, we can ensure that the action is recalculated.


sjd78 (commented):
Just curious if you tried `useMemo` instead of `useEffect`?

It seems like you'd be able to just do the whole thing as a single memo w/o state:
```tsx
  const action = useMemo(() => {
    if (!assessment) {
      return AssessmentAction.Take;
    } else if (assessment.status === "started") {
      return AssessmentAction.Continue;
    } else {
      return AssessmentAction.Retake;
    }
  }, [assessment]);
```

PR #1602, opened 2023-12-08
:bug: Update external links styling across the UI

sjd78:
Resolves: https://issues.redhat.com/browse/MTA-1846

Update the render of external links to be consistent across the UI:
  - Use `ExternalLink` consistently
  - The icon used is the small size of `ExternalLinkAltIcon`
  - The text and the icon are both part of the link

Use of `ExternalLink` updated to use the inline variant when used directly in table cells so it doesn't introduce extra height to the row. In a `td`, we want text and link content to be the same height.

Use updated in:
  - Dependencies detail drawer, application table, version link
  - Issues table, expanded row, details links
  - Migration waves, status sub-table, issue link

### DIFF
--- a/client/src/app/components/ExternalLink.tsx
+++ b/client/src/app/components/ExternalLink.tsx
@@ -1,5 +1,5 @@
 import * as React from "react";
-import { Flex, FlexItem, Icon, Text } from "@patternfly/react-core";
+import { Button, Icon } from "@patternfly/react-core";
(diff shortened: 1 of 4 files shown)


sjd78:
Screenshots of external link use in dependencies and issues tables:

screenshot-localhost_9000-2023 12 07-21_40_43

screenshot-localhost_9000-2023 12 07-21_39_42

PR #1605, opened 2023-12-11
:bug: Fix alignment on actions screen

ibolton336:
- Address issues with css alignment on assessment actions page

Before: 
Screenshot 2023-12-11 at 10 47 34 AM

After: 
Screenshot 2023-12-11 at 10 49 42 AM

### DIFF
--- a/client/src/app/pages/assessment/components/assessment-actions/components/dynamic-assessment-actions-row.css
+++ b/client/src/app/pages/assessment/components/assessment-actions/components/dynamic-assessment-actions-row.css
@@ -7,3 +7,7 @@
   background-color: var(--pf-v5-global--warning-color--100) !important  ;
   margin-right: 10px;
 }
+
+.actions-col {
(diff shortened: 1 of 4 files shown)


sjd78 (commented):
FWIW, I found that having a Button in a TD on any TR that is not inline makes the spacing weird.  Marking the button `<Button isInline ...>...</Button>` seems to take care of the problems most of the time.

ibolton336:
> FWIW, I found that having a Button in a TD on any TR that is not inline makes the spacing weird. Marking the button `<Button isInline ...>...</Button>` seems to take care of the problems most of the time.

Looks like that does partially fix the issue, but still seeing some style inconsistency with long questionnaire names.

Removing style changes to center the row content vertically & just use isInline for the buttons:

Screenshot 2023-12-11 at 1 17 49 PM

VS

Screenshot 2023-12-11 at 1 19 21 PM

PR #1609, opened 2023-12-11
:bug: only render archetypes with associated apps in filter

ibolton336:
Resolves https://issues.redhat.com/browse/MTA-1791

### DIFF
--- a/client/src/app/pages/issues/helpers.ts
+++ b/client/src/app/pages/issues/helpers.ts
@@ -3,6 +3,7 @@ import {
   AnalysisIssue,
   AnalysisIssueReport,
   AnalysisRuleReport,
+  Archetype,
 } from "@app/api/models";
 import {
   FilterCategory,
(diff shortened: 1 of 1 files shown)


sjd78 (changes requested):
Just hiding archetypes w/o dependencies seems weird.  I'd rather see the error message be "No data available".

But....hub doesn't seem to like getting an empty application id list.  And since the archetype filter is implemented by joining the application ids for the archetype selected, havinging the filter's `getServerFilterValue` return `[]` doesn't work.  Probably just give it something that will never work (like a "-1") and it'll be fine....

```js
      getServerFilterValue: (selectedOptions) => {
        const a = selectedOptions
          ?.map((option) => archetypes.find((item) => item.name === option))
          .filter(Boolean)
          .flatMap(
            ({ applications }) =>
              applications?.map(({ id }) => String(id)) ?? []
          )
          .filter(Boolean);

          return a?.length === 0 ? ["-1"] : a;
      },
```

client/src/app/pages/issues/helpers.ts
```diff
@@ -62,16 +63,28 @@ export const useSharedAffectedApplicationFilterCategories = <
         t("actions.filterBy", {
           what: t("terms.archetype").toLowerCase(),
         }) + "...",
-      selectOptions: archetypes.map(({ name }) => ({ key: name, value: name })),
-      getServerFilterValue: (selectedOptions) =>
-        selectedOptions
-          ?.map((option) => archetypes.find((item) => item.name === option))
+      selectOptions: archetypes
+        .filter(({ applications }) => applications && applications.length > 0)
```
sjd78, line 67: This still filters out archetypes w/o applications.  The hack in `getServerFilterValue` allows selection of an "empty" archetype.

sjd78 (approved):
LGTM

Thanks for making the change!

PR #1614, opened 2023-12-12
:bug: Fix session expiry when refresh tokens are disabled

ibolton336:
Resolves https://issues.redhat.com/browse/MTA-1255

- Since we no longer have to support windup reports in a new tab, this code can be cleaned up to use keycloaks internal session management

### DIFF
--- a/client/src/app/axios-config/apiInit.ts
+++ b/client/src/app/axios-config/apiInit.ts
@@ -1,14 +1,43 @@
 import axios from "axios";
+import keycloak from "@app/keycloak";
 
-export const initInterceptors = (getToken: () => Promise<string>) => {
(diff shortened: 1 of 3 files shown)


sjd78 (approved):
Tested and works to kick user back to login if the token cannot be refreshed.

In dev mode, there is a flash of an error screen before forwarding the user to the login screen.  Hopefully that doesn't show when the app is running a prod build.

Code looks ok.  I know `ReactKeycloakProvider` and `keycloak-js` are both deprecated.  Refactoring those out should be on the tech-debt list.

fabianvf:
@ibolton336 UI tests are red on this PR, is it possible this change broke them? I can't find an earlier PR with these failures

PR #1616, opened 2023-12-13
:bug: Only load keycloak when `AUTH_REQUIRED` is "true"

sjd78:
Only add the actual `AuthEnabledKeycloakProvider` keycloak handler if the `AUTH_REQUIRED` environment variable is set `"true"`.


djzager (approved):
VISACK

ibolton336 (approved):

fabianvf (approved):
LGTM

Will merge when CI finishes

PR #1625, opened 2023-12-14
:bug: Fix pagination effects - for client tables, use `filteredItems` instead of `items` length for `totalItemCount`

mturley:
Made the same fix in the new react-table-batteries code as well. Note to self: Perhaps we should rename the `totalItemCount` to `filteredItemCount` in the new version, since it really should be the count of all items just before pagination (after filtering), not the total number of items in the unfiltered collection.

For client-side tables, the `totalItemCount` is derived internally inside `getLocalTableControlDerivedState`, and it was incorrectly using `items.length`. Changing that to `filteredItems.length` makes the logic in `usePaginationEffects` behave as intended, which means if the user ends up on an invalid pagination page (e.g. they are on page 2 and they apply a filter that has results on one page), it will kick them back to the last valid page.

cc @ibolton336

### DIFF
--- a/client/src/app/hooks/table-controls/getLocalTableControlDerivedState.ts
+++ b/client/src/app/hooks/table-controls/getLocalTableControlDerivedState.ts
@@ -48,7 +48,7 @@ export const getLocalTableControlDerivedState = <
     items: sortedItems,
   });
   return {
-    totalItemCount: items.length,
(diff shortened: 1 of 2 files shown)


sjd78 (approved):
Tested on the same table that wasn't working before and it work now!  :smile: 

LGTM

PR #1662, opened 2024-01-11
:bug: Broken search string when navigating from deps page

ibolton336:
Resolves [[Dependencies]Application inventory page shows no applications once Accessed from dependencies](https://issues.redhat.com/browse/MTA-2007) 
Resolves https://issues.redhat.com/browse/MTA-2056

- Removes the auto applied search string when direct navigating using the side nav. This was applying filters that were not registered within the application table ( & other tables). I.E when navigating from the dependencies page, the search string was including an "applications.name" filter which is just referenced as "name" within the applications table. Future iterations may allow us to rewrite the search string when navigating between pages or standardizing on a specific key for app names on the back end. 
- Adds the ability to parse the url string for initial search values in the dependencies toolbar. 
- Adds associated clear filters handler to clear the url when the toolbar clear button is pressed. This is necessary to overwrite the initial filter values. Implemented this in dependencies page & applications table. There may be a need to incorporate this into filter toolbar at some point.

### DIFF
--- a/client/src/app/layout/SidebarApp/SidebarApp.tsx
+++ b/client/src/app/layout/SidebarApp/SidebarApp.tsx
@@ -77,26 +77,17 @@ export const SidebarApp: React.FC = () => {
           {selectedPersona === PersonaKey.MIGRATION ? (
             <NavList title="Global">
               <NavItem>
-                <NavLink
-                  to={Paths.applications + search}
-                  activeClassName="pf-m-current"
-                >
+                <NavLink to={Paths.applications} activeClassName="pf-m-current">
                   {t("sidebar.applicationInventory")}
                 </NavLink>
               </NavItem>
               <NavItem>
-                <NavLink
-                  to={Paths.archetypes + search}
-                  activeClassName="pf-m-current"
-                >
+                <NavLink to={Paths.archetypes} activeClassName="pf-m-current">
                   {t("sidebar.archetypes")}
                 </NavLink>
               </NavItem>
               <NavItem>
-                <NavLink
-                  to={Paths.reports + search}
-                  activeClassName="pf-m-current"
-                >
+                <NavLink to={Paths.reports} activeClassName="pf-m-current">
                   {t("sidebar.reports")}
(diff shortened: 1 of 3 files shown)


sjd78 (commented):
The changes LGTM.

Does the "clear all" function need to be applied to more tables?  Should it exist in all filterable tables?  If yes, either this PR can be expanded, or a followup issue needs to be created.

ibolton336:
> The changes LGTM.
> 
> Does the "clear all" function need to be applied to more tables? Should it exist in all filterable tables? If yes, either this PR can be expanded, or a followup issue needs to be created.

This one is tough because the toolbar is a pf library component which manages the chip state and the clear button state. The toolbar exposes the clearAllFilters callback for special cases like this one when there are initial filter values parsed and loaded from the url query params. Ideally I'd like to move this code into the filter toolbar based on the presence of initialFilterValues, but that doesnt seem possible at the moment.

This PR does cover both of the cases where we are using this initialFilterValue set logic, though. It is a bit annoying that we will have to remember this for any of these cases going forward.

sjd78:
> > The changes LGTM.
> > Does the "clear all" function need to be applied to more tables? Should it exist in all filterable tables? If yes, either this PR can be expanded, or a followup issue needs to be created.
> 
> This one is tough because the toolbar is a pf library component which manages the chip state and the clear button state. The toolbar exposes the clearAllFilters callback for special cases like this one when there are initial filter values parsed and loaded from the url query params. Ideally I'd like to move this code into our custom filter toolbar component based on the presence of initialFilterValues, but that doesn't seem possible at the moment.

Yeah that makes this one a bit rough for sure.

> 
> This PR does cover both of the cases where we are using this initialFilterValue set logic, though. It is a bit annoying that we will have to remember this for any of these cases going forward.

Makes sense.  Probably best to open an issue to log this future work and make it easy to find if/when the search string stuff comes up again.

ibolton336:
https://github.com/konveyor/tackle2-ui/issues/1667

PR #1666, opened 2024-01-17
:bug: Exclude non-required questionnaires in assessed calculation

ibolton336:
Resolves https://issues.redhat.com/browse/MTA-1967

Addresses an issue where the assessed boolean for archetypes is set to true when all questionnaires are marked as not-required. Uses the new assessment.required boolean surfaced in https://github.com/konveyor/tackle2-hub/commit/25522337608b2dbab5df85ce4a5b5a9fc78a2289 to filter out these false positives.

### DIFF
--- a/client/src/app/pages/applications/components/application-detail-drawer/components/assessed-archetypes.tsx
+++ b/client/src/app/pages/applications/components/application-detail-drawer/components/assessed-archetypes.tsx
@@ -1,5 +1,5 @@
 import React from "react";
-import { Ref } from "@app/api/models";
+import { Application } from "@app/api/models";
 import { Label, LabelGroup, Spinner } from "@patternfly/react-core";
 import { EmptyTextMessage } from "@app/components/EmptyTextMessage";
 import { useTranslation } from "react-i18next";
@@ -7,17 +7,17 @@ import { useFetchArchetypes } from "@app/queries/archetypes";
 import { useFetchAllAssessmentsWithArchetypes } from "@app/queries/assessments";
 
 interface IAssessedArchetypesProps {
-  archetypeRefs: Ref[] | undefined;
+  application: Application | null;
 }
 
 export const AssessedArchetypes: React.FC<IAssessedArchetypesProps> = ({
-  archetypeRefs,
+  application,
 }) => {
   const { t } = useTranslation();
-  const { archetypes, isFetching: isFetchingArchetypes } = useFetchArchetypes();
-  const applicationArchetypes = archetypes.filter(
-    (archetype) => archetypeRefs?.some((ref) => ref.id === archetype.id)
-  );
+  const {
+    archetypes: applicationArchetypes,
+    isFetching: isFetchingArchetypes,
+  } = useFetchArchetypes(application);
 
   const {
     assessmentsWithArchetypes,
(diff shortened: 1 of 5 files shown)


client/src/app/queries/assessments.ts
```diff
@@ -262,3 +264,33 @@ export const useFetchAssessmentsWithArchetypeApplications = () => {
...
   };
 };
+
+export const useFetchAllAssessmentsWithArchetypes = (
+  archetypes: Archetype[]
+) => {
+  const assessmentQueries = useQueries({
+    queries: archetypes.map((archetype) => ({
+      queryKey: ["assessmentsForArchetype", archetype.id],
+      queryFn: () => getAssessmentsByItemId(true, archetype.id), // Replace with actual API call
```
sjd78, line 274: Is this comment still valid?

sjd78: Also, kinda a bummer we can't just send a single request to "get assessments for this list of archetype ids".

client/src/app/pages/applications/components/application-detail-drawer/components/assessed-archetypes.tsx
```diff
@@ -0,0 +1,51 @@
...
+}
+
+export const AssessedArchetypes: React.FC<IAssessedArchetypesProps> = ({
+  archetypeRefs,
+}) => {
+  const { t } = useTranslation();
+  const { archetypes, isFetching: isFetchingArchetypes } = useFetchArchetypes();
+  const applicationArchetypes = archetypes.filter(
+    (archetype) => archetypeRefs?.some((ref) => ref.id === archetype.id)
+  );
```
sjd78, line 20: Just a thought -- this kind of thing could be embedded directly in the hook so the hook use could look like:
```js
  const { archetypes, ... } = useFetchArchetypes({ 
    includeIds: archetypeRefs.map((a) => a.id) 
  });
```

or even somethings more like:
```js
  const { archetypes, ... } = useFetchArchetypes({
    forApplication: application
  });
```

Then if every the rest endpoint gets a filter itself, that can rework can be handled directly in the query hook.

client/src/app/pages/applications/components/application-detail-drawer/components/assessed-archetypes.tsx
```diff
@@ -0,0 +1,51 @@
...
+
+  const {
+    assessmentsWithArchetypes,
+    isLoading: isFetchingAllAssessmentsWithArchetypesLoading,
+  } = useFetchAllAssessmentsWithArchetypes(applicationArchetypes || []);
```
sjd78, line 25: The `|| []` is easier to do as a default value in the hook function.

client/src/app/queries/assessments.ts
```diff
@@ -262,3 +264,33 @@ export const useFetchAssessmentsWithArchetypeApplications = () => {
     isLoading: assessmentsLoading || isArchetypesLoading,
   };
 };
+
+export const useFetchAllAssessmentsWithArchetypes = (
+  archetypes: Archetype[]
```
sjd78, line 269: Can just have `[]` as the default value and allow the param to be optional:
```suggestion
  archetypes: Archetype[] = []
```

client/src/app/pages/applications/components/application-detail-drawer/components/assessed-archetypes.tsx
```diff
@@ -0,0 +1,51 @@
...
+  const filteredArchetypes = assessmentsWithArchetypes
```
sjd78, line 26: These are assessed archetypes where at least one of the assessments is required.  I think this matches the PR description, but it was a bit confusing at first look.

sjd78 (commented):
A few comments and thoughts

client/src/app/pages/applications/components/application-detail-drawer/components/assessed-archetypes.tsx
ibolton336: Interesting - I wonder if the api supports this?

PR #1669, opened 2024-01-19
:ghost: Convert single/multi filter toolbar controls to use pf5 select 

ibolton336:
- Resolves:
   - https://issues.redhat.com/browse/MTA-1872 
   - https://issues.redhat.com/browse/MTA-467

  

 - Upgrade our usage of the PF select component to the pf 5 supported version.

### DIFF
--- a/client/src/app/components/FilterToolbar/SelectFilterControl.tsx
+++ b/client/src/app/components/FilterToolbar/SelectFilterControl.tsx
@@ -5,7 +5,6 @@ import {
   Select,
   SelectList,
   SelectOption,
-  SelectOptionProps,
   ToolbarFilter,
 } from "@patternfly/react-core";
 import { IFilterControlProps } from "./FilterControl";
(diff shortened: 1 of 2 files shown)


sjd78 (changes requested):
Some problems I found playing with the new filter toolbars:
  - Deselecting an item in a filter list doesn't work / each time you click the same entry, another chip is added
    - application inventory page, with multiple applications available
    - open the name filter dropdown with or without filter text
    - click to select a name
    - name cannot be unselected here, only by clicking the x in the name chips list

  - The filter type drop down does not auto-close

  - If the multiselect filter select drop down opens from the twistee, keyboard focus is left on the checkbox.  But on selecting an item (space bar on the checkbox), focus goes back to the filter input box and any further attempt to select items with the spacebar actually types in the filter and changes the list.  Only hitting enter will allow selecting a second item, but this also closes the dropdown.

For example:

https://github.com/konveyor/tackle2-ui/assets/3985964/45781083-23c0-4b2b-a892-4fff1d8b92c5

sjd78 (commented):
This review is just on the single select control `SelectFilterControl`.  The control looks and works ok.

I have some suggestions to simplify the code in `SelectFilterControl`.

client/src/app/components/FilterToolbar/SelectFilterControl.tsx
```diff
@@ -91,15 +103,29 @@ export const SelectFilterControl = <TItem, TFilterCategoryKey extends string>({
...
-        isDisabled={isDisabled || category.selectOptions.length === 0}
+        placeholder="Any"
       >
-        {renderSelectOptions(category.selectOptions)}
+        <SelectList>
+          {category.selectOptions.map((o: SelectOptionProps, index) => {
+            return (
+              <SelectOption
+                key={`${index}-${o.value}`}
+                isSelected={o.value == filterValue}
```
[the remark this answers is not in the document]
ibolton336: Ah yeah looks like there is an isSelected handler inside the selectoption component so no need to explicitly check here. Good catch.

ibolton336:
> Some problems I found playing with the new filter toolbars:
> 
> * Deselecting an item in a filter list doesn't work / each time you click the same entry, another chip is added
>   
>   * application inventory page, with multiple applications available
>   * open the name filter dropdown with or without filter text
>   * click to select a name
>   * name cannot be unselected here, only by clicking the x in the name chips list
> * The filter type drop down does not auto-close
> * If the multiselect filter select drop down opens from the twistee, keyboard focus is left on the checkbox.  But on selecting an item (space bar on the checkbox), focus goes back to the filter input box and any further attempt to select items with the spacebar actually types in the filter and changes the list.  Only hitting enter will allow selecting a second item, but this also closes the dropdown.
> 
> For example:
> 
>  screencast-localhost_9000-2024.02.13-13_24_54.mp4

Cleaned things up a bit to more closely match this example https://www.patternfly.org/components/menus/select#multiple-typeahead-with-checkboxes & fixed most of the bugs you mentioned.

client/src/app/components/FilterToolbar/SelectFilterControl.tsx
```diff
@@ -91,15 +75,25 @@ export const SelectFilterControl = <TItem, TFilterCategoryKey extends string>({
...
-        isDisabled={isDisabled || category.selectOptions.length === 0}
+        placeholder="Any"
       >
-        {renderSelectOptions(category.selectOptions)}
+        <SelectList>
+          {category.selectOptions.map((o: SelectOptionProps, index) => {
+            return (
+              <SelectOption {...o} key={`${index}-${o.value}`}>
+                {o.value}
+              </SelectOption>
```
sjd78, line 93: I took a look at the filters across the codebase that uses the single select to see what the data coming in look like.  Consider:

https://github.com/konveyor/tackle2-ui/blob/6217116e9c8c7d0481d12ae2b0f7e83940f88724/client/src/app/pages/review/components/application-assessment-summary-table/application-assessment-summary-table.tsx#L77-L95

Then the code there probably won't work since the `selectOptions` key is ignored.

Maybe more like this:
```tsx
  <SelectOption
    {...o}
    key={`${index}-${o.value}`}
    value={o.key}
  >
    {o.value}
  </SelectOption>
```

and then the `chips` would also probably need to be adjusted to convert the selected "key"(aka SelectOption value) to the display "value" (aka SelectOption children).

sjd78 (changes requested):
I've only looked at `SelectFilterControl` so far in this version of changes.  I spent a bunch of time checking out the filter category `selectOptions` data against the PF5 `SelectOption` data.  Looks like a small change could close the data gaps...

I'll have a look at the Multiselect behavior then code next.

client/src/app/components/FilterToolbar/SelectFilterControl.tsx
ibolton336: Updated to address this.

djzager (approved):
ack

Issue #1672, opened 2024-01-29
0.3.1 priorities


ibolton336:
- Archetype assessment/Review status components on archetype table
- Bugs around UI  not reflecting required status for questionnaires.
  -   Reports page landscape component
  - Application assessment status improvements. 
    - Need to nail down expectation when application has archetypes when assessments have been completed, but archived. 
    - Application status when ANY application assessment data exists should show application level status rather than archetype inherited status as it does now. 
    - Need to coordinate with HUB to see if improvements can be made to further rely on the assessed boolean from archetypes when questionnaires are all marked as not required. Currently assessed is "true" when this case arises which feels like a bug.
- Application tags dropdown improvements. Need to leverage new PF5 components to allow grouping of tags by category. 
- Have another look at table behavior when data is preloaded for filters from url. Current solution requires some customization that could potentially be baked into the table library. 
- Debug solution refinements for running analysis 
- Chips in sidebar is not sustainable. want a pre-filter for associated archetypes or applications

sjd78:
@ibolton336 -- Can you update this issue based on the current version or close it?

PR #1673, opened 2024-01-29
:sparkles: Add app name filter to archetypes page & drawer link

ibolton336:


client/public/locales/en/translation.json
```diff
@@ -265,6 +265,8 @@
     "applications": "Applications",
     "archetype": "Archetype",
     "archetypes": "Archetypes",
+    "archetypes_singular": "Archetype",
```
sjd78, line 268: We should really upgrade our i18next version so this key structure is correct (and matches current docs).  IIRC for the current version in use, the singular key should not have "_singular" appended.

client/src/app/pages/applications/components/application-detail-drawer/application-detail-drawer.tsx
```diff
@@ -190,15 +192,25 @@ export const ApplicationDetailDrawer: React.FC<
...
+                          {application.archetypes.length}{" "}
+                          {t("terms.archetypes", {
+                            count: application.archetypes.length,
+                            context:
+                              application.archetypes.length > 1
+                                ? "plural"
+                                : "singular",
+                          }).toLocaleLowerCase()}{" "}
+                        </Link>
+                      </>
```
sjd78, line 207: With this change, we require a user to click a link to even see what archetypes the application is linked to.  I like the click over to the archetype table with filters applied for the application, but I don't like that you can't just look at it in the drawer.

client/src/app/pages/archetypes/archetypes-page.tsx
```diff
@@ -206,15 +213,47 @@ const Archetypes: React.FC = () => {
...
+            archetypes.flatMap(
+              (archetype) =>
+                archetype?.applications
+                  ?.map((app) => app.name)
+                  .filter(Boolean) || []
+            )
+          ),
+        ].map((applicationName) => ({
+          key: applicationName,
+          value: applicationName,
```
sjd78, line 232: Having the key and the value for the `selectOptions` be the same is currently annoying me for no good reason.  I mean as long as the key value matches up with what is returned from `getItemValue()` all is well.

sjd78 (commented):
On the PR description,
  - What part of #1672 is this PR addressing?
  - Can you add a short caption for the screen shots or add an arrow to show what has changed?  Relying on my memory of how something currently looks, or trying to run the app to that point with similar data, to decipher the new screenshot is rough to my less-than-ideally-caffeinated brain.

All of the code changes look ok.  Really my apprehension is with swapping out a list of archetypes for a single link.  Feels like a loss of information in the details drawer.

ibolton336:
> On the PR description,
> 
> * What part of [0.3.1 priorities #1672](https://github.com/konveyor/tackle2-ui/issues/1672) is this PR addressing?
> * Can you add a short caption for the screen shots or add an arrow to show what has changed?  Relying on my memory of how something currently looks, or trying to run the app to that point with similar data, to decipher the new screenshot is rough to my less-than-ideally-caffeinated brain.
> 
Updated description. Sorry about that!

> All of the code changes look ok. Really my apprehension is with swapping out a list of archetypes for a single link. Feels like a loss of information in the details drawer.

The concern with the chips in drawer was scalability. Maybe we can find a happy medium to keep some sort of chip list / detail in drawer along with the link.

sjd78:
> > All of the code changes look ok. Really my apprehension is with swapping out a list of archetypes for a single link. Feels like a loss of information in the details drawer.
> 
> The concern with the chips in drawer was scalability. Maybe we can find a happy medium to keep some sort of chip list / detail in drawer along with the link.

Sticking with using labels to display the archetypes, this looks promising: https://www.patternfly.org/components/label#label-group-with-overflow

ibolton336:
> > > All of the code changes look ok. Really my apprehension is with swapping out a list of archetypes for a single link. Feels like a loss of information in the details drawer.
> > 
> > 
> > The concern with the chips in drawer was scalability. Maybe we can find a happy medium to keep some sort of chip list / detail in drawer along with the link.
> 
> Sticking with using labels to display the archetypes, this looks promising: https://www.patternfly.org/components/label#label-group-with-overflow

@JustinXHale Looking for input here. I think it was Ramon who requested the link instead of chips here since there may be 100s of tags. Even with the overflow, that might be asking too much of our side drawer to render all of those chips in a readable manner.

JustinXHale:
> > > > All of the code changes look ok. Really my apprehension is with swapping out a list of archetypes for a single link. Feels like a loss of information in the details drawer.
> > > 
> > > 
> > > The concern with the chips in drawer was scalability. Maybe we can find a happy medium to keep some sort of chip list / detail in drawer along with the link.
> > 
> > 
> > Sticking with using labels to display the archetypes, this looks promising: https://www.patternfly.org/components/label#label-group-with-overflow
> 
> @JustinXHale Looking for input here. I think it was Ramon who requested the link instead of chips here since there may be 100s of tags. Even with the overflow, that might be asking too much of our side drawer to render all of those chips in a readable manner.

Give me a couple days to look into this. I need to pay with some designs and see how we can take scalability into account when there can be a 100 tags

sjd78:
> > @JustinXHale Looking for input here. I think it was Ramon who requested the link instead of chips here since there may be 100s of tags. Even with the overflow, that might be asking too much of our side drawer to render all of those chips in a readable manner.
> 
> Give me a couple days to look into this. I need to pay with some designs and see how we can take scalability into account when there can be a 100 tags

Specifically for this PR, it is considering the display of the archetypes that an application is associated to.  I would be surprised if there are 100 different archetypes for an application, even if the number isn't really limited.  The drawer has 3 things for the application archetypes:
  - the associated list
  - from the associated list, which ones are assessed
  - from the associated list, which ones are reviewed

Maybe a table layout in another tab like we have in dependency detail drawer?

Maybe it is worth thinking about how similar the application drawer's details/archetypes section and tags tabs should look?

cc: @JustinXHale

ibolton336:
> > > @JustinXHale Looking for input here. I think it was Ramon who requested the link instead of chips here since there may be 100s of tags. Even with the overflow, that might be asking too much of our side drawer to render all of those chips in a readable manner.
> > 
> > 
> > Give me a couple days to look into this. I need to pay with some designs and see how we can take scalability into account when there can be a 100 tags
> 
> Specifically for this PR, it is considering the display of the archetypes that an application is associated to. I would be surprised if there are 100 different archetypes for an application, even if the number isn't really limited. The drawer has 3 things for the application archetypes:
> 
> * the associated list
> * from the associated list, which ones are assessed
> * from the associated list, which ones are reviewed
> 

Wondering if we should consider the inverse case - applications for a given archetype being > 100 applications:

Screenshot 2024-02-19 at 4 05 50 PM
Screenshot 2024-02-19 at 4 02 04 PM


> Maybe a table layout in another tab like we have in dependency detail drawer?


Like this? 

Screenshot 2024-02-19 at 4 18 03 PM

> Maybe it is worth thinking about how similar the application drawer's details/archetypes section and tags tabs should look?

Yeah I am not sure what we expect to see for tags. Currently we have this layout which already seems a bit stretched for an app with 40 tags:
Screenshot 2024-02-19 at 4 20 58 PM
Screenshot 2024-02-19 at 4 20 53 PM

> 
> cc: @JustinXHale

JustinXHale:
@ibolton336 i think a label group is great. that was my first thought, just wasnt sure if it would work if its 100 labels. But if that wont be an issue, i think the pattern is perfectly acceptable. We probably need to think about after how many labels to we want to show the more tag, i'm thinking 5.

ibolton336:
Latest changes move the code for link instead of label list from the app drawer to the archetype drawer as outlined here: 
https://issues.redhat.com/browse/MTA-2283

djzager (approved):
ack

PR #1683, opened 2024-02-12
:bug: Remove unneccessary metadata

ibolton336:
https://issues.redhat.com/browse/MTA-1721


sjd78 (approved):
LGTM

Although, if too many more things need to be removed, it may be worth converting to an allow list instead of a deny list.  Doing data transforms is always tricky.

PR #1687, opened 2024-02-15
:bug: Changed column name to 'Total Effort'

mguetta1:
Resolves: https://issues.redhat.com/browse/MTA-1970


sjd78 (commented):
The change LGTM, but it appears to break one of the e2e UI tests.  Is there a matching PR in the UI tests to adjust to this change?

ibolton336 (approved):

ibolton336:
Since this is not a blocker, I think we need to wait to merge it. Thanks.

PR #1690, opened 2024-02-16
:bug: Extend notification delay & fix broken delay customization

ibolton336:
https://issues.redhat.com/browse/MTA-2260
- Previously, the notification delay timeout was being overridden by the timeout set on the alert component itself. This fixes this accidental override.
- Lengthens timeout for notification clear globally to 8 seconds.


djzager (approved):
lgtm

PR #1693, opened 2024-02-21
:bug: application can only have one review

djzager:
https://issues.redhat.com/browse/MTA-1888

It appears that an application can only have a single review. So this changes the Tab title from `Reviews` -> `Review` and drops the redundant header inside.

### DIFF
--- a/client/src/app/pages/applications/components/application-detail-drawer/application-detail-drawer.tsx
+++ b/client/src/app/pages/applications/components/application-detail-drawer/application-detail-drawer.tsx
@@ -425,7 +425,7 @@ export const ApplicationDetailDrawer: React.FC<
           </Tab>
           <Tab
             eventKey={TabKey.Reviews}
-            title={<TabTitleText>{t("terms.reviews")}</TabTitleText>}
+            title={<TabTitleText>{t("terms.review")}</TabTitleText>}
           >
             <ReviewFields application={application} />
           </Tab>
--- a/client/src/app/pages/applications/components/application-detail-drawer/review-fields.tsx
+++ b/client/src/app/pages/applications/components/application-detail-drawer/review-fields.tsx
@@ -4,11 +4,8 @@ import {
   DescriptionListGroup,
   DescriptionListTerm,
   DescriptionListDescription,
-  Title,
-  TextContent,
 } from "@patternfly/react-core";
 import { Application, Archetype, Review } from "@app/api/models";
-import spacing from "@patternfly/react-styles/css/utilities/Spacing/spacing";
 import { useFetchReviewById, useFetchReviews } from "@app/queries/reviews";
 import { useFetchArchetypes } from "@app/queries/archetypes";
 import { EmptyTextMessage } from "@app/components/EmptyTextMessage";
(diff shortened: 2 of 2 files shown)


djzager:
Screenshot 2024-02-21 at 14 18 44

client/src/app/pages/applications/components/application-detail-drawer/application-detail-drawer.tsx
```diff
@@ -425,7 +425,7 @@ export const ApplicationDetailDrawer: React.FC<
           </Tab>
           <Tab
             eventKey={TabKey.Reviews}
-            title={<TabTitleText>{t("terms.reviews")}</TabTitleText>}
+            title={<TabTitleText>{t("terms.review")}</TabTitleText>}
```
mguetta1, line 428: Should be **reviews**.
An application can have more than one review inherited from multiple archetypes
app_review_mta-7 0 0-50

djzager: Then what exactly was meant by the [comment](https://issues.redhat.com/browse/MTA-1888?focusedId=23739303&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-23739303):
> Scott Dickerson please see attached [app_review_mta-7.0.0-50.png](https://issues.redhat.com/secure/attachment/13111122/13111122_app_review_mta-7.0.0-50.png), I think the best would be to title the tab in application side drawer "Review" as suggested in the bug summary.
Thanks

Based on this comment. The bug is fixed and should have been closed.

djzager: I keep reading and re-reading trying to understand what I'm missing. My apologies. These look like two comments in direct opposition to each other.

One says, `I think the best would be to title the tab in the application side drawer "Review"...` and the comment made on the line that changes it to `"Review"` is `Should be **reviews**`.

ibolton336: @JustinXHale 

Question for you here:
When an application has a review at the app level, we only show the app review in the drawer as expected / discussed.
Screenshot 2024-02-21 at 3 39 32 PM

Then when we discard the review for the app, we show the inherited archetype reviews. 

Screenshot 2024-02-21 at 3 39 39 PM

Screenshot 2024-02-21 at 3 39 50 PM

Want to figure out a better way to represent this in the UI. Any ideas here? There is some desire to change the review tab to read reviews / review depending on which scenario we are in. Wondering your take on this. Thanks!

mguetta1: You are so right! I confused myself...
I forgot that I asked to change the tab title to "Review" and the initial purpose of the bug

In bottom line, application can have one or more reviews, so Review/Reviews are both acceptable as the tab title :)
The original bug was to make the term "Review" in both application and archetype drawer side, so LGTM

Sorry about that.

JustinXHale: @ibolton336 I think Review or Reviews is appropriate. I learn more to 'Reviews' since applications and archetypes can have more than one review. But many applications might only have one 'Review' so review sits just fine as well, and if its more than one its no big deal. 

If the majority are okay with 'Review', which that seems to be the case, I'm good with 'Review' as well.

ibolton336: 👍 Thanks @JustinXHale

ibolton336 (approved):
LGTM :)

PR #1705, opened 2024-02-27
:bug: Add tooltip for migration wave delete option in app table

ibolton336:
Resolves https://issues.redhat.com/browse/MTA-1958

- Also adds translation entry for other similar message in toolbar


sjd78 (approved):
LGTM

PR #1707, opened 2024-02-27
:bug: Handle case when application.archetypes is passed but null

ibolton336:
Resolves https://issues.redhat.com/browse/MTA-2320

- Missed a case when calculating associated assessed archetypes to an application. If application is passed to the useFetchArchetypes query but application.archetypes is null, still returning all archetypes instead of an empty list.


sjd78 (approved):
LGTM

PR #1711, opened 2024-02-28
:bug: Adjust risk values displayed based on questionnaire filter

ibolton336:
Resolves https://issues.redhat.com/browse/MTA-2093
- Uses the assessment risk when a specific questionnaire is selected in the landscape filter rather than the application risk.


client/src/app/pages/reports/components/application-landscape/application-landscape.tsx
```diff
@@ -112,15 +117,16 @@ export const ApplicationLandscape: React.FC<IApplicationLandscapeProps> = ({
 
   const { assessmentsWithArchetypeApplications } =
     useFetchAssessmentsWithArchetypeApplications();
+  console.log(assessmentsWithArchetypeApplications);
```
sjd78, line 120: No need for `console.log` here.

client/src/app/queries/assessments.ts
```diff
@@ -224,40 +223,39 @@ const removeSectionOrderFromQuestions = (
 export const useFetchAssessmentsWithArchetypeApplications = () => {
   const { assessments, isFetching: assessmentsLoading } = useFetchAssessments();
 
+  const archetypeQueries = assessments
+    .map((assessment) => assessment?.archetype?.id)
+    .filter(Boolean)
+    .map((archetypeId) => ({
```
sjd78, line 229: Are assessments a 1:1 relationship to archetypes?  If multiple assessments can be linked to the same archetype then this list could contain the same archetype id multiple times and have multiple requests for that archetype.  Probably wouldn't be a big issue, but can be avoided.

sjd78 (changes requested):
Need to remove a debugging console.log

dymurray (approved):

dymurray (approved):

sjd78 (approved):

PR #1727, opened 2024-03-01
:bug: Adds spinner for isLoading prop in asessment status component

ibolton336:
- Replaces erroneous not available status that is shown when no assessments are available on an app


sjd78 (approved):
LGTM

PR #1735, opened 2024-03-04
:bug: Remove unnecessary query invalidations for reviews

ibolton336:
https://issues.redhat.com/browse/MTA-2359

 - Removes query invalidation & additional archetypes / reviews api calls inside each instance of a rendered (but hidden) drawer. This was causing a huge performance hit. 

 - Adds margin at top of reviews tab
Screenshot 2024-03-04 at 2 46 40 PM


sjd78 (approved):
Change LGTM

Just thinking about it, I don't understand why a simple fetch of the applications or archetypes should result in the reviews being refetched.  That's a bit of a strange flow.  I wonder why it was that way in the first place.

PR #1746, opened 2024-03-06
:seedling: Increase ulimit nofiles on image-build.yaml

sjd78:
Builds keep failing with EMFILE errors, so follow the advice from a similar issue work around: https://github.com/npm/cli/issues/4783#issuecomment-1908016260

### DIFF
--- a/.github/workflows/image-build.yaml
+++ b/.github/workflows/image-build.yaml
@@ -21,6 +21,7 @@ jobs:
       image_name: "tackle2-ui"
       containerfile: "./Dockerfile"
       architectures: '[ "amd64", "arm64", "ppc64le", "s390x" ]'
+      extra-args: "--ulimit nofile=4906:4096"
     secrets:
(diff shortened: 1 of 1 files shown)


sjd78:
The multiarch build with this nofiles change works great on my fork: https://github.com/sjd78/tackle2-ui/actions/runs/8174501095

(or at least it does if npm install doesn't start timing out / get rate limited)

sjd78:
Another followup -- The `registry.access.redhat.com/ubi9/nodejs-18:latest` image was upgraded on March 5 to version `1-98`.  This includes an upgrade of npm from 9.8.1 to 10.2.3.  This changes appears to be the cause of the nofiles size error.

Running the build locally with podman will encounter the EMFILE error without adding `--ulimit=nofile=4096:4096` to the build command.

Running the build with the previous image release `:1-88` does not encounter the EMFILE error locally or in github actions.

Issue #1745, opened 2024-03-05
[BUG] Need to be able to filter & sort by application analysis status


ibolton336:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Konveyor version

7.0.2

### Priority

Critical

### Current Behavior

Currently there is no way to filter or sort the applications table by analysis status. 

<img width="985" alt="Screenshot 2024-03-05 at 3 44 25 PM" src="https://github.com/konveyor/tackle2-ui/assets/11218376/b9c72dad-e361-4a19-a06d-9626dabef8a6">
<img width="1591" alt="Screenshot 2024-03-05 at 3 44 20 PM" src="https://github.com/konveyor/tackle2-ui/assets/11218376/053d8e9b-41d1-4918-9824-4fc5603a9ad6">

### Expected Behavior

Need a way to filter by analysis status. 

### How Reproducible

Always (Default)

### Environment

```markdown
- OS:
```

TalyaNaima:
working on this issue, please assign it to me

HadasahR:
I'm also working on this issue, please assign it to me as well

PR #1752, opened 2024-03-08
:seedling: Refactor `ReviewFields`

sjd78:
Refactor the `ReviewFields` component in advance of #1745:

  - The component is used on the Review tab of application and archetype details drawer

  - The input prop `reviews` was not used in any current containing component.  Push the `useFetchReviews()` hook down to the component itself.  This keeps the data fetch as close as possible to where it is used.

  - Move the component to `@app/components/detail-drawer` folder since it is used in more than one page grouping


ibolton336 (approved):
Tested and LGTM.

Issue #1753, opened 2024-03-08
Overflow missing on application table status 


ibolton336:
<img width="1029" alt="Screenshot 2024-03-08 at 11 08 48 AM" src="https://github.com/konveyor/tackle2-ui/assets/11218376/71094e50-d12d-46cb-8f64-a0fda1005c5b">

sjd78:
**Note**: This issue may be taken care of once #1934 is complete.  Reevaluate after #1934 is done.

sjd78:
Still an issue after #1934 

Same issues as #1998 but for other columns

PR #1756, opened 2024-03-08
:bug: Reallocate col spacing & add overflow style for status

ibolton336:
Closes: https://github.com/konveyor/tackle2-ui/issues/1753

### DIFF
--- a/client/src/app/components/IconedStatus.tsx
+++ b/client/src/app/components/IconedStatus.tsx
@@ -1,3 +1,4 @@
+import "./iconed-status.css";
 import React from "react";
 import { Flex, FlexItem, Icon, Tooltip } from "@patternfly/react-core";
 import { useTranslation } from "react-i18next";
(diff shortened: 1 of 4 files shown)


rszwajko:
The text can get truncated too much i.e. to one letter:
image
At this point we would need a tooltip not to loose information - ideally visible only when the text overflows. The default table implementation works well  also for complex nodes but ...requires overflow. It would work for previous code (without fix) i.e.
image

sjd78 (commented):
Following up on Radek's comment and the slack chat -

The containing `Flex` in `IconedStatus` is set to "nowrap".  Maybe for the smaller size breakpoints (the `sm` and `md` sizes maybe?) it can be changed to the normal "wrap". That would let the icons and label wrap in their container.

For the label `FlexItem`, it may be worth setting a min-width so at least a few letters will always be visible.  The `ch` unit is "Relative to the width of "0" (zero)" (I did not know that until just now).  So maybe a `min-width: 10ch` or similar?

Issue #1646, opened 2023-12-27
[BUG] Assessment inherited from an archetype is "completed" instead of "in progress" while not all questions in questionnaire were answered


mguetta1:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Konveyor version

0.3.0-rc.3

### Priority

Major

### Current Behavior

When an application is associated with an archetype, assess an questionnaire from the archetype and **save as draft** => the application assessment on the "Application inventory" page is "Completed" instead of "In progress"  

### How Reproducible

Always (Default)

### Anything else?

The issue is not reproducible using downstream build: mta-operator-bundle-container-7.0.0-48

PR #1758, opened 2024-03-11
:ghost: Add unit tests and hook for useAssessmentStatus

ibolton336:
Inspired by latest QE issues surfacing around status regressions. 
Resolves https://issues.redhat.com/browse/MTA-2410
Resolves https://issues.redhat.com/browse/MTA-2409 
Resolves https://github.com/konveyor/tackle2-ui/issues/1646 

** Integrates MSW with unit test RTL framework to handle mock data without needing to add any extra stubs.


sjd78 (commented):
I like this approach to have a bunch of unit tests running on a hook!

The next step would be to use MSW in jest and let the data come from that code set.  That would even let us run the UI in MSW mode with the unit test data to live dev against!

sjd78:
Does this also cover #1646?

client/src/app/pages/applications/components/application-assessment-status/application-assessment-status.tsx
```diff
@@ -15,77 +16,11 @@ export const ApplicationAssessmentStatus: React.FC<
 > = ({ application, assessments, archetypes, isLoading }) => {
   const { t } = useTranslation();
 
-  const applicationAssessments = assessments?.filter(
-    (assessment: Assessment) => assessment.application?.id === application.id
+  const assessmentStatusInfo = useAssessmentStatus(
+    assessments,
+    archetypes,
```
sjd78, line 21: I wonder if it would be easier just to have `useAssessmentStatus` just pull the assessments and archetypes directly since I don't think they're being further processed beyond what their fetch hooks are doing.

In general I'm thinking a custom data hook approach of only providing the differentiating information to really focus on what the component cares about will be helpful.  In this case the component only cares about the application it needs to render something for.

sjd78 (changes requested):
I really like this approach and the unit tests.  Being able to test against different application+archetype+assessment/questionnaire shapes makes me happy.

Only one comment about providing "generic fetch data" to the custom hook.

client/src/app/pages/applications/components/application-assessment-status/application-assessment-status.tsx
ibolton336: Yeah good call. I made those changes.

client/src/app/hooks/useAssessmentStatus.ts
```diff
@@ -0,0 +1,77 @@
...
+import { useFetchArchetypes } from "@app/queries/archetypes";
+import { useFetchAssessments } from "@app/queries/assessments";
+import { useMemo } from "react";
+
+export const useAssessmentStatus = (application: Application) => {
+  const { assessments } = useFetchAssessments();
+  const { archetypes } = useFetchArchetypes();
+
+  return useMemo(() => {
+    const applicationAssessments = assessments?.filter(
```
rszwajko, line 12: `?.` suggest that `applicationAssessments` can be nulish later on - however we assume they are not.
Similar `inheritedArchetypes`

client/src/app/hooks/useAssessmentStatus.ts
```diff
@@ -0,0 +1,77 @@
...
+      false;
+
+    return {
+      allArchetypesAssessed,
+      countOfFullyAssessedArchetypes: assessedArchetypesCount,
+      countOfArchetypesWithRequiredAssessments: assessmentsFromArchetypesCount,
+      hasApplicationAssessmentInProgress,
+      isApplicationDirectlyAssessed: isDirectlyAssessed,
+    };
+  }, [assessments, archetypes, application]);
```
rszwajko, line 76: Right now the memo will get re-calculated with every react query re-fetch of all three entities involved (assuming that `application` reference is stable between re-fetch).
We could avoid re-calculating due to applications by  providing only `application.id`  and `isDirectlyAssessed` as a dependencies.

ibolton336: I'm ok with this - we do have 
eslint[react-hooks/exhaustive-deps](https://github.com/facebook/react/issues/14920)
turned on so wondering if we are all on board with an exception here.

ibolton336: On second thought.. the unit tests broke when I removed the other dependencies from the array.

sjd78: This seems to be one of the times where `useMemo` can be helpful if all the things don't change, and adds minimal overhead if it changes a bunch.  I'd keep it the way it is.

Also, I don't think the refetches will reset the returned data arrays on every fetch.  I seem to remember there is some caching built in and the fetched data != the returned data for this kind of reason.

sjd78 (changes requested):
The tests and the hook look pretty good!  MSW with the `server.use()` as part of the test is super nice!  I wonder if there is a nicer way to manage the input test data such that the objects can be defined more as a graph to keep it "simpler".  That'll just stick in the back of my head for a bit...

A few very minor change requests.

client/src/mocks/server.ts
```diff
@@ -0,0 +1,21 @@
...
+  (req, res, ctx) => {
+    return res(ctx.json({}));
+  }
+);
+const handlers = [
+  ...applications,
+  ...archetypes,
+  ...assessments,
+  localeHandler,
+].filter(Boolean);
```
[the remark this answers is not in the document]
ibolton336: Ended up creating handlers in the test for testing specific cases. Might be good as a follow up to iterate on this approach.

client/src/app/pages/applications/components/application-assessment-status/application-assessment-status.tsx
```diff
@@ -94,28 +20,23 @@ export const ApplicationAssessmentStatus: React.FC<
...
-    tooltipCount = assessedArchetypesWithARequiredAssessment.length;
-  } else if (
-    applicationAssessments?.some(
-      (assessment) => assessment.status === "started"
-    )
-  ) {
+    tooltipCount = countOfArchetypesWithRequiredAssessments;
+  } else if (hasApplicationAssessmentInProgress) {
     statusPreset = "InProgress";
   }
```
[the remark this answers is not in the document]
ibolton336: I'd like to revisit this further as well. Maybe in a future PR since this is stable & fixes the bugs referenced.

sjd78 (approved):
LGTM!

I'm excited now to be able to unit test data hooks following this pattern.

rszwajko (approved):
Looks OK to me!

PR #1761, opened 2024-03-12
:bug: Wrap long file paths in issues drawer

ibolton336:
Resolves https://issues.redhat.com/browse/MTA-1961?filter=12422675


sjd78 (commented):
The full file path text is displayed but now the drawer requires horizontal scrolling:
screenshot-localhost_9000-2024 03 12-17_10_51

Maybe there is a way to have the text wrap at directory delineation?

TIL about `<wbr />`: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/wbr

Not safe but works as a POC:
```tsx
                      <div
                        dangerouslySetInnerHTML={{ __html:
                          fileReport?.file.split("/").join("/<wbr />") || "<unknown>"
                        }}
                      />
```

client/src/app/pages/issues/issue-detail-drawer/path-display.tsx
```diff
@@ -0,0 +1,22 @@
+import React from "react";
+
+interface PathDisplayProps {
+  path: string;
+}
+
+const PathDisplay: React.FC<PathDisplayProps> = ({ path }) => {
+  const formattedPath = path
+    .split("/")
```
sjd78, line 9: My only concern here is if the path will ever has a windows based backslash `\` separator.  I don't know if that is even something that will happen...

sjd78 (commented):
Setting up the `PathDisplay` looks nice.  Only question to about the path separator character.

sjd78 (approved):
After thinking about edge cases for path strings, I think this should work out well.

The edge case I got hung up on was, "what about if the path has an escaped separator character".  But then 2 things.  First, it only give text wrapping boundaries, so not a big impact.  Second, I don't think you can have an escaped separator character in a full file name.

```bash
> touch "Hello\\There.txt"
> touch "Hello\/There.txt"
touch: cannot touch 'Hello\/There.txt': No such file or directory
```

The first one works file on linux.  The second one does not.  Edge case solved!

PR #1767, opened 2024-03-13
:sparkles: integrate column management feature into apps table

### LINKED ISSUE #1765: Integrate col management to applications table
https://www.patternfly.org/components/table/react-demos/#column-management-with-draggable 

- Make use of the pf column mgmt functionality to alleviate space concerns with multiple status fields containing 2 icons each. This will allow the user to opt in / out of specific fields based on their interests.

ibolton336:
Resolves #1765


rszwajko (commented):
Some comments below (based on experience gathered from https://github.com/kubev2v/forklift-console-plugin/pull/64 and follow-ups)

client/src/app/hooks/table-controls/column/useColumnState.ts
```diff
@@ -0,0 +1,60 @@
...
+  const { persistTo = "state", persistenceKeyPrefix, initialColumns } = args;
+
+  const [columns, setColumns] = usePersistentState<
+    ColumnState<TColumnKey>[],
+    TPersistenceKeyPrefix,
+    "columns"
+  >({
+    defaultValue: initialColumns,
+    persistenceKeyPrefix,
+    ...(persistTo === "urlParams"
```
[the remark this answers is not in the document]
mturley: +1, it also wouldn't stick around if you navigate away from the page and back via the menu. probably needs to always be in localStorage or perhaps even support being lifted out and stored in the backend somewhere for the user profile

ibolton336:
https://github.com/konveyor/tackle2-ui/assets/11218376/4e090048-1c64-47cb-827b-5e6d12c1b01d

client/src/app/hooks/table-controls/useTableControlProps.ts
```diff
@@ -171,6 +172,18 @@ export const useTableControlProps = <
     },
   });
 
+  const getColumnProps = (columnKey: string) => {
+    const column = columns.find((c) => c.id === columnKey);
+    return {
+      isVisible: column?.isVisible,
+      order: column?.order,
+      // Any additional props or handlers for managing columns
```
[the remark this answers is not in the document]
ibolton336: Going to put a pin in the reorder bit for now.

client/src/app/pages/applications/applications-table/applications-table.tsx
```diff
@@ -826,16 +833,31 @@ export const ApplicationsTable: React.FC = () => {
...
-                  {...getThProps({ columnKey: "businessService" })}
-                  width={15}
-                />
-                <Th {...getThProps({ columnKey: "assessment" })} width={10} />
-                <Th {...getThProps({ columnKey: "review" })} width={10} />
-                <Th {...getThProps({ columnKey: "analysis" })} width={10} />
-                <Th {...getThProps({ columnKey: "tags" })} width={10} />
-                <Th {...getThProps({ columnKey: "effort" })} width={10} />
+                {/* {columnState.columns. isVisible then render this column} */}
+                {getColumnVisibility("name") && (
```
sjd78, line 837: I don't mind this approach.

The `visibility` prop exists and is used in the [PF docs](https://www.patternfly.org/components/table#cell-width-breakpoint-modifiers) to use the css breakpoints to show/hide columns.  Fun fact is that prop is not in any of the generated docs and it needs to be tracked down in sources.  I tried that and got lost pretty quick.  Probably isn't worth trying to use `visibility` for the user based show/hide column support.

client/src/app/hooks/table-controls/column/useColumnState.ts
```diff
@@ -0,0 +1,25 @@
...
+  initialColumns: ColumnState<TColumnKey>[];
+}
+
+export const useColumnState = <TColumnKey extends string>(
+  args: IColumnStateArgs<TColumnKey>
+): IColumnState<TColumnKey> => {
+  const [columns, setColumns] = useState<ColumnState<TColumnKey>[]>(
+    args.initialColumns
+  );
+  return { columns, setColumns };
```
sjd78, line 24: And this is where you'd have to add column visibility persistence past the component state level that is used now.  I predict there will be a QE ticket filed in less than 10 minutes of app testing say that the column visibility doesn't persist page page refreshes.

A chat before testing would help.  Maybe even loop docs in to level set expectations as to how long the column management state lives.

sjd78 (approved):
Overall the change LGTM.  A few comments, and only 1 console.log that should be taken out.

sjd78 (changes requested):
Now that I'm running it to play around, I have 2 comments, one of which needs to be resolved for sure:
  1. The column name in the list is not center aligned (needs to be fixed)
  2. I highly suggest that the Name column is setup to be required.  It can show up on the list, and the checkbox checked, but the checkbox being disabled.

Screen shot from FF:
image

client/src/app/pages/applications/applications-table/components/manage-columns-toolbar.tsx
```diff
@@ -0,0 +1,51 @@
...
+  const { t } = useTranslation();
+  const [isOpen, setIsOpen] = useState(false);
+  return (
+    <>
+      <ToolbarItem variant="overflow-menu">
+        <OverflowMenu breakpoint="md">
+          <OverflowMenuGroup groupType="button" isPersistent>
+            <OverflowMenuItem isPersistent>
+              <Button variant="link" onClick={() => setIsOpen(true)}>
+                Manage columns
```
rszwajko, line 32: how about an icon? there is a dedicated ColumnsIcon -> https://www.patternfly.org/design-foundations/icons/#all-icons
>Indicates the ability to manage columns for a table view
>

client/src/app/pages/applications/applications-table/components/manage-columns-toolbar.tsx
```diff
@@ -0,0 +1,51 @@
...
+              </Button>
+            </OverflowMenuItem>
+          </OverflowMenuGroup>
+        </OverflowMenu>
+      </ToolbarItem>
+
+      <ManageColumnsModal
```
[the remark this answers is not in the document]
sjd78: But since the modal is rendered on the table page for the page's configuration I'm not even sure what the state reset is for.  Adding the column management to any other table page will require the modal to be added there as well. :thinking: 

That doesn't invalidate the technique, it just make me curious.

client/src/app/hooks/table-controls/column/useColumnState.ts
ibolton336: Added local storage support for the column names.

client/src/app/pages/applications/applications-table/components/manage-columns-toolbar.tsx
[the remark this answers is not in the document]
rszwajko: the modal keeps tracks of the columns in the internal state i.e.
```
const [editedColumns, setEditedColumns] =
    useState<ColumnState<TColumnKey>[]>(columns);
```
Most of the times this should be the same as the external state (maintained by the `useLocalTableControls()`) but it's safer to keep one source of truth.

ibolton336:
Screenshot 2024-03-25 at 5 09 46 PM
Updated to match pr suggestions / UXD suggestions.

client/src/app/pages/applications/applications-table/components/manage-columns-toolbar.tsx
JustinXHale: > how about an icon? there is a dedicated ColumnsIcon -> https://www.patternfly.org/design-foundations/icons/#all-icons
> 
> > Indicates the ability to manage columns for a table view

i’m slightly indifferent. I think the manage column icon is fine, the only issue for me is our users wont know what it is bc they havent used it. 

updated: I'm in a UX Face 2 Face... general consensus is use the icon to save space

client/src/app/pages/applications/applications-table/components/manage-columns-modal.tsx
```diff
@@ -0,0 +1,103 @@
...
+    >
+      <DataList aria-label={title} id="table-column-management" isCompact>
+        {editedColumns.map(({ id, label, isVisible }, index) => (
+          <DataListItem key={index}>
+            <DataListItemRow className="custom-data-list-item-row">
+              <DataListControl>
+                <DataListCheck
+                  aria-labelledby={`check-${id}`}
+                  checked={isVisible}
+                  isDisabled={id === "name"}
```
sjd78, line 85: Humm....this feels a bit awkward.

I'd rather see a flag of some kind be passed along with the columns as a way to decided what columns are required.  That could be left for a future enhancement as long as some comments are added along the lines of "any column whose name is 'name' will not be allowed to be hidden".

client/src/app/pages/applications/applications-table/components/manage-columns-toolbar.tsx
```diff
@@ -0,0 +1,53 @@
...
+      {isOpen && (
+        <ManageColumnsModal
+          onClose={() => setIsOpen(false)}
+          description={t("Selected columns will be displayed in the table.")}
+          setColumns={setColumns}
+          columns={columns}
+          saveLabel={t("Save")}
+          cancelLabel={t("Cancel")}
+          title={t("Manage Columns")}
+        />
```
sjd78, line 49: As @rszwajko mentioned offline, the `t()` calls should be keys with the text added in the translations file to be consistent with the rest of the i18next usage

sjd78 (commented):
Looks really good.  The only highly recommended change would be for the i18next strings in the `ManageColumnsToolbar`.

rszwajko (approved):
Looks OK to me. Any open points can be addressed in follow-ups.

sjd78 (approved):
LGTM

Issue #1754, opened 2024-03-08
Missing application selection values in issues table filter


ibolton336:
<img width="811" alt="Screenshot 2024-03-08 at 11 11 57 AM" src="https://github.com/konveyor/tackle2-ui/assets/11218376/b0cbd0dc-698e-42c7-aec4-7d7e4e57f0d9">

rszwajko:
@ibolton336 
It seems that we are using there a server side `search` filter - see https://github.com/konveyor/tackle2-ui/blob/c2ef03c0293eac23d3e38e29caaf28a83de943a4/client/src/app/pages/issues/helpers.ts#L50

If I understand the issue correctly, we would like to use  `multiselect` filtering (choosing apps from the list).  However this assumes that we know the names already. Do we need to fetch available apps first?

PR #1771, opened 2024-03-13
:seedling: Use multi select filter for application names

rszwajko:
Resolves: #1754


rszwajko:
### New filter
Screenshot from 2024-03-13 15-50-47

ibolton336 (approved):
Tested & LGTM - nice touch with the sort as well.

Issue #1708, opened 2024-02-27
[BUG] Scrollable Modal - Visible Primary and Secondary Button


JustinXHale:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Konveyor version

Feb 27, 2024

### Priority

Minor

### Current Behavior

When scrolling in a modal the Primary and Secondary action buttons should always be visible. 
![image](https://github.com/konveyor/tackle2-ui/assets/91554933/0842af22-4554-4b38-83f3-4d2ed9900071)

### Expected Behavior

Primary & Secondary buttons should be visible. [PF Example](https://www.patternfly.org/components/modal/#scrollable-modals)

<img width="624" alt="image" src="https://github.com/konveyor/tackle2-ui/assets/91554933/f750bbef-7ecf-43d2-95aa-5a9271a0a21a">

### How Reproducible

Always (Default)

### Steps To Reproduce

Will apply to ever modal

Modal example
1. Application inventory
2. Create new

### Environment

```markdown
- OS:
```

rszwajko:
@JustinXHale
We have a lot of content inside this modal - it is possible to just pin the buttons as you requested but perhaps it's time to move it to a subpage?
 Quick draft looks the following: [Screencast from 2024-03-13 20-06-35.webm](https://github.com/konveyor/tackle2-ui/assets/64194103/1fd535ec-9b28-44f9-929f-079654b45698)

JustinXHale:
> @JustinXHale We have a lot of content inside this modal - it is possible to just pin the buttons as you requested but perhaps it's time to move it to a subpage? Quick draft looks the following: [Screencast from 2024-03-13 20-06-35.webm](https://github.com/konveyor/tackle2-ui/assets/64194103/1fd535ec-9b28-44f9-929f-079654b45698)

you make a good point. I think this is something we need to look at across the product to make sure everything aligns with the interactions the user expects across the board. Part of me says your right, the other half says though its somewhat long a full page isnt needed b/c the user doesnt actually have to fill in all that information. This is definitely something I will look at as I continue the audit

PR #1776, opened 2024-03-14
:bug: Decouple form actions from ApplicationForm

rszwajko:
Split ApplicationForm into a data hook and view component. Use extracted submit and cancel actions inside Modal's footer.

Resolves: #1708


rszwajko:
### After

image

client/src/app/pages/applications/components/application-form/application-form.tsx
```diff
@@ -623,27 +678,29 @@ export const ApplicationForm: React.FC<ApplicationFormProps> = ({
...
-          id="cancel"
-          aria-label="cancel"
-          variant={ButtonVariant.link}
-          isDisabled={isSubmitting || isValidating}
-          onClick={onClose}
-        >
-          {t("actions.cancel")}
-        </Button>
-      </ActionGroup>
+      {showFormButtons && (
```
ibolton336, line 681: Can we just remove this code since the buttons live in the modal now?

rszwajko: yes. I was not sure what's the plan for this form so I've kept the refactoring minimal.

ibolton336 (approved):
LGTM

client/src/app/pages/applications/components/application-form/application-form.tsx
ibolton336: If you want to remove the buttons that is fine by me. I appreciate you making the effort to minimize changes.

client/src/app/pages/applications/components/application-form/application-form-modal.tsx
```diff
@@ -0,0 +1,54 @@
...
+          key="cancel"
+          id="cancel"
+          aria-label="cancel"
+          variant={ButtonVariant.link}
+          isDisabled={formProps.isCancelDisabled}
+          onClick={onClose}
+        >
+          {t("actions.cancel")}
+        </Button>,
+      ]}
```
sjd78, line 49: If you want to keep the buttons with the Form itself, you could use a portal to a footer element.  This would follow the general pattern of setting a Modal's render container or some of the toolbar filter stuff.

Something like:
  - Setup a useRef() to a footer container
  - Pass that ref as the render container for the form buttons to the form component
  - In the form component, if a button container is passed, `createPortal` to it, otherwise render in place
  - "Bob's your uncle"

sjd78 (commented):
Looks ok as-is, but just a thought about how you could manage the buttons from the form itself.  Not a blocker.

client/src/app/pages/applications/components/application-form/application-form-modal.tsx
rszwajko: @sjd78 @ibolton336 
hmm ...portals...  we could use them but this would limit the functionality  i.e. what if we would like to trigger submit from 2 places? or from some menu, dropdown,, etc ?
But if you think it's the way to go then I can refactor the code in that direction.

sjd78: The relationship between the hook and the form component doesn't need to change.  Using a portal would just let the buttons live inside the form component but be rendered elsewhere.

sjd78 (approved):
After looking at what the createPortal-to-a-ref-in-the-Modal-footer would look like, and lots of discussion about it, the current setup is ok with me.

Going forward, we should think about if other tweaks to this approach would be good to apply across all form modals.  I'm sure this will not be the last "keep the buttons sticky to the bottom" kind of request we will encounter.

rszwajko:
For reference: version that uses portal to display buttons in the parent modal is available here ->  https://github.com/rszwajko/tackle2-ui/commit/ec93ae726822132cbddca4f67a3a538fb9f2f8fb

client/src/app/pages/applications/components/application-form/application-form.tsx
rszwajko: Removed the buttons as suggested.

PR #1780, opened 2024-03-15
:seedling: Refactor CI github actions, Dockerfile

sjd78:
Refactoring:
  - Split CI, one for repo tests and one for global tests

  - Only enable global CI on main (with a note)

  - Global CI configured like the multiarch build to cancel in progress runs in a new run starts

  - Repo CI bump the version of action `codecov/codecov-action` based on runtime warnings

  - Add `--ignoreScripts` to the `npm clean-install` in the Dockerfile to avoid running unnecessary postinstall scripts with the build running directly after the install.


rszwajko:
Looks OK to me.

djzager (approved):
visack

PR #1781, opened 2024-03-16
:ghost: force npm9 during image build

sjd78:
Ever since the build container we use [1][2] updated to use `npm@10`, the image builder github action has encountered a number of problems. The first was solved with #1746. The second more random problem has been with network connections erroring out when installing dependencies.

To work around the network and nofiles errors:
  1. enforce the use of `npm@9`
  2. pin the Dockerfile base images to specific version tags builder at `:1-88`, and runtime at `:1-93`

Additional solutions:
  - Add the force to `npm@9` directly in the Dockerfile for use in all circumstances
  - Find a way to use something like verdaccio to proxy/mirror the npmjs repo during the build so npmjs fetches can be pooled across the parallel image builds

[1] registry.access.redhat.com/ubi9/nodejs-18:latest
[2] https://catalog.redhat.com/software/containers/ubi9/nodejs-18/62e8e7ed22d1d3c2dfe2ca01


.github/workflows/image-build.yaml
```diff
@@ -22,6 +22,8 @@ jobs:
       containerfile: "./Dockerfile"
       architectures: '[ "amd64", "arm64", "ppc64le", "s390x" ]'
       extra-args: "--ulimit nofile=4096:4096"
+      pre_build_cmd: |
```
rszwajko, line 25: How about adding a new env variable i.e. FORCE_NPM_9 and then explicitly branching based on that?
The `sed` is more elegant but is not visible while looking at the commands  list.
Alternatively we could use some alarmist comment above the `clean-install` command ("Attention! Attention!....")

sjd78: Point taken.  I've made some more obvious changes to force `npm@^9`.

rszwajko (approved):
Looks explicit enough for me:)

ibolton336:
Looks like this is causing NPM install to fail locally. Error fixed by moving back to 
`    "npm": "^9.5.0"` in package.json

Error:
`➜  tackle2-ui git:(main) npm i
npm ERR! code EBADENGINE
npm ERR! engine Unsupported engine
npm ERR! engine Not compatible with your version of node/npm: @konveyor-ui/root@2.1.0
npm ERR! notsup Not compatible with your version of node/npm: @konveyor-ui/root@2.1.0
npm ERR! notsup Required: {"node":">=18.14.2","npm":"^9.5.0"}
npm ERR! notsup Actual:   {"npm":"10.2.0","node":"v21.1.0"}

npm ERR! A complete log of this run can be found in: /Users/ibolton/.npm/_logs/2024-03-25T13_35_37_824Z-debug-0.log
➜  tackle2-ui git:(main)`

ibolton336:
This was due to my local npm version falling outside the supported version range in package.json. Fixed with:
`npm install -g npm@9.5.0
`

Issue #1709, opened 2024-02-27
[Enhancement] Assessment wizard - Skip


JustinXHale:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Konveyor version

Feb 27, 2024

### Priority

Minor

### Current Behavior

When taking an assessment the user is unable to view questions in other sections. 
User are unable to proceed past step 1 of the assessment a though the questions are not required.

![image](https://github.com/konveyor/tackle2-ui/assets/91554933/fe72e2a4-2e54-400f-87ca-e49b106cf0fe)

### Expected Behavior

- Select stakeholders questions should be required or user should be able to bypass the question
- Users should be allowed to view other sections before selecting all responses

### How Reproducible

Always (Default)

### Steps To Reproduce

Open assessment wizard

### Environment

```markdown
- OS:
```

PR #1799, opened 2024-03-27
:sparkles: Reduce form fields in assessment wizard stakeholder page to a single dropdown

ibolton336:
Waiting for https://github.com/konveyor/tackle-ui-tests/pull/1092
UI tests PR: 1092

Resolves: https://github.com/konveyor/tackle2-ui/issues/1709

https://github.com/konveyor/tackle2-ui/assets/11218376/9969be00-dad7-4b3d-920e-27be7e3218c7

### DIFF
--- a/client/src/app/components/Autocomplete.tsx
+++ b/client/src/app/components/Autocomplete.tsx
@@ -18,6 +18,8 @@ import {
 const toString = (input: string | (() => string)) =>
   typeof input === "function" ? input() : input;
 
+const createCompositeKey = (group: string, id: number) => `${group}:${id}`;
+
 export interface AutocompleteOptionProps {
   /** id for the option */
   id: number;
@@ -72,7 +74,8 @@ export const Autocomplete: React.FC<IAutocompleteProps> = ({
   noResultsMessage = "No results found",
 }) => {
   const [inputValue, setInputValue] = useState(searchString);
-  const [tabSelectedItemId, setTabSelectedItemId] = useState<number>();
+  const [tabSelectedItemId, setTabSelectedItemId] = useState<string>();
+
   const [menuIsOpen, setMenuIsOpen] = useState(false);
 
   /** refs used to detect when clicks occur inside vs outside of the textInputGroup and menu popper */
@@ -83,22 +86,40 @@ export const Autocomplete: React.FC<IAutocompleteProps> = ({
     if (!selections || selections.length === 0) {
       return [];
     }
-    return options.filter(
-      ({ id }) => selections.findIndex((s) => s.id === id) > -1
-    );
-  }, [options, selections]);
+    return isGrouped
+      ? options.filter((option) => {
+          return selections.some((selection) => {
+            return (
+              selection.id === option.id && selection.group === option.group
+            );
+          });
+        })
+      : options.filter((option) => {
+          return (
+            selections.findIndex((selection) => selection.id === option.id) > -1
+          );
+        });
+  }, [options, selections, isGrouped]);
 
   const filteredOptions = useMemo(() => {
-    return options.filter(
-      ({ id, name }) =>
-        selections.findIndex((s) => s.id === id) === -1 &&
-        toString(name).toLowerCase().includes(inputValue.toLocaleLowerCase())
-    );
-  }, [options, selections, inputValue]);
+    return options.filter((option) => {
+      const isOptionSelected = selections.some((selection) => {
+        const isSelectedById = selection.id === option.id;
+        const isSelectedByGroup = isGrouped
+          ? selection.group === option.group
+          : true;
+        return isSelectedById && isSelectedByGroup;
+      });
+
+      const isNameMatch = toString(option.name)
+        .toLowerCase()
+        .includes(inputValue.toLowerCase());
+      return !isOptionSelected && isNameMatch;
+    });
+  }, [options, selections, inputValue, isGrouped]);
 
   const groupedOptions = useMemo((): GroupedOptions => {
     if (!isGrouped) {
-      // If not grouped, return an empty object or handle accordingly
       return {};
     }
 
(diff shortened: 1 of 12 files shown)


client/src/app/components/Autocomplete.tsx
```diff
@@ -12,11 +12,14 @@ import {
   SearchInput,
   Divider,
   Tooltip,
+  MenuGroup,
 } from "@patternfly/react-core";
 
 const toString = (input: string | (() => string)) =>
   typeof input === "function" ? input() : input;
 
+const createCompositeKey = (group: string, id: number) => `${group}:${id}`;
```
rszwajko, line 21: 1. Composite key triggered many changes in the component. Can we generate the IDs outside i.e. in `combineAndGroupStakeholderRefs`? This way the component will be able to re-use the same routines as for plain list.
2. we can generate the ID in any way we want (index in the array, random value, etc)  so joining group and id is safe. However the parsing is the composite key to get group and ID is risky. I'd rather use a lookup dictionary.

client/src/app/components/Autocomplete.tsx
```diff
@@ -256,9 +331,48 @@ export const Autocomplete: React.FC<IAutocompleteProps> = ({
...
   );
 
-  const menu = (
-    <Menu ref={menuRef} onKeyDown={handleMenuOnKeyDown} isScrollable>
-      <MenuContent>
+  const renderMenuItems = () => {
+    if (isGrouped && groupedOptions) {
+      return Object.entries(groupedOptions).map(([groupName, groupOptions]) => (
+        <React.Fragment key={groupName}>
+          <MenuGroup label={groupName}>
```
rszwajko, line 338: [Looking at the PF example](https://www.patternfly.org/components/menus/menu#titled-groups-of-items) a `MenuGroup` without a label is effectively the same as no group.
Perhaps we could avoid branching here by treating `isGrouped== false` as a group with no name (`groupName == undefined`)

rszwajko (commented):
Adding grouping to `Autocomplete` required massive changes. I feel we need to revisit the approach(some ideas in comments below). Alternatively we could break the component into parts and used them to build 2 specialized components i.e. GroupedAutocomplete and StandardAutocomplete.

client/src/app/components/Autocomplete.tsx
```diff
@@ -74,32 +87,82 @@ export const Autocomplete: React.FC<IAutocompleteProps> = ({
...
+    });
+  }, [options, selections, inputValue, isGrouped]);
+
+  const groupedOptions = useMemo((): GroupedOptions => {
+    if (!isGrouped) {
+      return {};
+    }
+
+    return filteredOptions.reduce((groups: GroupedOptions, option) => {
+      const groupName = option.group || "Ungrouped";
```
rszwajko, line 128: `Ungrouped` is an unlikely group label but I would still avoid hardcoding.
Some ideas:
1. use empty string ?
2. drop the items without the group - they would need to be handled separately

client/src/app/components/Autocomplete.tsx
```diff
@@ -256,42 +323,41 @@ export const Autocomplete: React.FC<IAutocompleteProps> = ({
...
 
+  const renderMenuItems = () => {
+    const effectiveGroupedOptions = isGrouped
+      ? groupedOptions
+      : { undefined: filteredOptions };
+
+    return Object.entries(effectiveGroupedOptions).map(
+      ([groupName, groupOptions], index) => (
+        <React.Fragment key={groupName || `ungrouped-${index}`}>
+          <MenuGroup label={groupName}>
```
rszwajko, line 334: 1. assuming that the "groupedOptions" contain the "Ungrouped" under empty string we can just ` label={groupName || undefined}`
2. we should actually first try the group label and fallback to the group name

client/src/app/components/Autocomplete.tsx
```diff
@@ -256,42 +323,41 @@ export const Autocomplete: React.FC<IAutocompleteProps> = ({
...
+            <MenuList>
+              {groupOptions.map((option) => (
+                <MenuItem
+                  key={option.uniqueId || option.id}
+                  itemId={option.id.toString()}
+                  onClick={(e) => handleMenuItemOnSelect(e, option)}
+                >
+                  {toString(option.name)}
```
rszwajko, line 342: we should first try the label

client/src/app/components/Autocomplete.tsx
```diff
@@ -307,12 +373,14 @@ export const Autocomplete: React.FC<IAutocompleteProps> = ({
       </FlexItem>
       <FlexItem key="chips">
         <Flex spaceItems={{ default: "spaceItemsXs" }}>
-          {selectedOptions.map(({ id, name, labelName, tooltip }) => (
-            <FlexItem key={id}>
+          {selectedOptions.map(({ id, name, group, labelName, tooltip }) => (
+            <FlexItem key={`${group}:${id}`}>
```
rszwajko, line 377: assuming the `id` is unique we can drop this changes

rszwajko (commented):
IMHO we should make the Autocomplete grouped-by-default - without the need to toggle a special mode using a flag (and branch based on it). If no `group` prop is provided (it's optional) then such option is put into "Ungrouped" category. If all options are there then we have the use case currently supported with `isGrouped === false`.
Main advantages:
1. no branches in the code and simpler code in general 
2. ability to mix grouped and non-grouped options (it's a legal use case)

client/src/app/components/Autocomplete.tsx
```diff
@@ -29,6 +30,11 @@ export interface AutocompleteOptionProps {
 
   /** the tooltip to display on the Label when the option has been selected */
   tooltip?: string | (() => string);
+  /** the group to display the option in */
+  group?: string;
+
+  /** a unique identifier for the option */
+  uniqueId?: string;
```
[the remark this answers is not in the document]
ibolton336: This is sort of tough since most of the database uses id: number. Would need some sort of data transformation & I am just not sure where to do it. Makes me rethink needing a separate component for the AutoCompleteGrouped. WDYT?

client/src/app/components/Autocomplete.tsx
```diff
@@ -29,6 +30,11 @@ export interface AutocompleteOptionProps {
 
   /** the tooltip to display on the Label when the option has been selected */
   tooltip?: string | (() => string);
+  /** the group to display the option in */
+  group?: string;
```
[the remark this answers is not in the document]
ibolton336: Since the group is purely presentation layer, I think we could just stick with the one field for this.

client/src/app/components/Autocomplete.tsx
```diff
@@ -47,13 +54,18 @@ export interface IAutocompleteProps {
   noResultsMessage?: string;
 }
 
+interface GroupedOptions {
```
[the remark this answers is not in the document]
ibolton336: +1 I think I follow you. So if we just pass a falsey value to the label in MenuGroup, then the element is not rendered, right?

client/src/app/components/Autocomplete.tsx
ibolton336: This part along with the ID field refactor piece feels like enough to push me toward just adding another grouped component.

client/src/app/components/Autocomplete.tsx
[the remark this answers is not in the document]
rszwajko: not sure if any falsy will work but `undefined` works this way in the [demo](https://www.patternfly.org/components/menus/menu/#titled-groups-of-items)

client/src/app/components/Autocomplete.tsx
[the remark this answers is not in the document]
rszwajko: OK. if needed we can always extend the interface later on.

client/src/app/components/Autocomplete.tsx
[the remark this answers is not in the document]
rszwajko: > most of the database uses id: number.
>
That's the source of the problem. If unique string based IDs were used then we could expect no conflicts after merging stakeholders and stakeholders group.

In the current situation we need to resolve the ID conflicts so we need a transformation. String-based unique IDs (as you proposed) are a good choice. However if we want to keep it all numbers we can as well design a mapper function that would separate both types i.e.  `({id}) => isGroup ? id : -d`

> Makes me rethink needing a separate component for the AutoCompleteGrouped.
>
IMHO the linear flow inside component is  the most important. If you feel a separate component will look the best then I'm OK with that.

rszwajko (approved):

PR #1800, opened 2024-03-27
:seedling: Refactor table `filterCategories` definitions, rename the prop `key` to `categoryKey`

rszwajko:
Key prop is a special prop used internally by React.


sjd78:
Ah yeah - that's probably why PF5 uses value/children pairs now for dropdown/select/option lists

ibolton336 (approved):
LGTM

sjd78 (approved):
LGTM

PR #1802, opened 2024-03-28
:seedling: Refactor FilterSelectOptionProps

rszwajko:
Key points:
1. do not extend PF SelectOptionProps interface - this allows to add
   custom props to the interface and provides better control what prop
   are forwarded to SelectOption
2. remove key prop as it's a special prop used internally by React
3. use value prop according to PF documentation - effectively replace
   previously used key prop
4. add label prop - to be used as human friendly representation of the
   value
5. add groupLabel prop
6. add chipLabel prop - for cases where it differs from the label prop
   (main use case are tag items)

### DIFF
--- a/client/src/app/components/FilterToolbar/FilterToolbar.tsx
+++ b/client/src/app/components/FilterToolbar/FilterToolbar.tsx
@@ -22,8 +22,12 @@ export enum FilterType {
 
 export type FilterValue = string[] | undefined | null;
 
-export interface FilterSelectOptionProps extends SelectOptionProps {
-  key: string;
+export interface FilterSelectOptionProps {
+  optionProps?: SelectOptionProps;
+  value: string;
+  label?: string;
+  chipLabel?: string;
+  groupLabel?: string;
 }
 
 export interface IBasicFilterCategory<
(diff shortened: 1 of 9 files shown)


sjd78 (changes requested):
Reading through the code and testing the general function of things and everything looks ok.  However, the Tags dropdown is not working correctly anymore.  It doesn't want to close when text is entered in the search box:

https://github.com/konveyor/tackle2-ui/assets/3985964/49087716-58fa-4545-b057-7103dc21c3eb

The type changes and the subsequent refactoring looks good.

The look of the multiselect looks ok, and it seems to behave properly if you don't type in search text.  Only a problem with search text entered.

rszwajko:
@sjd78 
the problem should be fixed in the recent force-push

sjd78:
> @sjd78 the problem should be fixed in the recent force-push

The menu now closes as expected.  :grinning:

sjd78 (approved):
LGTM

Not new with this PR, but perhaps in a followup the behavior where the search text stays present when the dropdown closes is awkward:
image

The only way to remove the text without clearing the selected options is to backspace or highlight/delete it.  The X button within the field clears the text AND the current selections.  IDK if it is a problem, just is ambiguous and awkward.

sjd78:
This PR looks to be a pre-req for #1815 which is a fix for 0.3.1.  Therefore this PR should target 0.3.1 and be cherry-picked.

@ibolton336, does that make sense to you?

PR #1804, opened 2024-03-29
:seedling: Setup CI workflow to use same container as Dockerfile

sjd78:
Lookup the "FROM .* as builder" image from the project's Dockerfile, and use that image to run the CI unit-test job.

This will ensure the unit-test job is using the same container image that the image builds will be using.  A problem with the container should show up quickly in the unit-testing.

### DIFF
--- a/.github/workflows/ci-repo.yml
+++ b/.github/workflows/ci-repo.yml
@@ -10,31 +10,40 @@ on:
       - "main"
       - "release-*"
 
-env:
-  # Note: This should match the node version(s) used in the base Dockerfile
-  node-version: "18"
-
 jobs:
+  unit-test-lookup-image:
+    runs-on: ubuntu-latest
+    outputs:
+      builder-image: ${{ steps.grepBuilder.outputs.builder }}
+    steps:
+      - uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


.github/workflows/ci-repo.yml
```diff
@@ -44,15 +53,3 @@ jobs:
 
       - name: Test
         run: npm run test -- --coverage --watchAll=false
-
-      - name: Upload to codecov (client)
```
rszwajko, line 48: do we drop code coverage?

rszwajko (commented):
Looks good. One question below.

.github/workflows/ci-repo.yml
sjd78: Yes mostly because it hasn't working in a long time, but also because we never really used it.

rszwajko: OK but I would move it to a separate small PR for visibility.

sjd78:
Kicked the codecov resolution to https://github.com/konveyor/tackle2-ui/issues/1820

rszwajko (approved):

PR #1817, opened 2024-04-04
:sparkles: Add empty state for empty groups / stakeholders list 

### LINKED ISSUE #1816: [BUG] Stakeholder with no associated groups shows empty expanded table
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Konveyor version

latest

### Priority

Critical

### Current Behavior

Screenshot 2024-04-04 at 11 44 00 AM

### Expected Behavior

_No response_

### How Reproducible

Always (Default)

### Steps To Reproduce

_No response_

### Environment

```markdown
- OS:
```

### Anything else?

_No response_

ibolton336:
Resolves #1816 
Screenshot 2024-04-04 at 3 25 17 PM
Screenshot 2024-04-04 at 3 25 06 PM
Screenshot 2024-04-04 at 3 24 59 PM


sjd78 (approved):
The Empty state bits LGTM

As @rszwajko pointed out, a bunch of our translations texts are more like string templates than translatable keys.  Since this is fairly prevalent across the codebase, I don't consider this a blocking problem.  Something to think about what/if we want to do something in future.

PR #1818, opened 2024-04-04
:ghost: Add `hack/setup-operator.sh` scripts and more dev doc

sjd78:
Add a helpful script to setup the operator and create a Tackle instance.  Update README.md files to match.  Added some useful commands to `hack/README.md`.

The script is based on the operator's scripts of similar names, but modified to install the dev version without building anything locally first.

For reference, here is the OLM install script: https://github.com/operator-framework/operator-lifecycle-manager/blob/master/scripts/install.sh


rszwajko (approved):
Thanks for sharing!

Issue #1821, opened 2024-04-05
[TECH DEBT] Remove use of `APIClient`


sjd78:
Only a few things in the UI still use the legacy `APIClient` class.  

Those uses should be transitions to using standard axios the same way as everything else.

Issue #1411, opened 2023-09-28
Rest api return type alignment


ibolton336:
Currently, there are discrepancies between return types in our rest.ts file which contains the axios api handlers for external api requests. There is some deprecated usage of an old ApiClient wrapper that is no longer relevant as of the move to react query. This issues is intended to capture:

- [ ] Work through the rest.ts file and check the function return types.  
- [ ] Ensure all queries are aligned with the return types of the api calls

sjd78:
This is ongoing and doesn't need to be tracked directly.

PR #1822, opened 2024-04-05
:seedling: Remove `APIClient`, update effected rest functions

sjd78:
Changes:
  - Remove `APIClient`
  - Update the affected functions in `rest.ts` to use axios directly
  - Double check the return types of the effected functions
  - Updated a few API model types based on current hub handler function responses
  - `getApplicationSummaryCSV()` URL query param was incorrectly named and could lead to more results returned than expected.  Query param updated to the correct name as per the OpenAPI definition / golang handler code.

Resolves: #1821
Part-of: #1411


client/src/app/api/models.ts
```diff
@@ -190,18 +190,17 @@ export type IdentityKind =
 
 export interface Identity {
```
rszwajko, line 191: we need to sync the API manually - it would be nice to have a link (as a comment)  to the original backend structure i.e. https://github.com/konveyor/tackle2-hub/blob/94330e1ff9cdb479db3b6aeb30f9abb6859c16bc/api/identity.go#L229

sjd78: Maybe a comment with a link to where the openapi.json is kept?

client/src/app/api/rest.ts
```diff
@@ -179,21 +178,17 @@ export const deleteReview = (id: number): Promise<Review> => {
...
-      applicationId: f,
-    }))
+    applicationIds.map((f) => ({ applicationId: f }))
   );
 };
 
-export const getApplicationSummaryCSV = (id: string): AxiosPromise => {
-  return APIClient.get(`${APP_IMPORT_CSV}?importSummaryId=${id}`, {
+export const getApplicationSummaryCSV = (id: string) => {
+  return axios.get<ArrayBuffer>(`${APP_IMPORT_CSV}?importSummary.id=${id}`, {
```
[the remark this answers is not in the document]
sjd78: `importSummaryId` isn't valid anywhere in tackle2-hub!

https://github.com/konveyor/tackle2-hub/blob/main/api/import.go#L92

client/src/app/api/models.ts
rszwajko: > There would be lots of comments if we started doing that
>
right. let's discuss this and provide some global solution as a separate issue/PR

client/src/app/api/rest.ts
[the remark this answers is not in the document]
rszwajko: it looks like a typo :) perhaps we should add a comment to  explain this?

ibolton336 (approved):

rszwajko (approved):
Looks good. The case of param name `importSummary.id` (and why it worked before!) is worth documenting but I leave it to you.

sjd78:
> Looks good. The case of param name `importSummary.id` (and why it worked before!) is worth documenting but I leave it to you.

It ~~worked~~ didn't fail before because the query param is not required.

PR #1843, opened 2024-04-12
:bug: Address required status bug

ibolton336:
Resolves https://issues.redhat.com/browse/MTA-2505
- Adds additional unit test coverage

### DIFF
--- a/client/src/app/pages/applications/components/application-assessment-status/tests/application-assessment-status.test.tsx
+++ b/client/src/app/pages/applications/components/application-assessment-status/tests/application-assessment-status.test.tsx
@@ -20,6 +20,110 @@ describe("useAssessmentStatus", () => {
     server.resetHandlers();
   });
 
+  it("Updates hasApplicationAssessmentInProgress to false when assessment is marked as not required", async () => {
+    server.use(
+      rest.get("/hub/archetypes", (req, res, ctx) => {
+        return res(
+          ctx.json([
+            createMockArchetype({
+              id: 1,
+              name: "archetype1",
+              applications: [],
+              assessed: false,
+              assessments: [],
+            }),
+          ])
+        );
+      }),
+      rest.get("/hub/assessments", (req, res, ctx) => {
+        return res(
+          ctx.json([
+            createMockAssessment({
+              id: 1,
+              application: { id: 1, name: "app1" },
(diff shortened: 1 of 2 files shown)


rszwajko:
I still see 'in progress' status in the described scenario: 
1. fully assess an application ("save and review") - assessment status changes to "completed"
2. disable the questionnaire (mark as not required)
3. go back to  application table - assessment status is "in progress"

rszwajko (approved):

client/src/app/pages/applications/components/application-assessment-status/tests/application-assessment-status.test.tsx
```diff
@@ -20,6 +20,110 @@ describe("useAssessmentStatus", () => {
...
+              id: 2,
+              application: { id: 1, name: "app1" },
+              questionnaire: { id: 2, name: "questionnaire2" },
+              status: "complete",
+              required: false,
+              sections: [],
+            }),
+          ])
+        );
+      })
```
sjd78, line 115: Having the second set of handlers is a bit confusing.

I suggest just doing a use once kind of setup so it is more obvious that the data will change on the second request.  

This may work (not tested):
```jsx
    rest.get("/hub/assessments", (req, res, ctx) => {
        return res.once(
          // first json
        );
      })
    rest.get("/hub/assessments", (req, res, ctx) => {
        return res(
          // json for the rest of the time
        );
      })
```

See: https://v1.mswjs.io/docs/api/setup-server/use#one-time-override

sjd78 (commented):
Change looks good, and I like the unit test!

I think the unit test could be rearranged to keep a single `server.use()` in place.

client/src/app/pages/applications/components/application-assessment-status/tests/application-assessment-status.test.tsx
sjd78: Or I just found you can use a generator function in this case -- way cooler:
https://github.com/mswjs/msw/discussions/1117#discussioncomment-3124548

It points at this bit of code for more examples: https://github.com/mswjs/msw/blob/6990c3a57a1f515fe977c2eaa7741005b885b8af/test/rest-api/generator.mocks.ts#L4

//cc @rszwajko

PR #1855, opened 2024-04-16
:bug: Fallback to tag category name for empty categories

rszwajko:
Resolves: https://issues.redhat.com/browse/MTA-2186


rszwajko:
### After the fix

image

Issue #1873, opened 2024-04-19
[Enhancement] Improve tag spacing


rszwajko:
Implements: https://github.com/konveyor/enhancements/issues/148

PR #1878, opened 2024-04-22
:sparkles: Extract IconWithLabel component

rszwajko:
Resolves: #1873


client/src/app/components/Icons/IconedStatus.tsx
```diff
@@ -164,28 +157,25 @@ export const IconedStatus: React.FC<IIconedStatusProps> = ({
...
-      {(preset === "InheritedReviews" ||
+    <IconWithLabel
+      iconTooltipMessage={presetProps?.tooltipMessage}
+      icon={
+        <Icon status={status || presetProps?.status} className={className}>
+          {icon || presetProps?.icon || <UnknownIcon />}
+        </Icon>
+      }
+      label={label || presetProps?.label}
+      hasTrailingItem={
```
sjd78, line 168: Why have a flag saying to include the value of another prop instead of just including the value of the prop if it is not undefined (which would be the default)?

client/src/app/components/Icons/IconWithLabel.tsx
```diff
@@ -0,0 +1,24 @@
+import React, { FC, ReactElement, ReactNode } from "react";
+import { Flex, FlexItem } from "@patternfly/react-core";
+import { IconWithOptionalTooltip } from "./IconWithOptionalTooltip";
+
+export const IconWithLabel: FC<{
+  iconTooltipMessage?: string;
+  icon: ReactElement;
+  label: ReactNode;
+  hasTrailingItem?: boolean;
```
sjd78, line 9: Seems like this flag isn't really needed.

sjd78 (changes requested):
Everything looks good other than the `hasTrailingItem` prop.

client/src/app/components/Icons/IconedStatus.tsx
rszwajko: @sjd78 
In general, I agree. The split was motivated by the way we currently use the component: the condition is quite long. Dedicated prop seemed more readable i.e.
```
hasTrailingItem={
    preset === "InheritedReviews" ||
    preset === "InheritedAssessments" ||
    preset === "InProgressInheritedAssessments" ||
    preset === "InProgressInheritedReviews"
  }
  trailingItem={
    <Tooltip content={getTooltipContent()}>
      <TopologyIcon />
    </Tooltip>
  }
```
If we remove the dedicated flag then a ternary operator will be required i.e.
```      
trailingItem={
    preset === "InheritedReviews" ||
    preset === "InheritedAssessments" ||
    preset === "InProgressInheritedAssessments" ||
    preset === "InProgressInheritedReviews" ? (
      <Tooltip content={getTooltipContent()}>
        <TopologyIcon />
      </Tooltip>
    ) : undefined
  }
```
If you think it's not too long then we can use single prop.

sjd78: I prefer that, or to move the trailingItem value outside of the full JSX construct

sjd78: Looking at it again quick -- you could embed a `trailingTopologyIcon` boolean to the presets and simplify the expression as well.

rszwajko:
@sjd78 
` trailingItemTooltipMessage` is a separate prop to increase flexibility - currently we use the same tooltip for both icons.

sjd78 (approved):
LGTM!

The `topologyIcon` preset prop looks nice.

PR #1885, opened 2024-04-25
:ghost: Update check for running analysis / cancel analysis 

ibolton336:
This PR narrows down the check for analysis tasks that are cancellable. Also the logic is updated and flipped for "disabling" the creation of a new analysis task. Analysis tasks should be creatable when there are no states for tasks outside of "succeeded" or "failed". This will future proof against any other additional task states that may be added in the future. The "scheduled" state was just added for 0.4 & is the motivation behind this PR ( as requested by @jortel )

### DIFF
--- a/client/src/app/pages/applications/applications-table/applications-table.tsx
+++ b/client/src/app/pages/applications/applications-table/applications-table.tsx
@@ -154,9 +154,7 @@ export const ApplicationsTable: React.FC = () => {
 
   const isTaskCancellable = (application: Application) => {
     const task = getTask(application);
-    if (task?.state && task.state.match(/(Created|Running|Ready|Pending)/))
-      return true;
-    return false;
+    return task?.state && !["Succeeded", "Failed"].includes(task.state);
   };
 
   const cancelAnalysis = (row: Application) => {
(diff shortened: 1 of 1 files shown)


client/src/app/pages/applications/applications-table/applications-table.tsx
```diff
@@ -155,8 +155,7 @@ export const ApplicationsTable: React.FC = () => {
 
   const isTaskCancellable = (application: Application) => {
     const task = getTask(application);
-    if (task?.state && task.state.match(/(Created|Running|Ready|Pending)/))
-      return true;
+    if (task?.state && !task.state.match(/(Succeeded|Failed)/)) return true;
```
sjd78, line 158: I'm more partial to an array include check in this circumstance, and I even like regex!

For example:
```js
  const isTaskCancellable = (application: Application) => {
    const task = getTask(application);
    return task?.state && !["Succeeded", "Failed"].includes(task.state);
  }
```

client/src/app/pages/applications/applications-table/applications-table.tsx
```diff
@@ -610,13 +609,13 @@ export const ApplicationsTable: React.FC = () => {
...
 
-  const isAnalyzingAllowed = () => {
+  const isAnalyzingDisabled = () => {
     const candidateTasks = selectedRows.filter(
       (app) =>
         !tasks.some(
           (task) =>
             task.application?.id === app.id &&
-            task.state?.match(/(Created|Running|Ready|Pending)/)
+            task.state?.match(/(Succeeded|Failed)/)
```
sjd78, line 618: Same here, could switch the array includes check instead of the regex.

ibolton336: Agreed on switching from regex being the better fit.

sjd78 (approved):
The checks read a bunch better now with the `.includes()` in place of the regex.

sjd78:
> * Update this logic to account for new states added to task engine on backend API side

Is this PR dependent on a hub PR getting merged?  If yes, please reference the PR or hub issue in the PR description.

ibolton336:
> Is this PR dependent on a hub PR getting merged? If yes, please reference the PR or hub issue in the PR description.

Description updated with more context! Could not find a specific hub PR for this. @sjd78

rszwajko:
@ibolton336 
integration tests fail because they cannot delete an application 
```
  AssertionError: Timed out retrying after 8000ms: expected '<button#analyze-application.pf-v5-c-button.pf-m-primary.pf-m-disabled>' to be 'enabled'
      at Analysis.startAnalysis (cypress/e2e/models/migration/applicationinventory/analysis.ts:199:43)
      (...)

No elements in the current DOM matched your query:

  > cy.get(span.pf-v5-c-menu__item-text).contains(Delete)
(...)
  From Your Spec Code:
      at clickItemInKebabMenu (cypress/utils/utils.ts:657:57)
      at Application.delete (cypress/e2e/models/migration/applicationinventory/application.ts:254:29)
     
```

Issue #954, opened 2023-05-31
Sidebar falls out of sync when hard navigating to pages outside of the current personas scope


ibolton336:

rszwajko:
Steps to recreate:
1.  select "Administration" view
2.  navigate directly to a page associated with "Migration" view i.e. `/controls/job-functions`
3. expected results: the sidebar shows "Migration" view as selected view (with corresponding pages)
   actual results: the sidebar for "Administration" view is show (see below)
![image](https://github.com/konveyor/tackle2-ui/assets/64194103/b00439db-c614-4b71-b108-ec3d573731bd)

PR #1889, opened 2024-04-26
:bug: Switch sidebar type based on routes

rszwajko:
Linked changes:
1. fix incorrect handling of Paths.assessment and Paths.questionnaire - both were incorrectly handled as dev routes
2. split Paths enum into type safe specialized readonly objects - DevPaths, AdminPaths and GeneralPaths

Resolves: #954


ibolton336 (approved):
LGTM

client/src/app/layout/SidebarApp/SidebarApp.tsx
```diff
@@ -8,201 +8,208 @@ import {
...
+);
 
-  const [selectedPersona, setSelectedPersona] = useLocalStorage<string | null>({
-    key: LocalStorageKey.selectedPersona,
-    defaultValue: null,
-  });
+const PersonaSidebar: FC<{
+  children: ReactElement;
+  selectedPersona: PersonaType;
+  targetPath: Paths;
```
sjd78, line 55: What the `targetPath` should be is a bit confusing.  I understand it as meaning, "when the persona is changed, change the url to this path".

What if `PersonaKey` and `PersonaType` looked more like this:
```ts
const PersonaKey = {
  ADMINISTRATION: {
    label: "Administration",
    targetPath: Paths.general,
  },
  MIGRATION: {
    label: "Migration",
    targetPath: Paths.applications,
  },
};

type PersonaType = keyof typeof PersonaKey;
```

Then the `targetPath` for a persona would live with the persona and would not need to be a prop at all.

Also, not that I know of any plans to add a new persona or another sidebar, a structure like this would allow adding another persona pretty easily.

sjd78 (changes requested):
I like it so far, but I have a suggestion around the `targetPath` prop.

sjd78:
I did find a strange behavior:


https://github.com/konveyor/tackle2-ui/assets/3985964/01ecad4f-3309-4501-834f-588c092c0006

ibolton336 (approved):

sjd78 (approved):
LGTM

Working from the `AdminPaths` and `GeneralPaths` constants through types and back out to the sidebar makes sense to me.  Typescript should raise enough alarms if something is done that doesn't line up properly that the change is worth it.

Issue #1774, opened 2024-03-13
[BUG] Selected filters seem to dissappear when not present in opened multi select filter


rszwajko:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Konveyor version

0.3

### Priority

Minor

### Current Behavior

Already selected filter (visible as a chip) disappears when its value is not present in currently opened multi select filter.
[Screencast from 2024-03-13 14-42-50.webm](https://github.com/konveyor/tackle2-ui/assets/64194103/98ee404d-5911-4bfb-82a2-edbdc81ae161)

### Expected Behavior

Selected value does not change.

### How Reproducible

Always (Default)

### Environment

```markdown
- OS:
```

PR #1903, opened 2024-05-08
:bug: Use separate chip groups for tag categories

rszwajko:
Use one ToolbarFilter per options group. First filter provides 
the option list. Remaining filters are hidden and used only for
side effects (separate chip groups). The approach follows
similar widgets used by Forklift plugin and OpenShift Console but
uses a flat list of toolbars (nesting trick requires the widget to be
visible all the time).

Related refactorings in MuliselectFilterControl:
1. drop unused feature - using dictionary type for selectedOptions
2. drop state that can be calculated:
   a) active item from focusedItemIndex
   b) selectedOptions from filters
3. centralize id calculations - use prefix based on category title
4. use label styling for tag category part of option

Resolves: #1774 

Reference-Url: https://github.com/kubev2v/forklift-console-plugin/pull/90
Reference-Url: https://github.com/openshift/console/blob/5ba18580676a25e4304df78253aad6a9832d4d56/frontend/public/components/filter-toolbar.tsx#L299

### DIFF
--- a/client/src/app/components/FilterToolbar/MultiselectFilterControl.tsx
+++ b/client/src/app/components/FilterToolbar/MultiselectFilterControl.tsx
@@ -5,7 +5,6 @@ import {
   MenuToggle,
   MenuToggleElement,
   Select,
-  SelectGroup,
   SelectList,
   SelectOption,
   TextInputGroup,
@@ -17,8 +16,8 @@ import {
 } from "@patternfly/react-core";
 import { IFilterControlProps } from "./FilterControl";
 import {
-  IMultiselectFilterCategory,
   FilterSelectOptionProps,
+  IMultiselectFilterCategory,
 } from "./FilterToolbar";
 import { css } from "@patternfly/react-styles";
 import { TimesIcon } from "@patternfly/react-icons";
@@ -31,6 +30,8 @@ export interface IMultiselectFilterControlProps<TItem>
   isScrollable?: boolean;
 }
 
+const NO_RESULTS = "no-results";
+
 export const MultiselectFilterControl = <TItem,>({
   category,
   filterValue,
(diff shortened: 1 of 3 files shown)


rszwajko:
image

ibolton336:
@JustinXHale

client/src/app/components/FilterToolbar/MultiselectFilterControl.tsx
```diff
@@ -42,37 +42,43 @@ export const MultiselectFilterControl = <TItem,>({
...
+  const [firstGroup, ...otherGroups] = [
+    ...new Set([
+      ...(category.selectOptions
+        ?.map(({ groupLabel }) => groupLabel)
+        .filter(Boolean) ?? []),
+      defaultGroup,
+    ]),
+  ];
+
+  const onFilterClearAll = (groupName: string) =>
```
sjd78, line 67: `s/onFilterClearAll/onFilterClearGroup/`

sjd78 (commented):
A few questions:
  - Should the text filter apply to the tag category? What should be shown in the list if you type part of the category name "3rd"?  (This kinda assumes the "3rd party" tag category is available on a fresh install)
 
  - When navigating the drop down with keyboard input, should a page up/down bring checkbox focus with it the same way as up/down does?  Right now if I hit page down, the checkbox focus does not follow and I need to use the arrow keys to scroll back to what I was looking at.

Visually looks great!

The chip groups look good and work great.

I'll take one more pass at the code later today.

client/src/app/components/FilterToolbar/MultiselectFilterControl.tsx
```diff
@@ -337,27 +225,116 @@ export const MultiselectFilterControl = <TItem,>({
...
+                      value={value}
+                      isFocused={focusedItemIndex === index}
+                      isSelected={filterValue?.includes(value)}
+                    >
+                      {!!groupLabel && <Label>{groupLabel}</Label>}{" "}
+                      {label ?? value}
+                    </SelectOption>
+                  )
+                ),
+                !filteredOptions.length && (
```
sjd78, line 266: nit-pick: I like this check looking like:
```tsx
                filteredOptions.length === 0 && (
```

But then again, the array spread you're using doesn't need to be that way either.   Could just be:
```tsx
  {filteredOptions.map(({ ... }) => {
    <SelectOption ... />
  }}
  {filteredOptions.length === 0 ? (
    <SelectOption NORESULTS />
  ) ? undefined}
```

client/src/app/components/FilterToolbar/MultiselectFilterControl.tsx
```diff
@@ -42,37 +42,43 @@ export const MultiselectFilterControl = <TItem,>({
...
+  const idPrefix = `filter-control-${category.categoryKey}`;
+  const withPrefix = (id: string) => `${idPrefix}-${id}`;
+  const defaultGroup = category.title;
 
-  const [focusedItemIndex, setFocusedItemIndex] = React.useState<number | null>(
-    null
+  const filteredOptions = category.selectOptions?.filter(({ label, value }) =>
+    String(label ?? value ?? "")
+      .toLowerCase()
+      .includes(inputValue?.trim().toLowerCase() ?? "")
```
sjd78, line 55: This will only filter by the option's "label" and does not include the option's "groupLabel".

So trying to filter the select options by a tag category doesn't work.

client/src/app/pages/applications/applications-table/applications-table.tsx
```diff
@@ -468,7 +468,7 @@ export const ApplicationsTable: React.FC = () => {
           }) + "...",
         selectOptions: tagItems.map(({ name, tagName, categoryName }) => ({
           value: name,
-          label: name,
+          label: tagName,
```
sjd78, line 471: Humm, looks like this change is what is making the select box filter only on a tag name and not include the tag category.

Ref 1: 
https://github.com/konveyor/tackle2-ui/blob/8c8a0f852d861febebc48674b8aceac6b1c3911f/client/src/app/queries/tags.ts#L82-L92

Ref 2:
`filteredOptions()` up in the `MultiselectFilterControl`

client/src/app/components/FilterToolbar/MultiselectFilterControl.tsx
```diff
@@ -337,27 +225,116 @@ export const MultiselectFilterControl = <TItem,>({
...
-        onSelect={(_, selection) => onSelect(selection as string)}
-        isOpen={isFilterDropdownOpen}
-      >
-        <SelectList id="select-multi-typeahead-checkbox-listbox">
-          {renderSelectOptions(() => true)}
-        </SelectList>
-      </Select>
-    </ToolbarFilter>
+    <>
+      {[
```
sjd78, line 229: Does wrapping the children in an array do anything useful here or could the just be removed and retain the fragment to capture all the children components?
```tsx
<>
  <ToolbarFilter ... />
  {otherGroups.map((groupName) => (
    <ToolbarFilter ... />
  ))}
</>
```

sjd78 (changes requested):
A few style comments, but the main one is with the `filteredOptions` handling.  That needs to be group aware so item filtering works as expected (i.e. filter is applied to both the tag category and tag name)

client/src/app/components/FilterToolbar/MultiselectFilterControl.tsx
rszwajko: good catch! previously  `label` == `value` in this scenario (and `value` contains both tag category and tag name)

sjd78 (approved):
Build of `filteredOptions` looks great!

LGTM

Issue #1883, opened 2024-04-24
Enable dependabot tracking and add CI tests for the Dockerfile


sjd78:
Since using the `:latest` label on the build and run time base images in the Dockerfile is dangerous (see #1746 and #1781), being able to use a specific tested version of the images is important.  Dependabot support tracking the images in use in a Dockerfile and creating a PR when never versions are available.  Also, since sometimes a problem with the build doesn't surface until the on-merge actions are run, CI specifically for the Dockerfile should be implemented.

Things to do:
  - Enable dependabot on the Dockerfile.  Multiple ecosystems are allowed in the config, so this style is allowable:
```yaml
version: 2
updates:
  - package-ecosystem: "npm"
    directory: "/"
    schedule:
      interval: "daily"
    allow:
      - dependency-name: "@patternfly/*"
        dependency-type: "direct"

  - package-ecosystem: docker
    directory: "/"
    schedule:
      interval: daily
```

  - Based on the on-merge workflow [image-build.yaml](https://github.com/konveyor/tackle2-ui/blob/main/.github/workflows/image-build.yaml) that builds the images and uploads to a repo, create an action:
    - If the Dockerfile has been touched, execute.  Else, skip.
    - Run the image build with the same settings as the multiarch build, but don't push to a repo
    - Pass CI on a Dockerfile change when the image builds all successfully complete (in the github action environment)

sjd78:
Part 1 implemented in #1908
Part 2 implemented in #1907

PR #1907, opened 2024-05-10
:sparkles: Test image build when build related files change

sjd78:
If the workflow is run from a PR, and the PR includes a change to the `Dockerfile` or `package-lock.json`, then run image builds for all of our target platforms.

The images are built but not pushed to any repository.

We want to be reasonably sure that any major build file changes will not cause the image-build-and-push on PR merge workflow to break.  Doing the image build here should reveal most problems much earlier.  For example, a npm version update in the build container could break github action `nofiles` or network access capabilities for the npm install.

See #1742, #1746, and #1781 for some other examples of when this check could have caught issues before a PR merge.

Supports: #1883

### DIFF
--- a/.github/workflows/ci-image-build.yml
+++ b/.github/workflows/ci-image-build.yml
@@ -1,0 +1,105 @@
+name: CI (test image build for a PR with build related changes)
+
+on:
+  pull_request:
+    branches:
+      - "main"
+      - "release-*"
+
+jobs:
+  checks:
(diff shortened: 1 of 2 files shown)


sjd78:
**Note**: This PR should be backported to **release-0.3** and **release-0.4** to help protect the active release branches.

djzager (commented):
VISACK. Good idea.

ibolton336 (approved):

PR #1908, opened 2024-05-10
:seedling: Have dependabot track `Dockerfile` referenced images

sjd78:
Configure dependabot to create PRs when the images we use in the `Dockerfile` have a newer version.  This allows us to reference specific container images and be automatically notified of newer versions.

Supports: #1883

CI added in #1907 will run against the PRs dependabot will open for `Dockerfile` changes.  This will allow us to have confidence that the update will be successful.


rszwajko (approved):
Great!

ibolton336 (approved):

PR #1913, opened 2024-05-14
:sparkles: Add DateRangeFilter

rszwajko:
Create a filter widget based on DateRangeFilter developed               
for Forklift plugin.                                                    
Use the new widget to filter migration waves on both start and end date.
                                                                        
Key points:                                                             
1. filter values are stored as strings in ISO 8601 time interval        
   format (date part only) i.e. "2024-04-01/2024-05-01".                
2. date range is a closed range (both ends are included)                
3. browser local time zone is used                                      
4. date format is hard-coded to "MM/DD/YYYY" to match the format        
   used in the UI                                                       
                                                                        
Additional changes:                                                     
1. force TZ=UTC for jest unit tests to ensure the same test results     
2. initialize dayjs plugins for tests                                   
                                                     
Reference-Url: https://github.com/kubev2v/forklift-console-plugin/pull/754
Reference-Url: https://en.wikipedia.org/wiki/ISO_8601#Time_intervals


rszwajko:
### Date range filter in action 
image

ibolton336 (approved):

sjd78 (approved):

PR #1920, opened 2024-05-22
:seedling:  Unit test getLocalFilterDerivedState

rszwajko:


sjd78 (approved):
LGTM

(with the extreme minor nit-pick that I like a line of white-space between `it()` calls so my eyes can more easily pattern match the block -- but that is a Scott preference thing)

ibolton336:
LGTM

Issue #1318, opened 2023-08-29
Update all new tables to use the ActionsColumn component from PF5


ibolton336:
Example here: https://github.com/konveyor/tackle2-ui/pull/1314/files

mturley:
As part of doing this, we may want to also adopt the design pattern for Delete actions of putting them below a separator (item object with `{ isSeparator: true }`) and use the `isDanger` property (makes the text red), as I'm doing in https://github.com/konveyor/tackle2-ui/pull/1328. See example here: https://www.patternfly.org/components/menus/menu#danger-menu-item

sjd78:
This work is related to https://issues.redhat.com/browse/MTA-2261

rszwajko:
Last remaining tables migrated in #3275

PR #1921, opened 2024-05-23
:sparkles: Update business services table to use ActionsColumn

mguetta1:
Related to #1318

image

OR

image


sjd78 (commented):
LGTM, I just want to check on the edit pencil icon in some other tables

sjd78 (approved):
LGTM

sjd78 (approved):

sjd78:
@mguetta1 - I like the one with the edit pencil.  Which option do you want to use?

rszwajko (changes requested):
@mguetta1  @ibolton336 
`AppTableActionButtons` is a wrapper that provides RBAC access check. Please provide some more context why we need to remove it.

ibolton336:
> @mguetta1 @ibolton336 `AppTableActionButtons` is a wrapper that provides RBAC access check. Please provide some more context why we need to remove it.

The idea initially was to standardize on a common approach in the table now that this new table component was available. Modifying the AppTableActionButtons component to make use of the ActionsColumn may be a better approach. The use of PF Flex inside the AppTableActionButtons causes some inconsistent spacing when compared to the tables using ActionsColumn.

sjd78:
> > @mguetta1 @ibolton336 `AppTableActionButtons` is a wrapper that provides RBAC access check. Please provide some more context why we need to remove it.
> 
> The idea initially was to standardize on a common approach in the table now that this new table component was available. Modifying the AppTableActionButtons component to make use of the ActionsColumn may be a better approach. The use of PF Flex inside the AppTableActionButtons causes some inconsistent spacing when compared to the tables using ActionsColumn.

Structurally I like the "new" style for the component layout.  However, with the different containing component, are we dropping some RBAC checks?  That could be ok so long as there is a new effort soon (i.e. in the release after 0.5) to refactor RBAC and apply it on all actions.  Humm, I may need to write up an issue to capture that work (and keycloak lib upgrades).

mguetta1:
> @mguetta1 - I like the one with the edit pencil. Which option do you want to use?

This option

mguetta1:
@rszwajko @ibolton336 @sjd78 
I don't think that there is any permission restrictions on controls. Anyway I added the `controlsWriteScopes` to the business service create button

ibolton336:
> @rszwajko @ibolton336 @sjd78 I don't think that there is any permission restrictions on controls. Anyway I added the `controlsWriteScopes` to the business service create button [69ff6b2](https://github.com/konveyor/tackle2-ui/pull/1921/commits/69ff6b235eade30d3afd62f41c912c93724d049c)

I think it'd be worth trying to save the component for reuse as Radek mentioned. That way we don't need a new PR for each place we make these changes & can avoid maintenance headaches.

mguetta1:
https://github.com/konveyor/tackle2-ui/assets/60384172/14457d94-b95c-4f36-a75e-a8d483077e15

client/src/app/components/ControlTableActionButtons.tsx
sjd78, line 1: This is probably better located in the folder `client/src/app/pages/controls` since the RBAC is setup for Controls.

sjd78 (changes requested):
This looks good.  I'd just like to see the new component in a different folder.

client/src/app/components/ControlTableActionButtons.tsx
mguetta1: Done in https://github.com/konveyor/tackle2-ui/pull/1921/commits/730f1e020f23c25897643fabec5c6b62d5cfc942

sjd78 (approved):
LGTM

PR #1922, opened 2024-05-23
:sparkles: Add custom matchers to client side filters

rszwajko:
Before this PR, all values were matched in the same way:
1. value needs to be string or have string representation via
   getItemValue
2. positive match is returned if (lowercased) value contains
   (lowercased) filter value.

After this PR, a custom algorithm can be provided using "matcher"
property. Direct motivation is DateRange filter which stores filter
values as ISO 8601 time intervals i.e. "2024-04-01/2024-05-01".
Positive match (value is in the range) requires parsing to date objects.

Reference-Url: https://en.wikipedia.org/wiki/ISO_8601#Time_intervals
Reference-Url: https://github.com/konveyor/tackle2-ui/pull/1913


sjd78 (approved):
LGTM

PR #1846, opened 2024-04-14
:bug: "Effort" with supplemental text info via tooltip 

kpunwatk:
Resolves https://issues.redhat.com/browse/MTA-2556

PR #1933, opened 2024-06-03
:bug: On Issues>Affected Application table, update column data-label to match displayed text

ibolton336:
Resolves: #1936 
UI Tests PR: 1124
Addresses mismatch between data column and displayed text for "Effort" Column in affected applications table

PR konveyor/tackle2-ui#1846 intended to revert back to using effort favoring tooltips to differentiate what is being displayed in fields labeled effort.  The Issues>Affected Application table effort column heading text was updated, but the data-label used was still "Total Effort".


sjd78 (approved):
LGTM and with this change, there are no more occurrences of "Total Effort" in the UI

PR #1930, opened 2024-05-31
:sparkles: Move Analysis Details to full page view

rszwajko:
Related changes:
1. extend BreadCrumbPath component to support non-links
2. refactor CSS styles for SimpleDocumentViewer to support both modal
   and in-page mode
3. create new route for the analysis details page:
   /applications/:applicationId/analysis-details/:taskId

Part-of: https://github.com/konveyor/tackle2-ui/issues/1929

PR #1939, opened 2024-06-05
 :sparkles: Link to drawer

rszwajko:


sjd78 (approved):
Ignoring the changes from #1930 in the first commit, looking at just commit 764de35c244192aa399296c2061bcdb25c1bc065, looks good.

The only minor thing is if once on the log viewer page, if the application name should be a link so a click on it will work just like the browser back.  That can be adjusted in a follow up.

PR #1940, opened 2024-06-05
:bug: Remove unnecessary initialData from useQuery calls

rszwajko:
Initial data is persisted in cache which results in:
1. isLoading flag always false
2. with custom staleTime, delaying first fetch until cache expires


rszwajko:
@sjd78  @ibolton336 
The bug was found when testing #1939 - `isFetchingApplications` was always false which breaks the the linking functionality.

Btw note that we use a different meaning of `isFetching` then React Query - under the hood it's translated to `isLoading` (which is a different state).

ibolton336 (approved):

sjd78 (approved):

PR #1946, opened 2024-06-06
:bug: Fix deps filter crash bug by limiting tag grouping calculation to tag filters

ibolton336:
Addresses issue https://github.com/konveyor/tackle2-ui/issues/1944 introduced by https://github.com/konveyor/tackle2-ui/pull/1903 

Currently the operation to build the tag category labels is very expensive & is causing maximum update depth exceeded errors when performing a filter operation in some cases. 

Screenshot 2024-06-06 at 2 15 12 PM

This PR introduces a quick fix to limit the label creation operation by categoryKey === 'tags'. We should circle back on this & see if there is a better way to implement the category / name grouping.

### DIFF
--- a/client/src/app/components/FilterToolbar/MultiselectFilterControl.tsx
+++ b/client/src/app/components/FilterToolbar/MultiselectFilterControl.tsx
@@ -57,14 +57,21 @@ export const MultiselectFilterControl = <TItem,>({
         .some((it) => it.includes(inputValue?.trim().toLowerCase() ?? ""))
   );
 
-  const [firstGroup, ...otherGroups] = [
-    ...new Set([
-      ...(category.selectOptions
-        ?.map(({ groupLabel }) => groupLabel)
-        .filter(Boolean) ?? []),
-      defaultGroup,
-    ]),
-  ];
+  const groupLabels = React.useMemo(() => {
+    if (category.categoryKey === "tags") {
+      return [
+        ...new Set([
+          ...(category.selectOptions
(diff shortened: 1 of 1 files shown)


rszwajko:
@ibolton336 @sjd78 
The root cause seem to be chip group name collision. Both "Language" filter and "Tags" filter create chip group with title "Language". After renaming one of them the problem is solved (see below).
However tag categories, in general, are editable. We might run into more conflicts. The solution might be using a prefix for chip groups created from tags categories i.e. ` Tags/Language`. What do you think?
```
diff --git a/client/src/app/pages/dependencies/dependencies.tsx b/client/src/app/pages/dependencies/dependencies.tsx
index 97db65ce..166ff7aa 100644
--- a/client/src/app/pages/dependencies/dependencies.tsx
+++ b/client/src/app/pages/dependencies/dependencies.tsx
@@ -75,7 +75,7 @@ export const Dependencies: React.FC = () => {
       },
       {
         categoryKey: "provider",
-        title: t("terms.language"),
+        title: "LL",
         type: FilterType.search,
         filterGroup: "Dependency",
         placeholderText:
```
Screenshot from 2024-06-07 17-10-50

rszwajko:
@ibolton336 @sjd78 
Further research revealed that the problem is in duplicated keys - ToolbarFilter derrives them from `categoryName` which in our filters usually comes from `category.title`.
I've created a draft PR with the fix ( #1949 ).

Issue #1950, opened 2024-06-10
Modify analysis wizard to accommodate language discovery


sjd78:
## Summary

With the addition of [automatic language discovery](https://github.com/konveyor/enhancements/issues/170), the analysis wizard target selection step needs to be enhanced.

Enhancement references:
  - https://github.com/konveyor/enhancements/issues/170
  - https://github.com/konveyor/enhancements/issues/155
  - https://github.com/konveyor/enhancements/pull/173 - [enhancement document](https://github.com/konveyor/enhancements/tree/master/enhancements/multiple-provider)

## Requirements

Base requirements:
  - Existing Wizard functionality remains the same
  - On the "Set targets" step:
    - The language selection box needs to be updated to dynamically source the option list, and the selected options
    - The language selection box needs to support an "All Language" and multiple selection of languages
    - The set of target cards displayed will continue to match the selections in the language box
    - Each target card will need a visual indicator of its language/provider
  - When creating the task resources, `TaskGroup.Addon: analyzer` needs to be replaced with `TaskGroup.Kind: analyzer`

Details on changes to the language selection box:
  - Provide a "(Show All)" option to display all targets regardless of language tags
  - Populate the options list from the `Provider` field of the fetched targets.
  - Initially select the languages based on the tags, of tag category "Language", across the applications selected for analysis.

Details on changes to the target cards:
  - Each target card should always display a label with it's provider.  This will allow cards that belong to different providers to be differentiated when more than one language is selected.

## HUB enhancements that may be needed:

None identified initially.  Target data is already available.  Tag category of "Language" is provided in the seed values.  Applications already have their tag data provided.

### Current State
![image](https://github.com/konveyor/tackle2-ui/assets/3985964/3c9aed7f-41af-47ce-9f3f-44c36ffa9f8c)

### Future State
_(the only real difference will be he constant language/provider tag on the cards)_

sjd78:
Note: Until language discovery is working consistently, we can manually tag a language to an application for testing purposes.

jortel:
The application `Analysis` fields used to correlated to task by addon.  This needs to be updated to correlate by kind == analyzer (when not blank) else fallback on addon == analyzer.  The fallback is needed for tasks created in previous releases.

Issue #1252, opened 2023-08-07
Update TargetCard to use CardHeader guidelines


gitdallas:
Refactor the TargetCard component a bit, using PF `CardHeader` for header content and checkbox. Show checkbox for both checked and unchecked.

Consider using selectableActions to handle changes - this may or may not make sense in our implementation.

sjd78:
@ibolton336 -- include in work for #1950 ?

ibolton336:
This was fixed in https://github.com/konveyor/tackle2-ui/pull/1251/files

ibolton336:
Changing the line in target-card.tsx 

`    <Card
      onClick={handleCardClick}
       isSelectable={!!cardSelected}`
to 

`    <Card
      onClick={handleCardClick}
       isSelectable={true}`

introduces a nasty PF bug where selecting a card results in triggering the onMove event via a stepChange event being fired. 

Update: event.preventDefault fixed this in the card onClick handler.

sjd78:
Yep, PR #2025 will also resolve this issue.  The `CardHeader.selectableActions` have been aligned and the click handler moved to `onChange` instead of the `Card.onClick`.

PR #1951, opened 2024-06-11
:sparkles: [AnalysisWizard]  Language discovery changes 

ibolton336:
Resolves https://github.com/konveyor/tackle2-ui/issues/1950

UI Tests PR: 1136

Needs: https://github.com/konveyor/tackle-ui-tests/pull/1136

Includes: 
- Provide a "(Show All)" option to display all targets regardless of language tags. Adds a new component to handle this  new menu type. 
- Populate the options list from the Provider field of the fetched targets.
- Initially select the languages based on the tags, of tag category "Language", across the applications selected for analysis.
- Update TS type for Target to reflect the provider type changing to a string[].
- Covers the hub change to the task model described here:
>The application Analysis fields used to correlated to task by addon. This needs to be updated to correlate by kind == "analyzer" (when not blank) else fallback on addon == "analyzer". The fallback is needed to correlate tasks created in previous releases.
- Each target card should always display a label with it's provider. This will allow cards that belong to different providers to be differentiated when more than one language is selected.

Screenshot 2024-06-12 at 3 37 47 PM

### DIFF
--- a/client/src/app/components/SimpleSelectCheckbox.tsx
+++ b/client/src/app/components/SimpleSelectCheckbox.tsx
@@ -1,0 +1,130 @@
+import React from "react";
+import {
+  Select,
+  SelectOption,
+  SelectList,
+  MenuToggle,
+  Badge,
+  SelectOptionProps,
+  MenuToggleElement,
+} from "@patternfly/react-core";
+import spacing from "@patternfly/react-styles/css/utilities/Spacing/spacing";
+
+export interface ISimpleSelectBasicProps {
+  onChange: (selection: string | string[]) => void;
+  options: SelectOptionProps[];
+  value?: string[];
+  placeholderText?: string;
+  id?: string;
+  toggleId?: string;
+  toggleAriaLabel?: string;
+  selectMultiple?: boolean;
+  width?: number;
+  noResultsFoundText?: string;
+  hideClearButton?: false;
+}
+
+export const SimpleSelectCheckbox: React.FC<ISimpleSelectBasicProps> = ({
+  onChange,
+  options,
+  value,
+  placeholderText = "Select...",
+  id,
+  toggleId,
+  toggleAriaLabel,
+  width,
+}) => {
+  const [isOpen, setIsOpen] = React.useState(false);
+  const [selectedItems, setSelectedItems] = React.useState<string[]>([]);
+  const [selectOptions, setSelectOptions] = React.useState<SelectOptionProps[]>(
+    [{ value: "show-all", label: "Show All", children: "Show All" }, ...options]
+  );
+
+  React.useEffect(() => {
+    setSelectedItems(value || []);
+  }, [value]);
+
+  React.useEffect(() => {
+    const updatedOptions = [
+      { value: "show-all", label: "Show All", children: "Show All" },
+      ...options,
+    ];
+    setSelectOptions(updatedOptions);
+  }, [options]);
+
+  const onToggleClick = () => {
+    setIsOpen(!isOpen);
+  };
+
+  const onSelect = (
+    _event: React.MouseEvent<Element, MouseEvent> | undefined,
+    selectionValue: string | number | undefined
+  ) => {
+    const value = selectionValue as string;
+    if (value === "show-all") {
+      if (selectedItems.length === options.length) {
+        setSelectedItems([]);
+        onChange([]);
+      } else {
+        const allItemValues = options.map((option) => option.value as string);
+        setSelectedItems(allItemValues);
+        onChange(allItemValues);
+      }
+    } else {
+      if (selectedItems.includes(value)) {
+        const newSelections = selectedItems.filter((item) => item !== value);
+        setSelectedItems(newSelections);
+        onChange(newSelections);
+      } else {
+        const newSelections = [...selectedItems, value];
+        setSelectedItems(newSelections);
+        onChange(newSelections);
+      }
+    }
+  };
+
+  return (
+    <Select
+      role="menu"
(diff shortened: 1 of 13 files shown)


client/src/app/components/SimpleSelectCheckbox.tsx
```diff
@@ -0,0 +1,133 @@
...
+    ]
+  );
+
+  React.useEffect(() => {
+    setSelectedItems(value || []);
+  }, [value]);
+
+  React.useEffect(() => {
+    const updatedOptions = [
+      { value: "select-all", label: "Select All", children: "Select All" },
```
rszwajko, line 52: IMHO  "select" is not well defined here.  Above, we ask user "select one or more target options..." (target === card).  User may think that by selecting a provider/language they select a target. 
Perhaps we should switch to filtering-related wording i.e. "show all" , "Filter by provider" ....

client/src/app/pages/applications/analysis-wizard/set-targets.tsx
```diff
@@ -15,24 +16,53 @@ import { TargetCard } from "@app/components/target-card/target-card";
...
+        .map((tag) => {
+          return {
+            category: findCategoryForTag(tag.id),
+            tag,
+          };
+        })
+        .filter((tagWithCat) => tagWithCat?.category?.name === "Language")
+        .map((tagWithCat) => tagWithCat.tag.name)
+    )
+  ).filter((name) => name !== undefined);
```
rszwajko, line 62: `filter(Boolean)` ?

client/src/app/queries/tasks.ts
```diff
@@ -27,7 +28,16 @@ export const useFetchTasks = (
     select: (allTasks) => {
       const uniqSorted = allTasks
         .filter((task) =>
-          filters?.addon ? filters.addon === task.addon : true
+          // If there are any tasks with the addon field, we will still need to consider those older
```
rszwajko, line 31: I would move it to a separate PR. First, it seems not directly lined with the enhancement. Second, this is really should be re-written with server side filtering (https://github.com/konveyor/tackle2-ui/pull/1956)

ibolton336: Moved to #1959

client/src/app/components/SimpleSelectCheckbox.tsx
ibolton336: +1 Changed to show verbiage.

sjd78:
e2e ui testing is failing: https://github.com/konveyor/tackle2-ui/actions/runs/9504942845/job/26198861030?pr=1951#step:10:1822

sjd78 (commented):
First look makes sense, but there is an inconsistency with the Target.provider field.

ibolton336:
> First look makes sense, but there is an inconsistency with the Target.provider field.

Fixed.

client/src/app/components/SimpleSelectCheckbox.tsx
```diff
@@ -0,0 +1,132 @@
...
+  options,
+  value,
+  placeholderText = "Select...",
+  id,
+  toggleId,
+  toggleAriaLabel,
+  width,
+}) => {
+  const [isOpen, setIsOpen] = React.useState(false);
+  const [selectedItems, setSelectedItems] = React.useState<string[]>([]);
```
rszwajko, line 38: The selection state and options are maintained already by the parent (`provider` and  `languageOptions`).
Can we rely only on them and remove local state here?

client/src/app/pages/applications/analysis-wizard/set-targets.tsx
```diff
@@ -15,24 +16,53 @@ import { TargetCard } from "@app/components/target-card/target-card";
...
+            category: findCategoryForTag(tag.id),
+            tag,
+          };
+        })
+        .filter((tagWithCat) => tagWithCat?.category?.name === "Language")
+        .map((tagWithCat) => tagWithCat.tag.name)
+    )
+  ).filter(Boolean);
+
+  const [provider, setProvider] = useState(initialProviders);
```
rszwajko, line 64: renaming the field to `providers` would improve readability.

client/src/app/components/SimpleSelectCheckbox.tsx
ibolton336: I think it makes sense to have internal component level state here since we are manipulating it based on the showAll checkbox value.

ibolton336:
> Looks good! I've found one bug during smoke tests:
> 
> 1. filter by "Java" and select some cards
> 2. go back in the wizard
> 3. enter the target selection step again: filter is missing, cards cannot be de-selected

Looks like this bug shows up as a result of introducing checkboxes for unselected cards:

```
    isSelectable={!!cardSelected}

```

I'd like to fix this in another issue if possible. Need to investigate the PF component. Reverted the change for now & https://github.com/konveyor/tackle2-ui/issues/1252 will remain open.

rszwajko:
@ibolton336 
Integration tests are failing at selecting the card:
```
AssertionError: Timed out retrying after 8000ms: Expected to find element: `input[type="checkbox"]`, but never found it. Queried from:
              > cy.get(#target-card-Containerization)
```

ibolton336:
> @ibolton336 Integration tests are failing at selecting the card:
> 
> ```
> AssertionError: Timed out retrying after 8000ms: Expected to find element: `input[type="checkbox"]`, but never found it. Queried from:
>               > cy.get(#target-card-Containerization)
> ```
Fixed.

rszwajko (approved):

Issue #1929, opened 2024-05-31
Add the ability to individually view all files attached to a task on the Analysis details view


sjd78:
## Summary

For an application, the activity of the current / most recently run analysis task:
  - is viewable in a modal that show the task's activity log in a rich viewer
  - the user may select a merged view, displaying the log merged with all the file attachments
  - the rich viewer display component used allows for code coloring based on a file type we infer

The enhancement will make a few changes:
  1. The activity log view will allow viewing each attached file individually
  2. The activity log view will be moved a full page view

Enhancements reference:
  - https://github.com/konveyor/enhancements/issues/159 
  - https://github.com/konveyor/enhancements/pull/180

### Viewing each file individually

Existing functionality to keep:
  - The default view will be the task's log in yaml format
  - The format may be changed by clicking on the "JSON" tab on the top right
  - The default task log view may be displayed with all attached log files merged into the activity log block by selecting the "Merged" checkbox

Allow viewing all of the task's attached files by adding a select box to the top of the viewer.  The select box should:
  - List the files attached to the task
  - Allow selection of the file and loading it into the view
  - Allow selection of the default task log view

Additional functionality:
  - When an attached file is selected, the text formatting should be switched to "plain text"
  - A few well known file extensions could be detected and the language formatting loaded into the view (json, yaml).  However, this is not required in this enhancement and extra formatting can be added in future as the kinds of attachments are better known.

### View moved to a full page

The current view opens in a modal window.  In order to have more screen space to view the logs and provide selection of attached files, move the view from the modal to a full page.

  - Existing links remain and open the new page
  - The new page does not need an entry on the left nav
  - The URL for the new page should allow for the main left nav to retain focus (e.g. `/applications/123/task-view/456`)
  - Breadcrumbs on the new page should allow for easy navigation back to the previous page and show what application and task is being displayed
  - Note: Future work will be done where the user will have a distinction between different task types (e.g. analysis and language detection) so the breadcrumbs should consider this.

Some potential examples of the bread crumbs (including some potential future uses):
  - [Applications]() > Bear > [Analysis Details]()
  - [Applications]() > Bear > [Analysis Details]() > Attachment > [konveyor-analyzer.output]()
  - [Applications]() > Bear > [Discovery Details]()
  - [Applications]() > Bear > [Analysis Task]() > Current
  - [Applications]() > Bear > [Analysis Task]() > May 31 (123)

### Current State
Open the view from the kebab menu:
[...]

PR #1954, opened 2024-06-11
:sparkles: View files task attachments

rszwajko:
Resolves: https://github.com/konveyor/tackle2-ui/issues/1929

Functional changes:
1. document to be viewed is selected from the list (first action above
   the editor) or via URL
2. display language toggle based on languages supported by the document.
   For attachments the list consists of plain text and optionally a
   second language discovered based on file extension (YAML or JSON).
3. replace "merged" checkbox with an option in the select

Related refactorings:
1. split getTaskById() query into specialized queries:
   a) getTaskById() returning Promise<Task> for working with the object
   b) getTaskByIdAndFormat() returning Promise<string> for displaying
      the task as a formatted text
2. configure AnalysisDetails component to respond to 2 routes: existing
   details route and newly added /applications/:applicationId/analysis-details/:taskId/attachments/:attachmentId
3. persist merge mode in the URL as search param "merged"

### DIFF
--- a/client/src/app/components/simple-document-viewer/SimpleDocumentViewer.tsx
+++ b/client/src/app/components/simple-document-viewer/SimpleDocumentViewer.tsx
@@ -1,7 +1,6 @@
 import * as React from "react";
 import { CodeEditor, Language } from "@patternfly/react-code-editor";
 import {
-  Checkbox,
   EmptyState,
   EmptyStateIcon,
   EmptyStateVariant,
@@ -10,9 +9,14 @@ import {
 } from "@patternfly/react-core";
 
 import "./SimpleDocumentViewer.css";
-import { useFetchTaskByID } from "@app/queries/tasks";
+import {
+  useFetchTaskAttachmentById,
+  useFetchTaskByIdAndFormat,
+} from "@app/queries/tasks";
 import { RefreshControl } from "./RefreshControl";
 import { LanguageToggle } from "./LanguageToggle";
+import { AttachmentToggle } from "./AttachmentToggle";
+import { TaskAttachment } from "@app/api/models";
 
 export { Language } from "@patternfly/react-code-editor";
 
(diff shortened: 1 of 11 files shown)


rszwajko:
### Analysis details with default option pre-selected
image

### Typical task attachment (text only)
Screenshot from 2024-06-11 19-46-03

### Example of a task attachment that supports both plain text and JSON
The screenshot was taken in dev mode - under production conditions the file name would need to end with ".json".

Screenshot from 2024-06-11 19-46-32

client/src/app/components/simple-document-viewer/SimpleDocumentViewer.tsx
```diff
@@ -22,63 +26,133 @@ type ControlledEditor = {
...
+        Language.plaintext,
+        name.endsWith(".yaml") && Language.yaml,
+        name.endsWith(".json") && Language.json,
+      ].filter(Boolean),
+    })),
+  ];
+
+  const [documents, setDocuments] = React.useState([...configuredDocuments]);
+  const selectedId =
+    documents.find(({ isSelected }) => isSelected)?.id ?? "LOG_VIEW";
```
sjd78, line 132: What about just using `selectedId` as the state instead of keeping the documents in state?  You can memoize `documents` based on `attachments` and `selectedId` and keep things a bit simpler.  Simpler by just computing the document attributes in one place.

Else, the initial state can just be `React.useState(configuredDocuments)`.  No need to make a copy of the array like that.

client/src/app/pages/applications/analysis-details/AnalysisDetails.tsx
```diff
@@ -45,11 +67,34 @@ export const AnalysisDetails: React.FC = () => {
...
+                    }),
+                  },
+                ]
+              : []),
           ]}
         />
       </PageSection>
       <PageSection className="simple-task-viewer-container">
-        <SimpleDocumentViewer documentId={Number(taskId)} height="full" />
+        <SimpleDocumentViewer
```
sjd78, line 89: Maybe do it like on other pages?

https://github.com/konveyor/tackle2-ui/blob/c2f4f067b14eacf6742e0033e2244ff864d00085/client/src/app/pages/archetypes/archetypes-page.tsx#L362-L366

sjd78 (changes requested):
UX changes:
  - The attachment select drop down background should be white instead of the default transparent --- or may the page's content background color white (the other pages do this, see: https://github.com/konveyor/tackle2-ui/blob/c2f4f067b14eacf6742e0033e2244ff864d00085/client/src/app/pages/archetypes/archetypes-page.tsx#L362-L366)
  
  - When a file has a known file type (e.g. `settings.yaml`), select that file type for display by default instead of the "Text" one

Looks really good.  I have a few comments.

client/src/app/components/simple-document-viewer/SimpleDocumentViewer.tsx
```diff
@@ -22,63 +26,133 @@ type ControlledEditor = {
...
+}) => {
+  const { task, refetch: refetchTask } = useFetchTaskByIdAndFormat({
+    taskId,
+    format: currentLanguage === Language.yaml ? "yaml" : "json",
+    enabled:
+      !!taskId && (selectedId === "LOG_VIEW" || selectedId === "MERGED_VIEW"),
+    merged: selectedId === "MERGED_VIEW",
+  });
+
+  const isAttachment = typeof selectedId === "number";
```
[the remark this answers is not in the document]
rszwajko: It's convenient to have one ID for all documents - except that sometimes you need to differentiate. Type check seemed the simplest one but we might improve in follow-up.

client/src/app/components/simple-document-viewer/SimpleDocumentViewer.tsx
rszwajko: I've switched to `selectId` as you suggested - some things look simpler but at cost of extra checks if selected ID is "in range" (has a matching document).

rszwajko:
@jortel @JustinXHale @rromannissen 
Please take a look at [screenshots above](https://github.com/konveyor/tackle2-ui/pull/1954#issuecomment-2161319797).
The version currently under review has some differences compared to the enhancement - most notably handling the "merged" view has changed. I've put all options in one place. Let me know if you would prefer a checkbox as before.

JustinXHale:
> @jortel @JustinXHale @rromannissen Please take a look at [screenshots above](https://github.com/konveyor/tackle2-ui/pull/1954#issuecomment-2161319797). The version currently under review has some differences compared to the enhancement - most notably handling the "merged" view has changed. I've put all options in one place. Let me know if you would prefer a checkbox as before.

would the default view be the Log View? I dont have any objections.

rszwajko:
@JustinXHale 
> would the default view be the Log View?
>
yes

rromannissen:
LGTM!

Issue #1938, opened 2024-06-05
Create a Task Manager drawer


sjd78:
## Summary

With the addition of [automatic language discovery](https://github.com/konveyor/enhancements/issues/170), and task management ([enhancements RFE](https://github.com/konveyor/enhancements/issues/171), [UI issue](https://github.com/konveyor/tackle2-ui/issues/1931)), the UI can be enhanced to provide a global Task Manager Drawer that is similar to the form and function of a [PatternFly notification drawer](https://www.patternfly.org/components/notification-drawer/react-demos).  This enhancement will allow for easy access to a high level list of tasks as a notification drawer that does not hide the content of the page.

Enhancement references:
  - https://github.com/konveyor/enhancements/issues/171
  - https://github.com/konveyor/enhancements/issues/170

## Requirements

Base requirements:
  - Implement a Task Manager drawer following the notification drawer component style and anchored to the UI's masthead
  - The notification badge on the masthead should show a count of tasks that are pending or active
  - Clicking on the notification badge will open the Task Manager drawer
  - Within the drawer, the tasks to be displayed should be grouped by kind (e.g. language discovery or analysis)

Task Manager drawer contents:
  - Design follows the base [notification drawer](https://www.patternfly.org/components/notification-drawer/design-guidelines) design guidelines
  - Each task to display will follow a notification style in layout and function
  - Display the full set of currently pending/scheduled and running/in-flight tasks
  - Tasks should be in order they're created (***Note: verify the sort order***)
  - Allow actions on the tasks via a kebab menu

Task entries in the drawer should show a compact view including:
  - Icon for the task's state:
    - if the task is scheduled (PatternFly icon: PendingIcon)
    - if the task is running (PatternFly icon: InProgressIcon)
  - Task id
  - Task kind
  - Task's application's name
  - Task priority (the raw numeric value is ok, or use buckets of some definition like Low=0, Normal=10, High>10)

Task actions available on the task's kebab menu:
  - Adopt the same set of individual task actions from #1931
  - Viewing a task's activity log should be possible by clicking a link on the task table, possibly the task's id or an action icon
  - With functionality of #1929, viewing the log will open as a full page view
  - Browser back button navigation, or breadcrumb navigation back to the previous screen opened under the task manager drawer should be supported
  - **TBD**: If the task activity log is viewed, should the URL change, show the log, but also keep the task manager drawer open?

## HUB enhancements that may be needed

[...]

PR #1955, opened 2024-06-12
:sparkles: Implement the Task Manager drawer

sjd78:
Resolves: #1938

Add a queued tasks count badge and a task list drawer.
  - Add a `TaskManagerContext` to control the count indicator and visibility of the drawer.  This is a top level context so the task manager is available on all pages.  The context itself monitors the task queue report endpoint[^1] that drives the notification badge count.

  - Use the standard `NotificationDrawer` attached to the `Page` layout to render the task manager item drawer.

  - The `TaskNotificationBadge` is integrated to the application's `HeaderApp`.

  - The `TaskManagerDrawer` shows the list of currently "queued" tasks[^2] in task id descending sort order (most recently created tasks on the top).

  - Each task item is rendered in a "collapsed" state and after a click will render "expanded".  The expanded state include the started timestamp.

  - **_Note 1:_** Until fetch more / infinite scroll is implemented in the task manager drawer list, only the first 20 queued tasks will be displayed.

  - **_Note 2:_** Task actions need to be added to the task items.  These should reuse the same set of actions as the task table in #1957.

Related changes:
  - Update the `HeaderApp` to handle visibility of masthead toolbar items at the `ToolbarGroup` level.

  - Rename `SSOMenu` to `SsoToolbarItem`.

  - Updated the typing on a few task related API functions.

[^1]: https://github.com/konveyor/tackle2-hub/issues/641
[^2]: https://github.com/konveyor/tackle2-hub/pull/640


client/src/app/components/task-manager/TaskManagerContext.tsx
```diff
@@ -0,0 +1,45 @@
...
+
+export const TaskManagerProvider: React.FC<{ children: React.ReactNode }> = ({
+  children,
+}) => {
+  const { taskQueue } = useFetchTaskQueue();
+  const [isExpanded, setIsExpanded] = useState(false);
+
+  return (
+    <TaskManagerContext.Provider
+      value={{
```
rszwajko, line 36: a new wrapper object will be created here every time -  we could prevent this with `useMemo`

client/src/app/components/task-manager/TaskManagerDrawer.tsx
```diff
@@ -0,0 +1,256 @@
...
+            </NotificationDrawerList>
+          )}
+        </NotificationDrawerBody>
+      </NotificationDrawer>
+    );
+  }
+);
+TaskManagerDrawer.displayName = "TaskManagerDrawer";
+
+const TaskStateToIcon: React.FC<{ task: TaskManagerTask }> = ({ task }) =>
```
rszwajko, line 129: We need to sync on this - in https://github.com/konveyor/tackle2-ui/pull/1957 I've created a mapper function [taskStateToIcon](https://github.com/rszwajko/tackle2-ui/blob/3ed1dd44038319e3e4742b8d0234b32a83cdd603/client/src/app/components/Icons/taskStateToIcon.tsx#L10) based on icons used for [Analysis task state](https://github.com/rszwajko/tackle2-ui/blob/3ed1dd44038319e3e4742b8d0234b32a83cdd603/client/src/app/components/Icons/IconedStatus.tsx#L60).
Key differences:
1.  `Check` icon - here it's Ready, I've chosen Succeeded - Analyze status uses the same icon for both which is confusing for me. Since the Succeeded state is an end state I would prefer to use the icon there.
2.   `Pause` - here it's Postponed state, no icons in my version - IMHO the meaning is too different. I'd rather skip the icon.
3.  `InProgress` - here it's Running, I've used it for Pending - your version fits better
4.  `Pending` -  it fits nicely (no icon in my version)
5. `Unknown` - here it's fallback, in my version it's only used for "placeholder" states used when no state value is received from the backend (`No Task`, `not_supported`).

client/src/app/components/task-manager/TaskManagerDrawer.tsx
```diff
@@ -0,0 +1,256 @@
...
+      <TaskIcon />
+    </Tooltip>
+  );
+
+const TaskItem: React.FC<{
+  task: TaskManagerTask;
+  expanded: boolean;
+  onExpandToggle: (expand: boolean) => void;
+}> = ({ task, expanded, onExpandToggle }) => {
+  const starttime = dayjs(task.started ?? task.createTime);
```
rszwajko, line 157: 1. Using `createTime` as fallback may be confusing:  `started` seems to be set after the task actually started running. Together with `terminated` they show how long the task was executing.
2. in #1957 I've used  `dayjs(date).format() ` which is `date in ISO8601, without fraction seconds e.g. '2020-04-02T08:02:17-05:00'`.

client/src/app/components/task-manager/TaskManagerDrawer.tsx
```diff
@@ -0,0 +1,256 @@
...
+
+const useTaskManagerData = () => {
+  const [pageSize, setPageSize] = useState(PAGE_SIZE);
+  const increasePageSize = () => {
+    setPageSize(pageSize + PAGE_SIZE);
+  };
+
+  const { result, isFetching } = useServerTasks(
+    {
+      filters: [{ field: "state", operator: "=", value: "queued" }],
```
rszwajko, line 207: nice!

client/src/app/components/task-manager/TaskManagerDrawer.tsx
```diff
@@ -0,0 +1,256 @@
...
+    () =>
+      !result.data
+        ? []
+        : result.data.map(
+            (task) =>
+              ({
+                id: task.id ?? -1,
+                createUser: task.createUser ?? "",
+                updateUser: task.updateUser ?? "",
+                createTime: task.createTime ?? "", // TODO: date?
```
rszwajko, line 230: 1. with "flat" object approach it makes sense to pre-format the props - so I would use string-formatted dates
2. similar logic for retrieving/formatting values is used in [toCells mapper](https://github.com/rszwajko/tackle2-ui/blob/3ed1dd44038319e3e4742b8d0234b32a83cdd603/client/src/app/pages/tasks/tasks-page.tsx#L204) - we might won't to use common step at some point.

client/src/app/api/rest.ts
```diff
@@ -348,10 +349,15 @@ export const getTasks = () =>
...
-export const deleteTask = (id: number) => axios.delete<Task>(`${TASKS}/${id}`);
+export const deleteTask = (id: number) => axios.delete<void>(`${TASKS}/${id}`);
 
 export const cancelTask = (id: number) =>
-  axios.put<Task>(`${TASKS}/${id}/cancel`);
+  axios.put<void>(`${TASKS}/${id}/cancel`);
+
+export const getTaskQueue = (addon?: string): Promise<TaskQueue> =>
+  axios
+    .get<TaskQueue>(`${TASKS}/report/queue`, { params: { addon } })
```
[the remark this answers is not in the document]
sjd78: FWIW, the `addon` param is not actually used but was included since it was in the hub PR/code.

client/src/app/components/task-manager/TaskManagerDrawer.tsx
sjd78: Humm ---- I just needed to make sure there is a date to use to display when the task is "expanded".  I'm just displaying a "XX minutes ago" kind of time with the tooltip of the actual date.  So I'm not too worried about it

rszwajko: well ....if it's more "last change" then we could as well do `terminated ?? started ?? createTime`.

client/src/app/components/task-manager/TaskManagerDrawer.tsx
sjd78: Potential future work to harmonize the display formatting of the task's fields.

client/src/app/components/task-manager/TaskManagerDrawer.tsx
sjd78: Only "queued" tasks are displayed in the drawer so we only care about how long since they were created (or started)

client/src/app/components/task-manager/TaskManagerDrawer.tsx
sjd78: I opened https://github.com/konveyor/tackle2-ui/issues/1969 to capture this work once all the main task functions are merged.

client/src/app/components/task-manager/TaskManagerDrawer.tsx
sjd78: I opened https://github.com/konveyor/tackle2-ui/issues/1969 to capture this work once all the main task functions are merged.

client/src/app/components/task-manager/TaskManagerDrawer.tsx
sjd78: I opened https://github.com/konveyor/tackle2-ui/issues/1969 to capture this work once all the main task functions are merged.

rszwajko (approved):

client/src/app/components/task-manager/TaskManagerDrawer.tsx
rszwajko: Created  PR: https://github.com/konveyor/tackle2-ui/pull/1984

client/src/app/components/task-manager/TaskManagerDrawer.tsx
rszwajko: Created #1994 .

Issue #1931, opened 2024-05-31
Create a task manager page and table


sjd78:
## Summary

With the addition of [automatic language discovery](https://github.com/konveyor/enhancements/issues/170) and to give the user more insight and control over discovery and analysis tasks, create a new __Task Manager__ page and table.

Enhancement references:
  - https://github.com/konveyor/enhancements/issues/171
  - https://github.com/konveyor/enhancements/issues/170

## Requirements

Base requirements:
  - Navigation to the page is allowed from the left nav as "Task Manager" on the Migration perspective
  - Navigation to the page with filters applied (from at least the application table with the application filter already applied)
  - Standard table functionality with filtering, sorting, pagination and the right side details view tray
  - Each row in the table corresponds to a task in the system

Table columns to initially include at least:
  - Application linked to the Task
  - Task ID
  - Task Kind (analysis, discovery, others?)
  - Task Status (Scheduled, Pending, In progress, Success, Fail, others?)
  - Task Priority
  - User that created the task (this column has no meaning when running auth disabled)

Table filters and sorting to initially include at least:
  - Application name
  - Task id
  - Task kind
  - Task status
  - Task priority
  - User name

Actions that can be performed:
  - Cancel tasks created by the user (tasks created by other users cannot be cancelled)
  - Enable/disable preemption for the task created by the user (preemption allows the task system to cancel a running lower priority task in order to allow the task to run)

Actions that cannot be performed for Tasks:
  - Create a new task.  This must be done through the analysis wizard or similar.
  - Delete a task

## HUB enhancements that may be needed

See konveyor/tackle2-hub#640 for the filtering, sorting and pagination of the tasks endpoint to be used by the table.  This will reduce the number of task activity logs accessed and sent to keep the table performant.

## Mock ups
Base task manager table:
![Task Manager Table](https://github.com/konveyor/tackle2-ui/assets/3985964/5f53f772-71d2-492c-81d2-2af158a3087a)

sjd78:
@dymurray, @JustinXHale, @rromannissen -- for your consideration

JustinXHale:
@sjd78 couple of questions:

1. Will users have BULK Action capabilities? Delete/Cancel mainly. 
2. What are the priority levels? Low-High, 1-100?

sjd78:
@jortel, @mansam -- also for your consideration (sorry about the late tags)

sjd78:
> @sjd78 couple of questions:
> 
[...]

PR #1957, opened 2024-06-12
:sparkles: Task table

rszwajko:
Functional changes:
1. create task manager entry in the sidebar navigation list
2. place task page under /tasks route
3. present tasks in a server-filtered table
4. row actions supported: canceling, enable/disable preemption flag
5. use column management to hide optional columns: pod, started,
   terminated

Related features:
1. make id property required in Task type
2. switch task update endpoint to use patch method
3. provide icon-to-state mapping that preserves original state names
   which are required for server filtering

Resolves: https://github.com/konveyor/tackle2-ui/issues/1931

### DIFF
--- a/client/src/app/api/rest.ts
+++ b/client/src/app/api/rest.ts
@@ -373,6 +373,9 @@ export const getTaskQueue = (addon?: string): Promise<TaskQueue> =>
     .get<TaskQueue>(`${TASKS}/report/queue`, { params: { addon } })
     .then(({ data }) => data);
 
+export const updateTask = (task: Partial<Task> & { id: number }) =>
+  axios.patch<Task>(`${TASKS}/${task.id}`, task);
+
 export const createTaskgroup = (obj: Taskgroup) =>
   axios.post<Taskgroup>(TASKGROUPS, obj).then((response) => response.data);
 
--- a/client/src/app/components/Icons/taskStateToIcon.tsx
+++ b/client/src/app/components/Icons/taskStateToIcon.tsx
@@ -29,6 +29,7 @@ export const taskStateToIcon = (state?: TaskState) => {
     case "Pending":
       return <InProgressIcon />;
     case "Created":
+    case "QuotaBlocked":
     case "Running":
     case "Ready":
     case "Postponed":
--- a/client/src/app/components/Icons/taskStateToIcon.tsx
+++ b/client/src/app/components/Icons/taskStateToIcon.tsx
@@ -1,0 +1,38 @@
+import { TaskState } from "@app/api/models";
+import React from "react";
+import { Icon } from "@patternfly/react-core";
+import CheckCircleIcon from "@patternfly/react-icons/dist/esm/icons/check-circle-icon";
+import TimesCircleIcon from "@patternfly/react-icons/dist/esm/icons/times-circle-icon";
+import InProgressIcon from "@patternfly/react-icons/dist/esm/icons/in-progress-icon";
+import ExclamationCircleIcon from "@patternfly/react-icons/dist/esm/icons/exclamation-circle-icon";
+import UnknownIcon from "@patternfly/react-icons/dist/esm/icons/unknown-icon";
+
+export const taskStateToIcon = (state?: TaskState) => {
+  switch (state) {
+    case "not supported":
+    case "No task":
+      return <UnknownIcon />;
+    case "Canceled":
+      return <TimesCircleIcon />;
+    case "Succeeded":
+      return (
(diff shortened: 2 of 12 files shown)


rszwajko:
image

rszwajko:
@JustinXHale @rromannissen @jortel 
Take a look at [screenshots above](https://github.com/konveyor/tackle2-ui/pull/1957#issuecomment-2163785736) - I've added 3 extra columns that looked very useful to me. They are initially hidden and can be enabled using column management dialog. Please let me know if you would rather remove them.

JustinXHale:
> @JustinXHale @rromannissen @jortel Take a look at [screenshots above](https://github.com/konveyor/tackle2-ui/pull/1957#issuecomment-2163785736) - I've added 3 extra columns that looked very useful to me. They are initially hidden and can be enabled using column management dialog. Please let me know if you would rather remove them.

I think we're at the maximum number of columns we should be using, its borderline cramped. I think you've added relevant information. We could potentially save a little space using and provide some extra clarity using [Relative Time](https://www.patternfly.org/ux-writing/numerics/#relative-time) or [Timestamp](https://www.patternfly.org/components/timestamp)

rszwajko:
@JustinXHale 
> I think we're at the maximum number of columns we should be using,
>
Yep. The default columns take all space on a 1920x1080 screen. To show optional columns I had to hide some other columns. 

> We could potentially save a little space using and provide some extra clarity using [Relative Time](https://www.patternfly.org/ux-writing/numerics/#relative-time) or [Timestamp](https://www.patternfly.org/components/timestamp)
>
Relative time makes harder to compare `started` with `terminated`. Both dates are related but may end up in different "humanized" units. As for timestamp, it looks promising.

rromannissen:
I agree with @JustinXHale 's remarks. Overall LGTM!

client/src/app/components/Icons/taskStateToIcon.tsx
```diff
@@ -0,0 +1,38 @@
...
+        <Icon status={"danger"}>
+          <ExclamationCircleIcon />
+        </Icon>
+      );
+    case "Pending":
+      return <InProgressIcon />;
+    case "Created":
+    case "Running":
+    case "Ready":
+    case "Postponed":
```
jortel, line 34: + case QuotaBlocked

jortel:
This looks great!

client/src/app/components/Icons/taskStateToIcon.tsx
sjd78, line 1: Ref for #1969

client/src/app/pages/tasks/tasks-page.tsx
```diff
@@ -0,0 +1,348 @@
...
+    priority = 0,
+    policy,
+    createUser,
+    pod,
+    started,
+    terminated,
+  }: Task) => ({
+    id,
+    application: application.name,
+    kind,
```
sjd78, line 218: To best support old tasks, this should probably be `kind: kind ?? addon`

#1970 / #1959 needs to be merged before analyzer tasks will be created with the kind set instead of addon

client/src/app/pages/tasks/tasks-page.tsx
```diff
@@ -0,0 +1,348 @@
...
+                    <Tr key={task.id} {...getTrProps({ item: task })}>
+                      <TableRowContentWithControls
+                        {...tableControls}
+                        item={task}
+                        rowIndex={rowIndex}
+                      >
+                        {columnState.columns
+                          .filter(({ id }) => getColumnVisibility(id))
+                          .map(({ id }) => (
+                            <Td key={id} {...getTdProps({ columnKey: id })}>
```
sjd78, line 305: This is the column id right?  The key should probably have the task.id in there as well.

sjd78 (commented):
LGTM with 3 minor comments

client/src/app/api/rest.ts
```diff
@@ -367,6 +367,9 @@ export const deleteTask = (id: number) => axios.delete<Task>(`${TASKS}/${id}`);
 export const cancelTask = (id: number) =>
   axios.put<Task>(`${TASKS}/${id}/cancel`);
 
+export const updateTask = (task: Partial<Task> & { id: number }) =>
+  axios.patch<Task>(`${TASKS}/${task.id}`, task);
+
```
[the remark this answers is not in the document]
rszwajko: It's needed to fully use `patch` capability i.e. send a single property.

sjd78 (approved):
LGTM

PR #1961, opened 2024-06-14
:seedling: Update the `Task` TypeScript interface

sjd78:
Update the `Task` interface, and related, to follow the current definitions of Task REST resource[^1] on the hub.

Supports: #1955, #1956, #1957

[^1]: https://github.com/konveyor/tackle2-hub/blob/1366d95b62ac7717826fd9f64c9e268013107008/api/task.go#L725-L748

### DIFF
--- a/client/src/app/api/models.ts
+++ b/client/src/app/api/models.ts
@@ -330,7 +330,7 @@ export interface Task {
   events?: TaskEvent[];
   errors?: TaskError[];
   activity?: string[];
-  attached?: Attachment[];
+  attached?: TaskAttachment[];
 }
 
 export interface TaskPolicy {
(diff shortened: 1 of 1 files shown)


client/src/app/api/models.ts
```diff
@@ -307,18 +307,62 @@ export type TaskState =
...
+export interface TaskError {
+  severity: string;
+  description: string;
+}
+
+export interface Attachment {
+  id: number;
+  name?: string;
+  activity?: number;
 }
```
rszwajko, line 366: I'd rather keep "Task" prefix  like in https://github.com/rszwajko/tackle2-ui/blob/e4b6895a65310d95edbb186472623c533045973d/client/src/app/api/models.ts#L325

sjd78: Ok

rszwajko (approved):

PR #1962, opened 2024-06-14
:bug:  Use initialSort property

rszwajko:
Before, initialSort property passed as parameter to useTableControlState() hook was not used as default value. Instead the defaults were calculated based on available sortable columns.


sjd78 (approved):
LGTM

PR #1963, opened 2024-06-17
:bug: Disabling useFetchCache query if RWX_SUPPORTED=false

mguetta1:
Resolves: https://issues.redhat.com/browse/MTA-3699

### DIFF
--- a/client/src/app/pages/repositories/Mvn.tsx
+++ b/client/src/app/pages/repositories/Mvn.tsx
@@ -110,7 +110,7 @@ export const RepositoriesMvn: React.FC = () => {
                         isAriaDisabled={inputDisabled}
                         onClick={() => setIsConfirmDialogOpen(true)}
                       >
-                        {isRWXSupported && isFetching ? (
+                        {isFetching ? (
                           <Text>
                             Loading...
                             <Spinner
--- a/client/src/app/pages/repositories/Mvn.tsx
+++ b/client/src/app/pages/repositories/Mvn.tsx
@@ -110,7 +110,7 @@ export const RepositoriesMvn: React.FC = () => {
                         isAriaDisabled={inputDisabled}
                         onClick={() => setIsConfirmDialogOpen(true)}
                       >
-                        {isFetching ? (
+                        {isRWXSupported && isFetching ? (
                           <Text>
                             Loading...
                             <Spinner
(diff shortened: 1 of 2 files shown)


sjd78 (commented):
@mguetta1 - Can you provide some context for this change?  Is there an issue or ticket for this?  At face value it looks ok, but I'm not sure how the `isRWXSupported` flag is meant to interact with the loading spinner.

mguetta1:
> @mguetta1 - Can you provide some context for this change? Is there an issue or ticket for this? At face value it looks ok, but I'm not sure how the `isRWXSupported` flag is meant to interact with the loading spinner.

Hi @sjd78, @dymurray 
I just opened a Jira ticket for this issue: https://issues.redhat.com/browse/MTA-3699

rszwajko (changes requested):
The loading state seems the consequence of the query being active.  IMHO if the entire feature is disabled we should disable the query as well.
React query uses `enable`  flag [for this purpose](https://tanstack.com/query/v4/docs/framework/react/guides/disabling-queries).  We use in several places i.e. [here](https://github.com/konveyor/tackle2-ui/blob/f389539cb2c6ec7720dad107227dc9edcbd77ca7/client/src/app/queries/tasks.ts#L240).

mguetta1:
> The loading state seems the consequence of the query being active. IMHO if the entire feature is disabled we should disable the query as well. React query uses `enable` flag [for this purpose](https://tanstack.com/query/v4/docs/framework/react/guides/disabling-queries). We use in several places i.e. [here](https://github.com/konveyor/tackle2-ui/blob/f389539cb2c6ec7720dad107227dc9edcbd77ca7/client/src/app/queries/tasks.ts#L240).

Done in https://github.com/konveyor/tackle2-ui/pull/1963/commits/ee7c7a2220ea4d626a90c807daf443c33a95cff3
@rszwajko better to pass `isRWXSupported` as an argument? I saw that `useFetchCache` is used only in `client/src/app/pages/repositories/Mvn.tsx`

rszwajko (approved):
LGTM! If there will be another use case for  `useFetchCache` then we can always refactor.

rszwajko:
@mguetta1 please edit the title to better reflect the fix.

PR #1966, opened 2024-06-18
:bug: Use exact match for filtering by application name

rszwajko:
Exact match ensures that active item is on page.

Example usage: /applications?activeItem=6&filters={"name":["bookserver-app"]}

Resolves: #1974


rszwajko:
### Before - contains-type matching 
Screenshot from 2024-06-18 14-54-21

### After - exact match
Screenshot from 2024-06-18 14-54-07

ibolton336 (approved):

sjd78 (approved):
What is weird is that the filter input box doesn't allow general text input.  It only allows selection of existing names.  So in that context, an exact match looks better than the `getItemValue` filter of "name contains the filter text".

In future it may be nice to support a glob input so you can do "Foo*" and pickups all the applications that start with "Foo" without having to select each one.

Issue #1967, opened 2024-06-18
:sparkles: Prefix application binary coordinates with mvn://.


jortel:
Prefix application binary coordinates with `mvn://`.
Current:
```
io.konveyor.demo:customers-tomcat:0.0.1-SNAPSHOT:war
```
Changed to:
```
mvn://io.konveyor.demo:customers-tomcat:0.0.1-SNAPSHOT:war
```

PR #1968, opened 2024-06-19
:sparkles: Add mvn:// prefix to application binary value in post request

ibolton336:
Resolves: https://github.com/konveyor/tackle2-ui/issues/1967

Note: validation is not necessary since the ui already breaks up the input into separate for fields and composes the binary coordinates under the hood.


rszwajko (approved):

sjd78 (approved):
LGTM

Issue #1970, opened 2024-06-20
Update Task creation and handling to prefer the kind field over the addon field


sjd78:
Based on comments in #1950 and #1951, any of the UI code touching Tasks should prefer the use of `kind` over `addon`.

-----

The application `Analysis` fields used to correlated to task by addon.  This needs to be updated to correlate by kind == "analyzer" (when not blank) else fallback on addon == "analyzer".  The fallback is needed to correlate tasks created in previous releases.

_Originally posted by @jortel in https://github.com/konveyor/tackle2-ui/issues/1950#issuecomment-2159284083_

PR #1971, opened 2024-06-20
:bug: Prefer `Task` field `kind` over `addon`

sjd78:
Resolves: #1970

  - When creating a new `Taskgroup`, specify `kind` instead of `addon`

  - Process Tasks where `kind` may be missing and `addon` is given instead (i.e. tasks created before this change)

  - Make `Task.id` and `Taskgroup.id` required, `New<>` as needed


sjd78:
Replaces #1959

ibolton336 (approved):

Issue #1975, opened 2024-06-21
[BUG] Add a method to view a task's log from the task table


sjd78:
From the task manager table, users need a way to jump to the analysis details for any task.

This can be:
  - A link from the task id (or any other column that makes sense)
  - An action from the kebab menu

No changes on navigation between page are required, but are ok if necessary.

sjd78:
Follow up as a gap from #1938

PR #1980, opened 2024-06-24
:bug: Add task detail links to tasks page

rszwajko:
1. extract base component to re-use for Analysis Details and Task Details
2. add 2 new routes: 
   a) /tasks/:taskId 
   b) /tasks/:taskId/attachments/:attachmentId
3. user can enter details screen by: 
   a) task status name (link) 
   b) "Task details" action (per row kebab actions)
4. persist page state in session storage for Applications
    and Task Manager pages to fix problems with returning
    from task details page.

Resolves: #1975


rszwajko:
### Details screen available via link and action
image

### Task details - log view
image

### Task details - attachment view
image

sjd78 (changes requested):
### What looks good?
  - code changes look good
  - navigation from the table looks good
  - URLs look good
  - attached files and merged views look good

### Breadcrumbs
What about having them include the "Task ${task.id}" text included?  So for a given URL like "/tasks/7" they look like:

[Tasks]() > [Task 7]() > Task details


### Title
The title shows the name of the task.  That is not enough to positively identify the task.  If you create an application "Bear" w/o sources initially, and then edit it to include sources, there will be multiple tasks named "Bear-language-discovery".

Maybe a title more like:

"Task details for task ${task.id}, ${task.name}"

rszwajko:
@sjd78 
Below the new wording.  Note that `Task 9` is last segment and is inactive (not a link) because it's the current page. Active links denote previous pages. At least this is my understanding of [PF design guide ](https://www.patternfly.org/components/breadcrumb/design-guidelines/).
 
image

rszwajko:
@sjd78 
one important fix was added - persisting page state in session storage:paging, sorting and filtering (Application page only). This should fix problems with returning back to main page from details pages.

client/src/app/pages/tasks/tasks-page.tsx
```diff
@@ -61,7 +61,11 @@ export const TasksPage: React.FC = () => {
 
   const tableControlState = useTableControlState({
     tableName: "tasks-table",
-    persistTo: { filter: "urlParams" },
+    persistTo: {
+      filter: "urlParams",
```
sjd78, line 65: urlParams instead of sessionStorage like the other 2?

rszwajko: It's quite verbose and doesn't seem useful for copy-pasting manually i.e. to chat/mail.
Technically it would work. If you prefer this solution we can give a try.

sjd78: Really I just noticed the difference between application table and this table -- sessionStorage for filter in application table, and current in the filter here.  I don't really have a strong feeling either way.

rszwajko: I would propose to test mixed approach and change later on if necessary.

sjd78 (approved):
LGTM

Issue #1969, opened 2024-06-20
[BUG] Harmonize the render of task data between the task drawer and the task table


sjd78:
Followup to: #1955, #1957 

The task manager drawer and the task table both have code to render tasks.  We should look at the code and extract common work to maintain visual content consistency across the UI.

This can include:
  - Status icons
  - Status icon tooltips
  - create, start, terminate date formats and intervals

rszwajko:
@sjd78 
Created #1994  as part of this issue. The improvement here is to be better at displaying local time. The format will be still different then in the drawer because use cases are different - in the table we need an easy way to establish a sort order which makes ISO 8601 is a better choice. In the drawer we can keep the humanized/localized formats.

rszwajko:
Moved the remaining part to a separate issue to be handled in the next release cycle.

PR #1994, opened 2024-07-03
:bug: Drop UTC offset for local time used in tasks table

rszwajko:
The API provides date/time in UTC. For user convenience we display it in the browser time zone. Before this PR, UTC offset for the local time zone was included.

The format used is a subset of ISO 8601 which allows for predictable sorting in the table.
Format string: "YYYY-MM-DD HH:mm:ss".
Example date: "2024-07-03 13:21:44".

Part-of: #1969


rszwajko:
### Before
Screenshot from 2024-07-03 14-04-03
### After
Screenshot from 2024-07-03 15-02-08

sjd78 (approved):
LGTM

PR #1996, opened 2024-07-03
:seedling: Add data attributes to table rows

sjd78:
To enable easier and safer selection of table rows by testing code, add some data attributes to the tr:

  - `data-item-id` - Holds the `idProperty` value of a table's items.  This is typically the item's id number but may be a frontend generated ID in some cases.

  - `data-item-name` - If the table property `dataNameProperty` is provided, hold the name property of a table's item.

For any tables that use `/use.?*TableControls\(/`, add a `dataNameProperty` and if necessary, adjust the `idProperty`.  A few tables for items without a good item id and name property were left with just the `idProperty`.

### DIFF
--- a/client/src/app/pages/applications/manage-imports-details/manage-imports-details.tsx
+++ b/client/src/app/pages/applications/manage-imports-details/manage-imports-details.tsx
@@ -74,6 +74,7 @@ export const ManageImportsDetails: React.FC = () => {
   const tableControls = useLocalTableControls({
     tableName: "manage-imports-details",
     idProperty: "Application Name",
+    dataNameProperty: "Application Name",
     items: imports || [],
     columnNames: {
       name: t("terms.name"),
--- a/client/src/app/pages/controls/business-services/business-services.tsx
+++ b/client/src/app/pages/controls/business-services/business-services.tsx
@@ -83,7 +83,8 @@ export const BusinessServices: React.FC = () => {
 
   const tableControls = useLocalTableControls({
     tableName: "business-services-table",
-    idProperty: "name",
+    idProperty: "id",
+    dataNameProperty: "name",
     items: businessServices,
     columnNames: {
       name: t("terms.name"),
--- a/client/src/app/hooks/table-controls/types.ts
+++ b/client/src/app/hooks/table-controls/types.ts
@@ -1,6 +1,10 @@
 import { TableProps, TdProps, ThProps, TrProps } from "@patternfly/react-table";
 import { ISelectionStateArgs, useSelectionState } from "@migtools/lib-ui";
-import { DisallowCharacters, DiscriminatedArgs } from "@app/utils/type-utils";
+import {
+  DisallowCharacters,
+  DiscriminatedArgs,
+  KeyWithValueType,
+} from "@app/utils/type-utils";
 import {
   IFilterStateArgs,
   ILocalFilterDerivedStateArgs,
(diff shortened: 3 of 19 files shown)


rszwajko (commented):
Looks good!

client/src/app/pages/controls/business-services/business-services.tsx
```diff
@@ -83,7 +83,8 @@ export const BusinessServices: React.FC = () => {
 
   const tableControls = useLocalTableControls({
     tableName: "business-services-table",
-    idProperty: "name",
+    idProperty: "id",
```
rszwajko, line 86: good catch! please mention those fixes in the commit message (or extract to a small PR)?

rszwajko (approved):

client/src/app/pages/controls/business-services/business-services.tsx
sjd78: > good catch! please mention those fixes in the commit message (or extract to a small PR)?

I mentioned that kind of change in the description as "and if necessary, adjust the idProperty".  Could be called out more explicitly.

Also interesting that it really didn't do anything bad or wrong with any behavior.  Apparently the name on the entity is enforced unique so it worked just fine.  But now both bits of info are available on the data attributes. :-)

client/src/app/pages/applications/manage-imports-details/manage-imports-details.tsx
```diff
@@ -74,6 +74,7 @@ export const ManageImportsDetails: React.FC = () => {
   const tableControls = useLocalTableControls({
     tableName: "manage-imports-details",
     idProperty: "Application Name",
```
mguetta1, line 76: `idProperty: "id"`

client/src/app/pages/applications/manage-imports-details/manage-imports-details.tsx
```diff
@@ -74,6 +74,7 @@ export const ManageImportsDetails: React.FC = () => {
...
+    dataNameProperty: "Application Name",
```
mguetta1, line 77: I am probably missing something...
Should it be one of the `columnNames` value?

mguetta1 (changes requested):

client/src/app/pages/applications/manage-imports-details/manage-imports-details.tsx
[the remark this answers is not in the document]
sjd78: (resolving this comment and responding the other comment -- it will be the same discussion)

client/src/app/pages/applications/manage-imports-details/manage-imports-details.tsx
sjd78: (resolving this comment and responding the other comment -- it will be the same discussion)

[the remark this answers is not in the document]
sjd78: (resolving this comment and responding the other comment -- it will be the same discussion)

client/src/app/pages/applications/manage-imports-details/manage-imports-details.tsx
sjd78: Also, while the `columnNames` for the table properties often line up with the actual item field name, there is no requirement that it do so.  In some tables, the columns are "UI synthetic" being derived from multiple data sources.

That's just a long way of saying the `idProperty` and `dataNameProperty` values do not need to have a corresponding entry in `columnNames`.

mguetta1 (approved):

PR #2005, opened 2024-07-10
:bug: Add tooltips for Priority and Preemption columns

rszwajko:
Resolves: https://issues.redhat.com/browse/MTA-3235

### DIFF
--- a/client/public/locales/en/translation.json
+++ b/client/public/locales/en/translation.json
@@ -510,6 +510,10 @@
       "save": "Failed to save {{type}}."
     }
   },
+  "tooltip": {
+    "priority": "Tasks priority(Low-High). Impacts tasks scheduling policy.",
+    "preemption": "If enabled, allows the scheduler to cancel a running task and free the resources for higher priority tasks."
+  },
   "validation": {
     "email": "This field requires a valid email.",
     "max": "This field must be less than {{value}}.",
--- a/client/src/app/pages/tasks/tasks-page.tsx
+++ b/client/src/app/pages/tasks/tasks-page.tsx
@@ -13,7 +13,15 @@ import {
   ToolbarContent,
   ToolbarItem,
 } from "@patternfly/react-core";
-import { Table, Tbody, Th, Thead, Tr, Td } from "@patternfly/react-table";
+import {
+  Table,
+  Tbody,
+  Th,
+  Thead,
+  Tr,
+  Td,
+  ThProps,
+} from "@patternfly/react-table";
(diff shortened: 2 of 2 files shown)


client/src/app/pages/tasks/tasks-page.tsx
```diff
@@ -189,6 +190,11 @@ export const TasksPage: React.FC = () => {
     columnState,
   } = tableControls;
 
+  const tooltips: Record<string, ThInfoType> = {
+    priority: { tooltip: t("tooltip.priority") },
+    preemption: { tooltip: t("tooltip.preemption") },
+  };
```
sjd78, line 196: The tooltip keys you added are at `wizard.tooltip.*`

Probably need to create a top level "tooltip" section, or move it to the "message" section.

client/src/app/pages/tasks/tasks-page.tsx
```diff
@@ -189,6 +190,11 @@ export const TasksPage: React.FC = () => {
     columnState,
   } = tableControls;
 
+  const tooltips: Record<string, ThInfoType> = {
```
sjd78, line 193:
```suggestion
  const tooltips: Record<string, ThProps["info"]> = {
```

It's easy to import `ThProps` and then reference the info prop.  Then there are no oddball imports required.

sjd78 (changes requested):

sjd78 (approved):
LGTM

rszwajko:
Screenshot from 2024-07-11 17-07-10
Screenshot from 2024-07-11 17-06-46

client/public/locales/en/translation.json
```diff
@@ -510,6 +510,10 @@
       "save": "Failed to save {{type}}."
     }
   },
+  "tooltip": {
+    "priority": "Tasks priority(Low-High). Impacts tasks scheduling policy.",
```
mguetta1, line 514: Hi, I am not sure about this description regarding the priority value.The values can range from zero to any positive whole number.
`Low-High` doesn't say much to me. wdyt?

mguetta1: Suggestion: Please remove `(Low-High)`. The user can check the release notes documentation that says: `Priority: Priority of the task. The value is from zero to any positive integer. The higher the value in this column, the higher the priority of the task.`

Or use something like: `Tasks priority, 0 to any positive integer (Low-High). Impacts tasks scheduling policy.` as the tooltip

sjd78: Change made in #2028

PR #2006, opened 2024-07-11
:bug: Filter credentials by createUser property

rszwajko:
Drop conditional rendering as it clashes with column type definitions:
the dynamic column is interpreted as optional property which is not
allowed by the current typing in the table hooks.

Resolves: https://issues.redhat.com/browse/MTA-3143
Co-authored-by: Ian Bolton <ibolton@redhat.com>


ibolton336 (approved):

sjd78 (approved):
LGTM, and I don't mind the user being there even in auth disabled mode when it'll only ever show "admin.noauth".

Issue #2007, opened 2024-07-11
[BUG] Cannot edit an application that has a credential/identity assigned


sjd78:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Konveyor version

0.5-beta.1

### Priority

Critical

### Current Behavior

If an application has a source credentials or a maven settings assigned, the application cannot be edited.

The edit modal opens and changes can be made, but the Save action will fail with an error 400.

### Expected Behavior

Save should complete successfully.

### How Reproducible

Always (Default)

### Steps To Reproduce

1. Create an application
2. Create a credential
3. Manage the credentials for the application created in 1 and select the credential created in 2
4. Edit the application
5. Make a change anywhere and click 'Save'

### Environment

```markdown
Any
```

sjd78:
This is a regression from #1985

PR #2008, opened 2024-07-11
:bug: Fix application edit data source

sjd78:
Resolves: #2007
Resolves: https://issues.redhat.com/browse/MTA-3232

Summary of changes:
  - The application edit form needs to use the base `Application` data and not the `DecoratedApplication` as its basis for generating an update request.  Extra information added by the decoration will be unknown to the hub endpoint and will cause issues.

  - All references to `Application` in the application table code have been updated to `DecoratedApplication`

  - Any adjustments needed when a `DecoratedApplication` is passed to a child component as an `Application` have been made

  - The dereferencing part of decoration has been moved to a `direct` container.  This will allow the reuse of the `Application` property name without replacing it.

### DIFF
--- a/client/src/app/pages/applications/applications-table/applications-table.tsx
+++ b/client/src/app/pages/applications/applications-table/applications-table.tsx
@@ -70,7 +70,7 @@ import {
 
 // Queries
 import { getArchetypeById, getAssessmentsByItemId } from "@app/api/rest";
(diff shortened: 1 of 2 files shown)


rszwajko (approved):
Code looks good! No problems in smoke tests.

Issue #2017, opened 2024-07-17
[BUG] Application with analysis status of "Completed With Error" cannot be analyzed.


jortel:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Konveyor version

latest

### Priority

Critical

### Current Behavior

Application with analysis status of "Completed With Error" cannot be analyzed.  When checked the [Analyze] button remains disabled.

### Expected Behavior

Should be able to re-analyze.

### How Reproducible

Always (Default)

### Steps To Reproduce

1. Create an application
2. Analyze it.  For example: Daytrader source+deps  EAP, Containerization.
3. After complete-with-error
4. Try to re-analyze.

### Environment

```markdown
- OS:
```

PR #2018, opened 2024-07-17
:bug: Update task terminal state check to include "SucceededWithErrors"

sjd78:
Resolves: #2017

The check used to only allow starting an analysis if an application's analysis task is still running
(i.e. not in a terminal state) needs to also consider the frontend synthetic state `SucceededWithErrors` as a non-running / terminal state.


jortel (approved):
lgtm

PR #2019, opened 2024-07-17
:bug: Address targets falling out of sync after label selection changes

ibolton336:
Resolves: https://issues.redhat.com/browse/MTA-3180

- Currently, the set-options screen allows edit of target labels which can be set manually on the Advanced options step or by way of selecting a target card on the target card step. This PR brings the form labels and selectedTarget form values in sync by updating the selected targets to reflect any form label changes in the set options page. This fixes the mta-3180 scenario by causing validation to reflect the now accurate state of the selectedTargets form value. 

https://github.com/user-attachments/assets/d580f857-8236-4e30-bf28-ebd3a4c73792


sjd78 (approved):
LGTM

Issue #2022, opened 2024-07-23
[BUG] Make OpenJDK 21 the default rather than 11


dymurray:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Konveyor version

v0.5.0-beta.3

### Priority

Minor

### Current Behavior

When selecting openjdk as the target it defaults to 11

### Expected Behavior

I want it to default to 21

### How Reproducible

Always (Default)

### Environment

```markdown
- OS:
```

Issue #2010, opened 2024-07-12
[BUG] Opening a version dropdown de-selects analysis target


rszwajko:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Konveyor version

0.5beta

### Priority

Major

### Current Behavior

Attempt to choose a target version i.e. OpenJDK 21 results in the target being silently de-selected.
User needs to select it  again.
![Screenshot from 2024-07-12 13-28-33](https://github.com/user-attachments/assets/c8f1f1f3-3fc6-4c68-90dd-d163e56aea5e)

### Expected Behavior

Choosing a version should not impact the selection.

### How Reproducible

Always (Default)

### Steps To Reproduce

1. enter target selection step and chose language Java
2. chose a target that allows to chose a version - test on OpenJDK target
3. click the card to select the target
4. click the dropdown to open a list (at this moment the selection is removed from card)
5. click inside the list to chose a version - tested with OpenJDK 21

Issue #1981, opened 2024-06-25
[BUG] Custom migration target: The description textbox exceeds the target card


mguetta1:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Konveyor version

0.5.0-alpha.1

### Priority

Major

### Current Behavior

![Selection_580](https://github.com/konveyor/tackle2-ui/assets/60384172/8883172d-16e2-45fd-8c34-0c54a0a63ece)

### How Reproducible

Always (Default)

### Environment

```markdown
- OS: Fedora 36
- Google Chrome: 126.0.6478.114
```

rromannissen:
Reproduced in 0.5.0-alpha.2

PR #2025, opened 2024-07-24
:bug: Fix target card behavior and rendering

sjd78:
Resolves: #2022
Resolves: #2010
Resolves: #1981
Resolves: #1252
Resolves: https://issues.redhat.com/browse/MTA-3094

Target card behavior and rendering on the target selection step of the analysis wizard and on the custom migrations target page have been updated:
  - All of the card contents render inside the card itself (#1981 / MTA-3094)
  - When changing the specific label selection for a target, the card selection will not changes (#2010)
  - The target label select now sorts the labels in numeric natural sort descending order.  `OpenJDK 21` will be initially selected and appear above `OpenJDK 17` in the select list. (#2022)
  - Layouts used on the card have been refactored to use base layouts instead of the `EmptyState` component
  - The target label select box has been updated to the current `SimpleSelectBasic` component for current PF5 alignment
  - When card selection is enabled (on the target selection step of the analysis wizard), the selection checkbox is always displayed and the normal `Card` hover styles have been disabled.
  - The `CardHeader.selectableActions` have been aligned to current PF and the click handler moved to `onChange` instead of the `Card.onClick`.  The `selectableActionId` needs to be unique on the page for the select handling to work properly.

Some refactoring on the `SetTargets` component used by the analysis wizard have been made to simplify rendering code.

### DIFF
--- a/client/src/app/components/target-card/target-card.css
+++ b/client/src/app/components/target-card/target-card.css
@@ -2,9 +2,25 @@
   background: none;
 }
 
-.select-card__component__empty-state {
-  padding: 0 !important;
+/* do NOT change background-color or box-shadow on hover with the TargetCard is selectable */
+.target-card.pf-m-selectable {
+  --pf-v5-c-card--m-selectable--hover--BackgroundColor: var(
+    --pf-v5-c-card--BackgroundColor
+  );
+  --pf-v5-c-card--m-selectable--hover--BoxShadow: var(
+    --pf-v5-c-card--BoxShadow
+  );
 }
+
+/*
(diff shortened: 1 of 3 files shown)


sjd78:
Custom migration targets now:
screenshot-localhost_9000-2024 07 24-00_56_57

ibolton336 (changes requested):
Still seeing the bug outlined here when navigating back in the wizard.
https://github.com/konveyor/tackle2-ui/pull/1951#pullrequestreview-2125884981

ibolton336 (approved):
Awesome job identifying that bug.

PR #2032, opened 2024-07-29
:bug: Add application risk filter of "Unassessed"

sjd78:
Resolves: https://issues.redhat.com/browse/MTA-2816

Application risk options are one of:
  - High (red)
  - Medium (yellow)
  - Low (green)
  - Unknown (unknown)
  - Unassessed (unassessed)

Added the __Unassessed__ option to the risk filter list on the application table to have a complete
set of risk filters.

Note: To help test risk levels, the questionnaire `hack/import-questionnaire/assign-risk.yaml` has been added.  The single question has a 1 to 1 mapping between answer and risk level.  As long as that questionnaire is the only one active on the system, it works well to quickly assign risks to applications.

### DIFF
--- a/client/src/app/pages/applications/applications-table/applications-table.tsx
+++ b/client/src/app/pages/applications/applications-table/applications-table.tsx
@@ -490,6 +490,7 @@ export const ApplicationsTable: React.FC = () => {
           { value: "yellow", label: "Medium" },
           { value: "red", label: "High" },
           { value: "unknown", label: "Unknown" },
+          { value: "unassessed", label: "Unassessed" },
         ],
         getItemValue: (item) => item.risk ?? "",
       },
--- a/client/src/app/pages/applications/applications-table/applications-table.tsx
+++ b/client/src/app/pages/applications/applications-table/applications-table.tsx
@@ -17,7 +17,12 @@ import {
   Modal,
   Tooltip,
 } from "@patternfly/react-core";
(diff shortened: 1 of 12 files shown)


client/src/app/pages/applications/applications-table/applications-table.tsx
```diff
@@ -499,6 +499,7 @@ export const ApplicationsTable: React.FC = () => {
           { value: "yellow", label: "Medium" },
           { value: "red", label: "High" },
           { value: "unknown", label: "Unknown" },
+          { value: "unassessed", label: "Unassessed" },
         ],
         getItemValue: (item) => item.risk ?? "",
```
rszwajko, line 504: it seems the problem is caused by unknown value send from the server - we could add some extra protection here to prevent this happening in the future

sjd78: It is also confusing that a risk of "Unknown" is an actual valid thing to pick

And "Unassessed" just means the assessments have not yet been completed.

Both are actual valid distinct values.

sjd78: I added normalize code to the current version to make sure we only filter match on valid and known Risk type values.

rszwajko (approved):
Looks good! Thanks for sharing the test questionnaire!

PR #1698, opened 2024-02-22
:sparkles: Add url filter to drawer results for dependency apps

ibolton336:
Resolves https://issues.redhat.com/browse/MTA-2008
&& 
https://github.com/konveyor/tackle2-ui/issues/1498

PR #2033, opened 2024-07-30
:bug: Persist filters in URL for applications table

rszwajko:
Fix regression after: https://github.com/konveyor/tackle2-ui/pull/1980
Resolves: https://issues.redhat.com/browse/MTA-1390

Initial filter values work best if the table is not persisting state between reloads (default strategy, component state). If the table uses session or local storage then any existing filter overwrites the default value. From the end user perspective the functionality works only on the first load with a clean browser (no previous state).

Switching to URL filter persistence provides the same functionality and ensures we have only one source of filters.

Reference-Url:  https://github.com/konveyor/tackle2-ui/pull/1698


sjd78 (approved):
LGTM

Issue #2009, opened 2024-07-12
[BUG] More selected languages then available in the Analysis Wizard


rszwajko:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Konveyor version

0.5beta

### Priority

Minor

### Current Behavior

A bug shows 4 languages although only one is available:
![Screenshot from 2024-07-12 13-26-41](https://github.com/user-attachments/assets/554dd70c-cbcd-4db4-8e09-25e8ece8d286)

### Expected Behavior

Only available languages are included.

### How Reproducible

Always (Default)

### Steps To Reproduce

1. use default settings - only Java provider 
2. create an app that uses multiple languages - found on https://github.com/sonatype/nexus-public
3. wait until language discovery finished successfully
4. start Analysis Wizard and go to target selection step

sjd78:
/triage accepted
/priority normal

The count on the select box shows the total number of languages tagged across all selected applications.

The drop down shows the available target languages.

Those numbers will not match if a tag exists that does not have a matching target's language.

The count should be based on the selectable options and not the tag count.

PR #2045, opened 2024-08-02
:bug: Use standard filters for languages in the Set Targets step

rszwajko:
Before, in the Set Targets step, if the languages were not pre-selected automatically then no cards were visible. In order to see target cards user had to select at least one language.
After this fix,  all target cards will be visible.

Additional changes:
1. wait (display a spinner) until all necessary queries finished before displaying Set Targets step
2. improve error handling
    1. display error when targets could not be loaded
    2. silently ignore failed optional queries and continue without extra featues
3. extract data fetching into one hook

Resolves: https://github.com/konveyor/tackle2-ui/issues/2009
Resolves: https://issues.redhat.com/browse/MTA-3515


sjd78 (commented):
There are a few things that could be done a bit different, but the biggest issue is actually when the initial selection is made.  The targets and tag catagories may not have landed yet...and that makes the UI look strange for a few moments.

client/src/app/pages/applications/analysis-wizard/set-targets.tsx
```diff
@@ -48,22 +49,23 @@ export const SetTargets: React.FC<SetTargetsProps> = ({ applications }) => {
...
-        })
-        .filter((tagWithCat) => tagWithCat?.category?.name === "Language")
-        .map((tagWithCat) => tagWithCat.tag.name)
-    )
-  ).filter(Boolean);
+  const allProviders = targets.flatMap((target) => target.provider);
+  const languageOptions = Array.from(new Set(allProviders));
 
-  const [providers, setProviders] = useState(initialProviders);
+  const initialProviders = unique(
```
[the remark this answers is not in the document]
rszwajko: good catch @sjd78 ! we do need to handle error and loading state properly. However targets and tag categories are not equally important - failure to load the targets is a blocker while missing categories just lower user experience (no smart pre-filtering based on application tags)

rszwajko:
@sjd78 
Changes:
1. improved error handling (but based on wrapper that "waits" for data )
2. filtering instead of selecting targets - based on feedback from https://issues.redhat.com/browse/MTA-3515 

image

rszwajko:
@sjd78 @ibolton336 
In the last commit I've managed to use our almighty table hooks to power the card view. This brings us closer to PF4 [Card View](https://www.patternfly.org/patterns/card-view#card-view) pattern.
Missing parts:
1. re-enable `useSetting("ui.target.order")`
2. verify if we need the form to store the selection - the hook could handle that in similar way as in the table case

Fun fact ....it's not the first time when I'm moving from card to table view :) Check out https://github.com/oVirt/ovirt-web-ui/pull/1600

client/src/app/pages/applications/analysis-wizard/set-targets.tsx
```diff
@@ -168,25 +212,51 @@ export const SetTargets: React.FC<SetTargetsProps> = ({ applications }) => {
...
 };
+
+export const SetTargets: React.FC<SetTargetsProps> = ({ applications }) => {
+  // pre-fetch data but leave error handling to the real page
+  const { isLoading, applicationProviders } = useEnhancedTargets(applications);
+  if (isLoading) {
+    return (
+      <Bullseye>
+        <Spinner size="xl" />
+      </Bullseye>
```
sjd78, line 253: Ideally this would be one of those fancy skeleton previews (https://www.patternfly.org/components/skeleton)

sjd78 (changes requested):
Looking really good.  Mostly comments in the data hook.

client/src/app/pages/applications/analysis-wizard/set-targets.tsx
```diff
@@ -7,30 +7,89 @@ import {
...
+  // 1. missing target order setting is not a blocker (only lowers user experience)
+  // 2. targets without manual order are put at the end
+  const targetsWithOrder = targets.map((target) => {
+    const index = targetOrder.findIndex((id) => id === target.id);
+    return {
+      target,
+      order: index === -1 ? targets.length : index,
+    };
+  });
+  targetsWithOrder.sort((a, b) => a.order - b.order);
```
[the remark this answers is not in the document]
rszwajko: The fallback here is the default order - modern JS sort algorithms should be stable.

client/src/app/pages/applications/analysis-wizard/set-targets.tsx
rszwajko: IMHO the panel is too small - skeletons look the best on full page.

rszwajko:
### Optional queries failed during initial load

The step loads correctly but lacks the features that depend on the optional queries:
1. custom ordering
2. pre-filtering based on discovered language tags

https://github.com/user-attachments/assets/49671add-2386-410b-b06a-a4899fce903a

### Required query failed

Without the targets the step has no data to work on - however the user still may skip the step and continue to the next one.

https://github.com/user-attachments/assets/e4e8d77d-787d-494a-b6b3-bb7a5b410201

client/src/app/pages/applications/analysis-wizard/set-targets.tsx
sjd78: Good point!

client/src/app/pages/applications/analysis-wizard/set-targets.tsx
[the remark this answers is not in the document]
sjd78: Old habits ... never trust implicit sorting

rszwajko:
@JustinXHale 
please take a look at screenshots above - basically we are moving towards [standard Card View](https://www.patternfly.org/patterns/card-view/) which means that we can add more filters or other features (sorting, paging, actions). We can also add a  toggle that would allow users to switch to table or list view.

client/src/app/pages/applications/analysis-wizard/set-targets.tsx
```diff
@@ -169,24 +234,59 @@ export const SetTargets: React.FC<SetTargetsProps> = ({ applications }) => {
...
+
+  return (
+    <SetTargetsInternal
+      {...{
+        initialFilters: applicationProviders,
+        targets,
+        isError,
+        isLoading,
+        languageProviders,
+      }}
```
sjd78, line 289: I don't remember seeing this form before.  Interesting...

JustinXHale:
@sjd78 @rszwajko LGTM!

client/src/app/pages/applications/analysis-wizard/set-targets.tsx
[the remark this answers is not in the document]
rszwajko: We have 2 modes here: default (the same as in the data received from the server) or custom (manual ordering).
Changing the default order to some predefined is possible but I feel that the ultimate solution would be to add sorting widget and allow users to decide.
In this PR I would keep the current solution.

client/src/app/pages/applications/analysis-wizard/set-targets.tsx
rszwajko: Yes. The "internal" part is rendered once we have all the data - this allows to initialize the state with correct values.

Issue #2053, opened 2024-08-13
[BUG] Preemption column help needs re-wording.


jortel:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Konveyor version

0.5.1

### Priority

Minor

### Current Behavior

Preemption column (?) says a running task will be canceled.  It is not canceled, it is re-scheduled.

![Screenshot from 2024-08-13 08-11-48](https://github.com/user-attachments/assets/926d2af8-bbba-48d6-8d61-4544222b8254)

### Expected Behavior

Re-worded.

### How Reproducible

Always (Default)

### Environment

```markdown
- OS:
```

PR #2059, opened 2024-08-22
:bug: Change Preemption column tooltip to indicate that tasks are re-scheduled

rszwajko:
Resolves: #2053


rszwajko:
image

sjd78 (approved):
LGTM

@jortel - Does this version of the tooltip text look correct to you?

client/public/locales/en/translation.json
```diff
@@ -526,7 +526,7 @@
   },
   "tooltip": {
     "priority": "Tasks priority, a non-negative number, impacts tasks scheduling policy. Lowest priority is 0. Higher priority tasks run before lower priority tasks.",
-    "preemption": "If enabled, allows the scheduler to cancel a running task and free the resources for higher priority tasks."
+    "preemption": "If enabled, allows the scheduler to re-schedule a running task and free the resources for higher priority tasks."
```
jortel, line 529: recommend: 
"preemption": "If enabled, allows the scheduler to reschedule other running tasks to free cluster resources needed to run higher priority tasks."

rszwajko: @jortel 
> reschedule other running tasks
>
This part is not clear for me. Does it mean that preemption flag set on task A affects scheduling of all other tasks (but not directly task A) ?

jortel: Yes. exactly.

PR #2077, opened 2024-09-04
:bug:  Close dropdown after selection or click outside the list

rszwajko:
Fix applies to bulk selection dropdown and to filter categories dropdown on all table-based screens.

Resolves:  https://issues.redhat.com/browse/MTA-3321


sjd78 (approved):
LGTM

PR #2079, opened 2024-09-04
:bug: Support custom serializers in table hooks

rszwajko:
Changes:
1.  add persistence provider to feature level persistence (IFeaturePersistenceArgs) but not to table layer persistence (ITablePersistenceArgs). This simplifies the provider code (feature-specific providers). 
2. remove meta-persistence target "default" - no target has the same effect.
3. when in persistence provider mode use default values when the deserialized value is nulish. This solves the problem of initialFilterValues being overwritten in target cards scenario (the hook is called more then once and useState() based logic fails to detect initial load).

Using the newly added feature, store the filters selected in the Target step (Analysis wizard) inside react-hook form in the same way as other values. 

Resolves: https://issues.redhat.com/browse/MTA-3438


rszwajko:
https://github.com/user-attachments/assets/38186aed-6d31-41ce-b6b7-d19d4f239b53

sjd78 (approved):
LGTM

PR #2098, opened 2024-09-16
:sparkles: Add edit icon with ActionColumns property in the Archetypes table

Shevijacobson:
**Changes Made:**

- Removed the edit action from the ActionsColumn.

- Added a dedicated edit icon outside the ActionsColumn, providing a clearer and more focused user interface.

Part of #1318 

Before the change:

Screenshot from 2024-09-29 10-24-26

After the change:

Screenshot from 2024-09-29 10-28-03

### DIFF
--- a/client/src/app/pages/archetypes/archetypes-page.tsx
+++ b/client/src/app/pages/archetypes/archetypes-page.tsx
@@ -1,7 +1,6 @@
 import React, { useState } from "react";
 import { Trans, useTranslation } from "react-i18next";
 import { useHistory } from "react-router-dom";
-
 import {
   Button,
   ButtonVariant,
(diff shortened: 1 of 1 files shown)


mguetta1 (approved):
LGTM

rszwajko (commented):
Code looks good! 2 minor comments:
1. it would be nice to rephrase PR title to better reflect the change - it will be used as git commit title and stay in the repo history forever:)
2. if the change has some visual effect it's also nice to add before/after screenshots in a separate commit

rszwajko:
@mguetta1 
It seems we have a general problem when multiple cells with `isActionCell` property are displayed - in list view (small screen) they get stacked  (i.e. see below). Can you open an issue to track this? I think that with upgrade to newer PatterFly version we could solve it by using OverflowMenu as shown [here](https://www.patternfly.org/components/table#actions-overflow)
image

sjd78 (approved):
LGTM

mguetta1:
> @mguetta1 It seems we have a general problem when multiple cells with `isActionCell` property are displayed - in list view (small screen) they get stacked (i.e. see below). Can you open an issue to track this? I think that with upgrade to newer PatterFly version we could solve it by using OverflowMenu as shown [here](https://www.patternfly.org/components/table#actions-overflow) image

An issue has been opened to address the above: #2111

PR #2099, opened 2024-09-17
:sparkles: Add cancel option for multiple analyses

MiriSafra:
Added support for cancelling multiple analyses simultaneously. Integrated functionality into the actions menu for bulk applications.
Resolves: https://issues.redhat.com/browse/MTA-3639

**New Bulk Cancel Analyses Button**
Screenshot from 2024-10-30 11-20-18
**Cancel Analyses Button Disabled When No Applications Are Selected**
Screenshot from 2024-10-30 11-07-24
**Cancel Analyses Button Disabled When No Running Analyses Are Selected**
Screenshot from 2024-10-30 11-07-02
**Confirm Dialog Triggered on Cancel Analyses Button Click**
Screenshot from 2024-10-30 11-05-06
**Action Cancels All Running Analyses for Selected Applications**
Screenshot from 2024-10-30 11-06-27


client/src/app/pages/applications/applications-table/applications-table.tsx
```diff
@@ -575,6 +574,23 @@ export const ApplicationsTable: React.FC = () => {
         >
           {t("actions.delete")}
         </DropdownItem>,
+        ...(tasksReadAccess && tasksWriteAccess
+          ? [
+              <DropdownItem
+                key="applications-bulk-cancel"
+                isDisabled={
+                  !selectedRows.some((application: DecoratedApplication) =>
```
rszwajko, line 582: We allow here a mix of cancellable and non-cancellable tasks. I think this is the right choice because the status may change quickly before user will be able to invoke the action.
Note however that this is different then our usual way of  handling bulk actions where ALL items need to match the criteria.

client/src/app/pages/applications/applications-table/applications-table.tsx
```diff
@@ -575,6 +574,23 @@ export const ApplicationsTable: React.FC = () => {
...
+                    isTaskCancellable(application)
+                  )
+                }
+                onClick={() => {
+                  handleCancelBulkAnalysis();
```
rszwajko, line 587: We need a confirmation action here - similar to deleting apps - search for references to `applicationsToDelete`.

rszwajko (changes requested):
The action looks good! The missing part is confirmation dialog - user needs to confirm all destructive actions to prevent accidental loss of data.
Please also attach screenshots presenting the change.

client/src/app/pages/applications/applications-table/applications-table.tsx
rszwajko: @sjd78 ^^

rszwajko (approved):
LGTM! Thanks for contributing!
As a follow-up, could you create an issue to implement bulk canceling? Similar as it's done for deleting applications.

sjd78 (changes requested):
There is something off with the `isTaskCancellable` test.  When I select a single application that has no anlaysis task at all, the cancel action should be disabled. Right now it is not.

client/src/app/pages/applications/applications-table/applications-table.tsx
sjd78, line 1: Please update the `isTaskCancellable` function to use this one:

```jsx
  const isTaskCancellable = (application: DecoratedApplication) => {
    const task = application.tasks.currentAnalyzer;
    return !!task && !TaskStates.Terminal.includes(task?.state ?? "");
  };
```

The difference here is if no analysis task exists, there is nothing to cancel.  And technically `task?.state ?? ""` calculates to `""` and that is, in fact, not in the `TaskStates.Terminal` list.

MiriSafra (commented):
Updated `isTaskCancellable` with `!!task` check.

sjd78 (approved):
LGTM

PR #2100, opened 2024-09-18
:sparkles: adding filter by analysis to Application-inventory table

HadasahR:


client/src/app/pages/applications/applications-table/applications-table.tsx
```diff
@@ -499,7 +503,24 @@ export const ApplicationsTable: React.FC = () => {
...
+        title: t("terms.analysis"),
+        type: FilterType.multiselect,
+        placeholderText:
+          t("actions.filterBy", {
+            what: t("terms.analysis").toLowerCase(),
+          }) + "...",
+        selectOptions: Array.from(taskStateToAnalyze).map(
+          ([taskState, displayStatus]) => ({
+            value: taskState,
+            label: t(`${displayStatus}`),
```
rszwajko, line 517: The displayed values should match the values that the user sees in the table. Here the things get a bit tricky. 
Basically you need to repeat the same flow as done in the table by [ApplicationAnalysisStatus component](https://github.com/konveyor/tackle2-ui/blob/c04bb35390b3bd06e425dc8cead7cd92df2dd478/client/src/app/pages/applications/applications-table/applications-table.tsx#L929). There the received state is mapped via `taskStateToAnalyze`  BUT later on inside [IconedStatus](https://github.com/konveyor/tackle2-ui/blob/c04bb35390b3bd06e425dc8cead7cd92df2dd478/client/src/app/components/Icons/IconedStatus.tsx#L69)
it's mapped again - the `label` value is the value that the user sees.

client/src/app/pages/applications/applications-table/applications-table.tsx
```diff
@@ -499,7 +503,24 @@ export const ApplicationsTable: React.FC = () => {
...
+          })
+        ),
+        getItemValue: (item) => item?.tasks.currentAnalyzer?.state || "No Task",
```
rszwajko, line 520: The value here should match the `value` prop in the  `selectOptions` list. If this would not work please let me know - there are more options.

rszwajko (changes requested):
The code for new functionality is added in the correct places but there are some problems with mapping state values. Unfortunately adding code to legacy class of this size is never easy :)  

General comments:
1. wording: the PR title is OK (you could start with a capital letter though), the description is repeating the title so you could skip it. However leave a link to the issue/bug you are fixing i.e. Resolves: #1745 . In git history you will find some examples
2. check out your IDE configuration - it seems the IDE is not auto-formatting on save action. We have a script for linting - run from command line (in the project root) `npm run lint` . You will see the list of all violations (we somehow accumulated few of them). Then you can fix (only) yours. General cleanup is a candidate for another PR (`npm run lint -- --fix` will auto-fix all easy formatting problems in the whole project)

HadasahR:
Thanks, I'll check it out.

‫בתאריך יום ה׳, 26 בספט׳ 2024 ב-23:14 מאת ‪Radoslaw Szwajkowski‬‏ <‪
***@***.***‬‏>:‬

client/src/app/pages/applications/applications-table/applications-table.tsx
```diff
@@ -499,7 +503,38 @@ export const ApplicationsTable: React.FC = () => {
...
+      {
+        categoryKey: "analysis",
+        title: t("terms.analysis"),
+        type: FilterType.multiselect,
+        placeholderText:
+          t("actions.filterBy", {
+            what: t("terms.analysis").toLowerCase(),
+          }) + "...",
+
+        selectOptions: Object.values(applications)
```
rszwajko, line 516: `applications` is already an array - no need to use `Object.values`

client/src/app/pages/applications/applications-table/applications-table.tsx
```diff
@@ -499,7 +503,38 @@ export const ApplicationsTable: React.FC = () => {
...
+          .map((a) => {
```
rszwajko, line 517: You will need to re-use the same mapper for sorting so let's create a named function for that.
The steps remain the same i.e. (pseudo-code)
```
const mapAnalysisStateToLabel = (value: TaskState) => {
   // for consistency let's use the same wa of converting as the rest of the code
   const presetKey: IconedStatusPreset = getTaskStatus(value);
  // as for now the presets are created inside the  IconedStatus
  // we need refactor this code and extract this functionality to a function i.e.buildPresets() 
  // this is best to do under a different PR 
   const presets = buildPresets(t);
   // IMHO fallback to "Unknown" makes most sense
   const label = presets[presetKey]?.label ?? t("terms.unknown");
   return label;
}
```
Links:
 [getTaskStatus(value)](https://github.com/konveyor/tackle2-ui/blob/c04bb35390b3bd06e425dc8cead7cd92df2dd478/client/src/app/pages/applications/components/application-analysis-status.tsx#L24)

client/src/app/pages/applications/applications-table/applications-table.tsx
```diff
@@ -499,7 +503,38 @@ export const ApplicationsTable: React.FC = () => {
...
+
+            let label = taskStateToAnalyze.get(value as TaskState) || value;
+
+            if (label === "NotStarted") {
+              label = "Not started";
+            }
+            return { value, label };
+          })
+          .filter((v, i, a) => a.findIndex((v2) => v2.label === v.label) === i)
+          .sort((a, b) => a.value.localeCompare(b.value)),
```
rszwajko, line 532: we should sort on label - note also that `universalComparator` helper adds some extra power to `localeCompare` ("natural sorting")

client/src/app/pages/applications/applications-table/applications-table.tsx
```diff
@@ -345,6 +348,7 @@ export const ApplicationsTable: React.FC = () => {
       businessService: app.businessService?.name || "",
       tags: app.tags?.length || 0,
       effort: app.effort || 0,
+      analysis: app.tasks.currentAnalyzer?.state || "",
```
rszwajko, line 351: we should sort on labels visible on the screen -  see the propsed `mapAnalysisStateToLabel` below

rszwajko (changes requested):
The PR is going the right direction but it needs a preliminary refactoring - [extracting presets creation to a re-usable function ](https://github.com/konveyor/tackle2-ui/blob/c04bb35390b3bd06e425dc8cead7cd92df2dd478/client/src/app/components/Icons/IconedStatus.tsx#L65). I recommend to create a small PR for that and then re-use this functionality in this PR. Ping me if you need help with that.

client/src/app/pages/applications/applications-table/applications-table.tsx
TalyaNaima: the functions mentioned here are pushed in #2125

client/src/app/pages/applications/applications-table/applications-table.tsx
mguetta1: Done in https://github.com/konveyor/tackle2-ui/pull/2100/commits/a727e1a95fc07f5ff37d6121ed83f69f04526f64

client/src/app/pages/applications/applications-table/applications-table.tsx
mguetta1: Done in https://github.com/konveyor/tackle2-ui/pull/2100/commits/a727e1a95fc07f5ff37d6121ed83f69f04526f64

client/src/app/pages/applications/applications-table/applications-table.tsx
mguetta1: Done in https://github.com/konveyor/tackle2-ui/pull/2100/commits/a727e1a95fc07f5ff37d6121ed83f69f04526f64

client/src/app/pages/applications/applications-table/applications-table.tsx
mguetta1: Done in https://github.com/konveyor/tackle2-ui/pull/2100/commits/a727e1a95fc07f5ff37d6121ed83f69f04526f64

client/src/app/pages/applications/applications-table/applications-table.tsx
```diff
@@ -526,7 +537,28 @@ export const ApplicationsTable: React.FC = () => {
...
+          }) + "...",
+
+        selectOptions: applications
+          .map((a) => {
+            const value = a.tasks.currentAnalyzer?.state || "No task";
+            const label = mapAnalysisStateToLabel(value as TaskState, t);
+            return { value, label };
+          })
+          .filter((v, i, a) => a.findIndex((v2) => v2.label === v.label) === i)
+          .sort((a, b) => universalComparator(a.value, b.value)),
```
rszwajko, line 557: @mguetta1 
please sort on labels

rszwajko (commented):
Looks good! One minor comment (see below). It would be also nice to add a final screenshot of this feature (the existing ones are outdated)

HadasahR:
Thanks for the feedback!

‫בתאריך יום ה׳, 30 בינו׳ 2025 ב-13:21 מאת ‪Radoslaw Szwajkowski‬‏ <‪
***@***.***‬‏>:‬

client/src/app/pages/applications/applications-table/applications-table.tsx
mguetta1: Done. Thanks!

mguetta1:
https://github.com/user-attachments/assets/c2522031-18a5-48b4-b23e-ce29bdc5091a

rszwajko (approved):

Issue #1837, opened 2024-04-11
Enhance default analysis mode to be aware of application configurations


sjd78:
## Summary
When starting the analysis wizard, and selecting the analysis mode, a better default choice could be automatically made.  This can avoid the case where there is a validation message displayed by static default choice.  For example, we would like to avoid the analysis wizard opening looking like this:

<img width="1472" alt="Screenshot 2024-04-10 at 9 34 17 AM" src="https://github.com/konveyor/tackle2-ui/assets/11218376/dfbcf175-5349-4178-ba0a-a47d08d57e82">

## Change requested
If 1 application has been selected to be analyzed and the application is configured to have...
  - Binary definitions only &rarr; select Binary
  - Source definitions only &rarr; select Source + Dependencies
  - Source and Binary definitions &rarr; select Source + Dependencies
  - No definitions &rarr; No default selection

If >1 application has been selected to be analyzed, two options are available:
  - Default to a static Source + Dependencies
  - Follow the same rules as for a single application and if they all select the same option, select that one.  If any are different, default to a static Source + Dependencies

## Reference
The enhancement follows up on #1364 and https://github.com/konveyor/tackle2-ui/pull/1819#issuecomment-2047852789

dymurray:
Closing due to inactivity or age, please reopen if you feel this was done in error

sjd78:
Obsolete

PR #2106, opened 2024-09-22
 :sparkles: Fix Divergent Branch Issue and Update Default Mode Logic

yael-spinner:
This update introduces a function that changes the default state based on the type of application being deployed, ensuring optimal configuration for different scenarios while maintaining backward compatibility.
Related issue: #1837  
Related enhancement PR: #209


client/src/app/pages/applications/analysis-wizard/set-mode.tsx
```diff
@@ -71,7 +71,9 @@ export const SetMode: React.FC<ISetMode> = ({ isSingleApp, isModeValid }) => {
             toggleAriaLabel="Analysis mode dropdown toggle"
             aria-label={name}
             value={value}
-            onChange={onChange}
+            onChange={(value) => {
```
rszwajko, line 74: this change is only reformatting - I would keep the original format as it seems less verbose.

client/src/app/pages/applications/analysis-wizard/analysis-wizard.tsx
```diff
@@ -188,6 +218,19 @@ export const AnalysisWizard: React.FC<IAnalysisWizard> = ({
     mode: "all",
   });
 
+  useEffect(() => {
```
rszwajko, line 221: It seems to me that applications cannot change once the wizard was opened.
Can you describe the scenario that requires re-setting the defaults?

client/src/app/pages/applications/analysis-wizard/analysis-wizard.tsx
```diff
@@ -51,6 +51,37 @@ interface IAnalysisWizard {
...
+const determineMode = (
+  applications: Application[]
+):
+  | "binary"
+  | "source-code-deps"
+  | "source-code"
+  | "binary-upload"
+  | undefined => {
+  // If only one application is selected
+  if (applications.length === 1) {
```
rszwajko, line 63: Please provide some reference/description for the algorithm (issue on GH?)

rszwajko (commented):
Please check my questions below.

client/src/app/pages/applications/analysis-wizard/set-mode.tsx
yael-spinner: I returned to the previous state

client/src/app/pages/applications/analysis-wizard/analysis-wizard.tsx
yael-spinner: The bug occurred because the call to the function was performed in a click event, and in the case of a source type application, the value of the variable was not updated during the call to the function. As a result, only the binary files were transferred successfully.
To fix the bug, we added the useEffect function to ensure that the variable is updated automatically whenever the component is updated according to changes in the applications data. This way the useEffect ensures that the default value will be true and avoids a situation where an incorrect state is first displayed after the component is loaded.

client/src/app/pages/applications/analysis-wizard/analysis-wizard.tsx
yael-spinner: We added a function according to the conditions requested in the issue
And there are comments in the code explaining how the function works
Attached is a link to the GitHab issue https://github.com/konveyor/tackle2-ui/issues/1837

client/src/app/pages/applications/analysis-wizard/analysis-wizard.tsx
yael-spinner: issue #1837

client/src/app/pages/applications/analysis-wizard/analysis-wizard.tsx
```diff
@@ -51,6 +51,37 @@ interface IAnalysisWizard {
...
+    // If the application has only binary definitions, return "binary"
+    // If no definitions are present, return undefined
+    return repository || (repository && binary)
+      ? "source-code-deps"
+      : binary && binary !== ""
+      ? "binary"
+      : undefined;
+  }
+  // Check if all selected applications have only binary definitions
+  const allBinary = applications.every((app) => app.binary);
```
rszwajko, line 75: 1. `applications.every((app) => app.binary)` will work also for one element array -  it seems that you could delete the single-app case above and rely only on multiple-app case.
2. the code is self-documenting - please leave the comments only in places where variable names are not enough

client/src/app/pages/applications/analysis-wizard/analysis-wizard.tsx
```diff
@@ -163,10 +194,9 @@ export const AnalysisWizard: React.FC<IAnalysisWizard> = ({
   const methods = useForm<AnalysisWizardFormValues>({
     defaultValues: {
       artifact: null,
-      mode: "source-code-deps",
+      mode: determineMode(applications),
       formLabels: [],
       selectedTargets: [],
-      // defaults will be passed as initialFilterValues to the table hook
```
rszwajko, line 169: this comment is needed to explain why `targetFilters` are left undefined (instead of empty array)

client/src/app/pages/applications/analysis-wizard/set-mode.tsx
rszwajko: it seems the changes were not pushed yet

client/src/app/pages/applications/analysis-wizard/analysis-wizard.tsx
rszwajko: FYI I need to check this case.

rszwajko (commented):
Thanks for providing the context for the changes.  I have some more comments (see below). I need also verify how `useEffect` behaves in this place - will comment here when I'm done.

client/src/app/pages/applications/analysis-wizard/analysis-wizard.tsx
sjd78: This `useEffect()` does not make much sense.  It reads as "Use the default mode unless the default mode is falsey. If it is falsey, set a default of `source-code-deps`".

Really this could be done up in the `useForm` block.

client/src/app/pages/applications/analysis-wizard/analysis-wizard.tsx
```diff
@@ -163,10 +194,9 @@ export const AnalysisWizard: React.FC<IAnalysisWizard> = ({
   const methods = useForm<AnalysisWizardFormValues>({
     defaultValues: {
       artifact: null,
-      mode: "source-code-deps",
+      mode: determineMode(applications),
```
sjd78, line 197:
```suggestion
      mode: determineMode(applications) ?? "source-code-deps",
```
That way, if a mode can't be figured out from the selected apps, the fallback/default mode is the original default.

client/src/app/pages/applications/analysis-wizard/analysis-wizard.tsx
```diff
@@ -51,6 +51,37 @@ interface IAnalysisWizard {
...
+  if (applications.length === 1) {
+    const { repository, binary } = applications[0];
+    // If the application has a repository or both repository and binary definitions, return "source-code-deps"
+    // If the application has only binary definitions, return "binary"
+    // If no definitions are present, return undefined
+    return repository || (repository && binary)
+      ? "source-code-deps"
+      : binary && binary !== ""
+      ? "binary"
+      : undefined;
```
sjd78, line 72: This looks good.  Just put it in a function, iterate through the applications array and get the mode for each application.  At the end either all of the modes are the same, or they are different.  Same = return that mode, different = return undefined.

sjd78 (changes requested):
Looks on the right track, but it can be paired down a bit and still work properly.

client/src/app/pages/applications/analysis-wizard/analysis-wizard.tsx
rszwajko: @sjd78 unfortunately @yael-spinner  hit a real bug -  the `AnalysisWizard` gets re-rendered with every change on the applications screen which means that the defaults are being set at start time when no apps are selected.
So `useEffect` is necessary here as a workaround. The good news is that using (our old friend) `key` property we can force re-creating the wizard when selected apps change. The following fix worked for me:
```
iff --git a/client/src/app/pages/applications/applications-table/applications-table.tsx b/client/src/app/pages/applications/applications-table/applications-table.tsx
index d102d69a..d4affbb4 100644
--- a/client/src/app/pages/applications/applications-table/applications-table.tsx
+++ b/client/src/app/pages/applications/applications-table/applications-table.tsx
@@ -1060,6 +1060,7 @@ export const ApplicationsTable: React.FC = () => {
 
         <TaskGroupProvider>
           <AnalysisWizard
+            key={selectedRows.map(({ name }) => name).join()}
             applications={selectedRows}
             isOpen={isAnalyzeModalOpen}
             onClose={() => {
(END)

```

sjd78: Ah right!  I forgot about the order state gets loaded up in modals like this.  Using the `key` prop does make sense.

For some reference: https://react.dev/learn/preserving-and-resetting-state#resetting-a-form-with-a-key

So use the `key` attribute and drop `useEffect()`?

yael-spinner: @sjd78  @rszwajko I changed the code according to your comments

client/src/app/pages/applications/analysis-wizard/set-mode.tsx
yael-spinner: I uploaded the changes

client/src/app/pages/applications/analysis-wizard/analysis-wizard.tsx
yael-spinner: I fixed it

client/src/app/pages/applications/analysis-wizard/analysis-wizard.tsx
yael-spinner: I fixed it

client/src/app/pages/applications/analysis-wizard/analysis-wizard.tsx
yael-spinner: I returned to the previous state

yael-spinner (commented):
I updated the code according to the previous requests and comments,  If there are any questions, I'm available for clarification."

client/src/app/pages/applications/analysis-wizard/analysis-wizard.tsx
```diff
@@ -51,6 +51,28 @@ interface IAnalysisWizard {
...
+  if (applications.length === 0) return undefined;
+  const modes = applications.map((app) => {
+    const { repository, binary } = app;
+    return repository || (repository && binary)
+      ? "source-code-deps"
+      : binary && binary !== ""
+      ? "binary"
+      : undefined;
+  });
+  const firstMode = modes[0];
```
rszwajko, line 71: Note that this could be reduced to  problem of removing duplicates i.e. 
 ```
// using array -> set -> array
const modes =Array.from(new Set(application.map(....
// or using radash
const modes = unique(application.map(...
```

rszwajko (approved):
Looks good!

client/src/app/pages/applications/analysis-wizard/analysis-wizard.tsx
yael-spinner: I fixed it and also changed the title
is it ok now

Issue #1937, opened 2024-06-05
Add Tasks tab to the Applications table application details drawer


sjd78:
## Summary

With the addition of [automatic language discovery](https://github.com/konveyor/enhancements/issues/170), and [general task management](https://github.com/konveyor/tackle2-ui/issues/1931), the application table's application detail drawer needs to be enhanced to provide a view into the tasks associated with an application.  The details drawer will allow a user to list and view previous and current tasks performed on an application.

Enhancement references:
  - https://github.com/konveyor/enhancements/issues/171
  - https://github.com/konveyor/enhancements/issues/170

## Requirements

From the application table, when an application is selected, a details drawer opens.  The drawer shows multiple tabs of information about the application.  The existing "Reports" tab allows a user to view the most recent "analysis details" for the application.  This is the activity log of the analysis task.

Base requirements:
  - Add a new tab called "Tasks" to contain a simple table of the tasks associated to the application
  - Note: See the Dependencies details drawer for an example of a simple table within a tab in the drawer
  - The table should have the standard table functionality with filtering, sorting, and pagination
  - Due to space limitation, bulk actions / multiselect functionality can be omitted
  - The limited horizontal space will limit the number of columns that can be usefully shown.  Base candidates:
    - Task id
    - Task kind
    - Status (perhaps abbreviated)
    - Priority

Task actions:
  - Adopt the same set of individual task actions from #1931
  - Action on the tasks may be allowed based on task kind and if the current user is the creator of the task.  Put the set of action on a kebab menu for a task, with actions not currently permitted to be disabled with a tooltip explaining why.
  - Viewing a task's activity log should be possible by clicking a link on the task table, possibly the task's id or an action icon
  - With functionality of #1929, viewing the log will open as a full page view
  - Browser back button navigation, or breadcrumb navigation back to the application table with the selected application's detail drawer open to the Task tab should be supported

### Current State

View an application's most recent analysis task's activity log:
![screenshot-192 168 39 66-2024 06 04-22_32_25](https://github.com/konveyor/tackle2-ui/assets/3985964/260ecd81-81b1-420f-85c9-2f85970eba25)

### Future State

Show the task list:
_mockups not yet available_

Perform an action on a task:
_mockups not yet available_

View the tasks's activity log:
_mockups not yet available / may not be needed here_

sjd78:
@dymurray, @JustinXHale, @rromannissen, @jortel, @mansam -- for your consideration

PR #2123, opened 2024-10-08
:sparkles: Add Tasks tab to the Applications table application details drawer

EstyBiton:
Enhancing the application details drawer by adding a "Tasks" tab, displaying a table of tasks associated with the application, implementing filtering, sorting, and pagination functionalities, and including columns for Task ID, Task Kind, Status, and Priority. 

Fixes: #1937 

screenshot of result:
Screenshot from 2024-10-30 11-36-28


client/src/app/pages/tasks/TaskDetails.tsx
```diff
@@ -2,27 +2,54 @@ import React from "react";
...
 
 export const TaskDetails = () => {
   const { t } = useTranslation();
   const { taskId, attachmentId } = useParams<TaskDetailsAttachmentRoute>();
+  const { isFApplication } = useParams<TaskFromApp>();
+  const result = isFApplication === "true" ? true : false;
+  const [applicationName, setApplicationName] = useState<string | undefined>();
+  const [applicationId, setApplicationId] = useState<number | undefined>();
+
+  useEffect(() => {
```
sjd78, line 20: @rszwajko Can you have a look at these changes?  I don't remember exactly how these components hang together.

sjd78 (changes requested):
The work done is quite good.  There are some things that need to be changed.

1. On the side bar table, either drop the priority column or make the initial width of the drawer larger.  With the current width of the drawer, the action kebab are not visible.  The other option is to make the drawer something like 100px wider initially so the task action kebab is shown.

2. When linking to the task details, the URL is breaking.  You can follow the same format as `navigateToAnalysisDetails()`, but it should really be its own path.

3. The status column should use the same column style as the tasks-page (status icon, name, and link).  See: https://github.com/konveyor/tackle2-ui/blob/0cfe60b490e386a360a46e38769b56057845d377/client/src/app/pages/tasks/tasks-page.tsx#L244-L257

4. Need to fix the link that are being generated on task manager table (see specific comment).

I have a few more specific comment in the review.

EstyBiton:
@sjd78 Thank you for your review! I’ve made the following updates based on your feedback:
1.   Removed  priority field and unnecessary widths.
2.  Updated rendering: The table now renders like the task manager page, featuring a status icon with a link to view task details.
3. Changed to compact table: I added the variant="compact" prop to enhance the appearance of the table.

client/src/app/pages/tasks/TaskDetails.tsx
rszwajko: I missed this comment. I will take a look.

client/src/app/pages/tasks/TaskDetails.tsx
rszwajko, line 1: Note that after the changes the file duplicates the functionality of `AnalysisDetails` - in fact analysis details is just a task that is highlighted for user convenience. The code behind is/was generic - if you replace the IDs in the URL you should be able to view any task. Ideally you should be able to re-use this code after extracting to some "base" class - similar how both `TaskDetails` and `AnalysisDetails` use `TaskDetailsBase`. IMHO the refactoring (extracting base functionality) is small enough to do in this PR.

It could look the following
```
const  AnalysisDetails = () =>   
  <GoodNameForBaseComponent  
    formatDetailsPath={(applicationId,taskId) => formatPath(Paths.applicationsAnalysisDetails, {
        applicationId,
        taskId,
      });
   formatTitle={(taskName) => `Analysis details for ${taskName}`}
// whatever else is needed
  />  

const  ApplicationTaskDetails = () =>   
  <GoodNameForBaseComponent  
    formatDetailsPath={(applicationId,taskId) => ...);
   formatTitle={(taskName) => ...}
// whatever else is needed
  />  

```

client/src/app/pages/applications/components/application-detail-drawer/application-detail-drawer.tsx
```diff
@@ -521,3 +554,206 @@ const TabReportsContent: React.FC<{
...
+}> = ({ application, task }) => {
+  const { t } = useTranslation();
+  const history = useHistory();
+  const urlParams = new URLSearchParams(window.location.search);
+  const filters = urlParams.get("filters");
+  const deserializedFilterValues = deserializeFilterUrlParams({ filters });
+  const tableControlState = useTableControlState({
+    tableName: "tasks-apps-table",
+    persistTo: "urlParams",
+    persistenceKeyPrefix: TablePersistenceKeyPrefix.tasks,
```
rszwajko, line 570: it seems we have a collision with `TasksPage`

client/src/app/pages/applications/components/application-detail-drawer/application-detail-drawer.tsx
```diff
@@ -521,3 +554,206 @@ const TabReportsContent: React.FC<{
     </>
   );
 };
+
+const TabTasksContent: React.FC<{
```
rszwajko, line 558: Ideally we should be able to reuse the code from `TasksPage`. In the current state this requires a refactoring that would extract the "base" component. The best strategy would be to use a different PR for that and merge it first. This way we (and our test suite) ensure that the refactoring caused no regression.

After the refactoring both `TasksPage` and `TabTasksContent` should use the base and provide the custom parts. There are few differences:
1. identifiers i.e. `tableName`, `persistenceKeyPrefix`
2. visible columns - note that hiding (via `initialColumns`)  looks easier then maintaining 2 lists of columns
3. query params - `implicitFilters`
4.  possibly more

If "base" component variant would be not achievable we can break the TasksPage into smaller parts and try to reuse those.

rszwajko (commented):
The PR looks pretty impressive! It basically adds the entire feature. We could merge it now however too keep the code [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself) we need a follow-up refactorings. 
It would be better to merge the main refactoring (extracting base table component) as a prerequisite (different PR), merge it and then use it here. Please let me know if you can work on this (details in the comments). If not we can consider other variants.

client/src/app/pages/applications/components/application-detail-drawer/application-detail-drawer.tsx
EstyBiton: @rszwajko Thank you for the feedback! I would really like to continue working on this task, but I currently have a challenge. My internship with the company has concluded, and I no longer have access to the computer where the product environment was properly set up. Unfortunately, running Konveyor on Minikube in the Windows operating system presents a significant issue: the local Konveyor UI is inaccessible.
 If the issue with installing Konveyor is resolved, I’d be happy to continue working on this task. Please let me know.

sshveta:
@rszwajko @sjd78 Can we please review/merge these PRs .

sjd78 (approved):
Looks ok to me

rszwajko (approved):
Functionality works as expected according to my smoke tests. The recommended refactorings can be applied in follow-ups.

PR #2152, opened 2024-11-03
:sparkles:Build presets & map analysis state functions

HadasahR:
refactoring code according to comments in issue #2100
these functions are used in the PR there

### DIFF
--- a/client/src/app/pages/applications/components/application-analysis-status.tsx
+++ b/client/src/app/pages/applications/components/application-analysis-status.tsx
@@ -1,7 +1,11 @@
 import React from "react";
 
 import { TaskState } from "@app/api/models";
-import { IconedStatus, IconedStatusPreset } from "@app/components/Icons";
+import {
+  buildPresetLabels,
+  IconedStatus,
+  IconedStatusPreset,
+} from "@app/components/Icons";
 export interface ApplicationAnalysisStatusProps {
   state: TaskState;
 }
@@ -43,22 +47,3 @@ export const mapAnalysisStateToLabel = (
   const label = presets[presetKey]?.label ?? presets.Unknown.label;
   return label;
 };
-
-export const buildPresetLabels = (
-  t: (key: string) => string
-): Record<IconedStatusPreset, { label: string }> => ({
-  InProgressInheritedReviews: { label: t("terms.inProgress") },
-  InProgressInheritedAssessments: { label: t("terms.inProgress") },
-  InheritedReviews: { label: t("terms.completed") },
-  InheritedAssessments: { label: t("terms.completed") },
-  Canceled: { label: t("terms.canceled") },
-  Completed: { label: t("terms.completed") },
-  CompletedWithErrors: { label: t("terms.completedWithErrors") },
-  Error: { label: t("terms.error") },
-  Failed: { label: t("terms.failed") },
-  InProgress: { label: t("terms.inProgress") },
-  NotStarted: { label: t("terms.notStarted") },
-  Scheduled: { label: t("terms.scheduled") },
-  Ok: { label: t("terms.ok") }, // Add Ok with a label
-  Unknown: { label: t("terms.unknown") }, // Add Unknown with a label
-});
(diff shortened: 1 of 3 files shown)


client/src/app/pages/applications/components/application-analysis-status.tsx
```diff
@@ -34,3 +34,30 @@ export const ApplicationAnalysisStatus: React.FC<
 > = ({ state }) => {
   return <IconedStatus preset={getTaskStatus(state)} />;
 };
+export const mapAnalysisStateToLabel = (value: TaskState) => {
+  const presetKey: IconedStatusPreset = getTaskStatus(value);
+  const { t } = useTranslation();
```
rszwajko, line 39: Unfortunately we cannot use the translation hook here as it's an ordinary method -  check  [Rules of Hooks](https://react.dev/warnings/invalid-hook-call-warning).
We could pass the `t` function as a param or build the presets outside.

client/src/app/pages/applications/components/application-analysis-status.tsx
```diff
@@ -34,3 +34,30 @@ export const ApplicationAnalysisStatus: React.FC<
...
+  const presets = buildPresets(t);
+  const label = presets[presetKey]?.label ?? t("terms.unknown");
+  return label;
+};
+export const buildPresets = (
```
rszwajko, line 44: 1. The purpose of the refactoring is to improve the code but keep the functionality the same.Here you change some of the labels - compared with IconedStatus component.
2. After we extracted the code to a separate reusable function we need to use it in the original place i.e.

```
 export const IconedStatus: React.FC<IIconedStatusProps> = ({
   preset,
   status,
@@ -62,11 +83,12 @@ export const IconedStatus: React.FC<IIconedStatusProps> = ({
   tooltipCount = 0,
 }: IIconedStatusProps) => {
+  const messages = buildPresetsLabels(t);
   const presets: IconedStatusPresetType = {
     InProgressInheritedReviews: {
       icon: <InProgressIcon />,
       status: "info",
-      label: t("terms.inProgress"),
+      label: messages.InProgressInheritedReviews.label,
       tooltipMessage: t("message.inheritedReviewTooltip", {
         count: tooltipCount,
       }),
```

rszwajko (changes requested):
The commit with refactoring needs to be moved to separate PR. One way to do it is as follows:
```
// do a backup - rename a branch to "..._v1" and push to GH
// NOTE that I've changed char "&" to word "and" 
// "&" is an operator in bash shell and requires escaping/quoting 
 » git branch -M buildPresets_and_mapAnalysisState_functions_v1  
 » git push origin HEAD
// obtain commit hash
 » git log -1 | cat                                                                                                                    130 ↵
commit c070fa6827ff42de08cf510c792428beed7c3a81
Author: HadasahR <d025834828@gmail.com>
Date:   Sun Nov 3 14:05:13 2024 +0200

    refactoring code as suggested in #2100
    
    Signed-off-by: HadasahR <d025834828@gmail.com>
// create a new branch "..._v2"
 » git checkout -b buildPresets_and_mapAnalysisState_functions_v2 origin/main
// copy the commit to new branch
» git cherry-pick c070fa6827ff42de08cf510c792428beed7c3a81
// force push to the PR branch 
» git push origin HEAD:"buildPresets_&_mapAnalysisState_functions" --force
```

client/src/app/pages/applications/components/application-analysis-status.tsx
mguetta1: Hi @rszwajko
Looks like `buildPresets` is not related only to analysis status but to any status (assessment, review, etc.) where do you suggest to move this functionality please?

client/src/app/pages/applications/components/application-analysis-status.tsx
mguetta1: Done in https://github.com/konveyor/tackle2-ui/pull/2152/commits/d2096ca78fa99877f5a7eed32db9402a4e3c83eb

client/src/app/pages/applications/components/application-analysis-status.tsx
rszwajko: hmm...right. Probably better to put there in the same place as the original code - in the `client/src/app/components/Icons/IconedStatus.tsx`

PR #2164, opened 2024-12-09
:bug: Replace EmptyState with NoDataEmptyState component for consistency

mguetta1:
Resolves: https://issues.redhat.com/browse/MTA-4336

Before:
task-manager-before

After:
task-manager-after

### DIFF
--- a/client/src/app/pages/tasks/tasks-page.tsx
+++ b/client/src/app/pages/tasks/tasks-page.tsx
@@ -2,9 +2,6 @@ import React, { ReactNode } from "react";
 import { useTranslation } from "react-i18next";
 import { Link, useHistory } from "react-router-dom";
 import {
-  EmptyState,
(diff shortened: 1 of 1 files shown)


rszwajko (commented):
On this screen we do not distinguish between no-data-available and no-results-found(no data due to applied filters).  
The current message is inaccurate when there is no data at all. The proposed message goes in the opposite direction. 
The best solution would be to separate both use cases. This however would require more knowledge about applied filters - probably based on `getFilterHubRequestParams()`. 

PatternFly guidelines for reference:
1. no results found (due to filtering) - https://www.patternfly.org/components/empty-state/design-guidelines#no-results
2. no data (yet) - https://www.patternfly.org/components/empty-state/design-guidelines#level-of-detail

sjd78 (approved):
Retaining the original title makes this PR ok.

PR #2165, opened 2024-12-11
✨ Fix Divergent Branch Issue and Update Default Mode Logic

### LINKED ISSUE #1837: Enhance default analysis mode to be aware of application configurations

shirael:
This update introduces a function that changes the default state based on the type of application being deployed, ensuring optimal configuration for different scenarios while maintaining backward compatibility.
Resolves: https://github.com/konveyor/tackle2-ui/issues/1837

### DIFF
--- a/client/src/app/pages/applications/analysis-wizard/analysis-wizard.tsx
+++ b/client/src/app/pages/applications/analysis-wizard/analysis-wizard.tsx
@@ -51,6 +51,16 @@ interface IAnalysisWizard {
   isOpen: boolean;
 }
 
+const determineMode = (
+  applications: Application[]
+):
+  | "binary"
+  | "source-code-deps"
+  | "source-code"
+  | "binary-upload"
+  | undefined => {
+  if (applications.length === 0) return undefined;
+};
 const defaultTaskData: TaskData = {
   tagger: {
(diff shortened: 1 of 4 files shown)


sjd78:
PR is obsolete.  It has been fully superseded by #2954 and #2979.

PR #2178, opened 2025-01-31
:seedling: Update ci-global.yml use ttl.sh and global-ci-bundle.yml

sjd78:
With `upload-artifact@v3` deprecated and shut off, move to use the `global-ci-bundle.yml` workflow for global CI.  This
change drops the need for artifacts and just uses the ttl.sh service.  This technique is already being used in a few other
Konveyor projects as well.

Ref: https://github.blog/changelog/2024-04-16-deprecation-notice-v3-of-the-artifact-actions/
Ref: https://ttl.sh

Requires: https://github.com/konveyor/ci/pull/76

### DIFF
--- a/.github/workflows/ci-global.yml
+++ b/.github/workflows/ci-global.yml
@@ -39,24 +39,24 @@ concurrency:
 jobs:
   build-and-upload-for-global-ci:
     runs-on: ubuntu-latest
+    outputs:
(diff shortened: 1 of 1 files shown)


sjd78:
Note: I did something really weird force pushing or amending commit or something.  Recreating the PR clean.

PR #2197, opened 2025-03-13
:bug: Update the bulk selector to only consider the current page

sjd78:
Resolves: https://issues.redhat.com/browse/MTA-4685

For tables with row selection, update the bulk selection component so that it works consistently across client local and server managed filter/sort/pagination.  With this change, the bulk selector will allow:
  - Allow clearing all selections
  - Allow selecting all row in the current page
  - Show the count of selected rows

Client local tables could support additional selections such as "select all filtered rows", and "select all rows".  At this time, maintaining consistency across tables is better than segmenting bulk selection functionality depending on table data source type. Only select none and select current page is supported.

Summary of changes:
  - Split expansion logic out of `ToolbarBulkSelector` to `ToolbarBulkExpander` (do one thing per component)
  - Remove unused `isItemSelectable` prop from the `useSelectionState()` hook
  - Move data functionality out of `ToolbarBulkSelector` into `useBulkSelector()` to keep data handling separate from the render
  - Refactor to simplify `useSelectionState()` and move selection tracking out of `useTableContolProps()`


sjd78:
Screen shot with 3 items selected and the split button open:
image

client/src/app/components/ToolbarBulkSelector.tsx
```diff
@@ -1,152 +1,139 @@
...
 
-export const ToolbarBulkSelector = <T,>({
-  currentPageItems,
+/**
+ * Selection options:
+ *   - Default Button / Select Page
+ *   - Split / None, Count: 0
+ *   - Split / Page (every record on the current page), Count: records in current page
+ *   - Split / All Filtered (every record matching the filter), Count: filtered records
+ *   - Split / All (every record, every page), Count: items size
```
rszwajko, line 32: we might not be able to deliver this feature in every case i.e. on screens with server-side paging or with incomplete/stale data. We should be able to toggle the feature for screens that we know that will work.

client/src/app/hooks/table-controls/useLocalTableControls.ts
```diff
@@ -34,16 +34,17 @@ export const useLocalTableControls = <
...
+    isEqual: (a, b) => a[args.idProperty] === b[args.idProperty],
+  });
+
   return useTableControlProps({
     ...args,
     ...state,
     ...derivedState,
-    // TODO we won't need this here once selection state is part of useTableControlState
-    selectionState: useSelectionState({
-      ...args,
```
rszwajko, line 43: IMHO we have a bug here  - `items` prop which is hidden inside `args` contains all items. I think it should work similar to screens with server side paging (only the current page with all filters applied).
Even better if this could be configured (to allow select across all pages) but still the default should be safe.

client/src/app/hooks/table-controls/useLocalTableControls.ts
```diff
@@ -34,16 +34,17 @@ export const useLocalTableControls = <
...
+  const selectionState = useSelectionState({
+    ...args,
+    isEqual: (a, b) => a[args.idProperty] === b[args.idProperty],
+  });
+
   return useTableControlProps({
     ...args,
     ...state,
     ...derivedState,
-    // TODO we won't need this here once selection state is part of useTableControlState
```
rszwajko, line 41: I know those comments are a bit outdated but I would keep them because the area is complex

client/src/app/hooks/table-controls/useSelectionState.ts
```diff
@@ -0,0 +1,92 @@
+import { useState, useCallback, useMemo } from "react";
+
+export interface ISelectionStateArgs<T> {
```
rszwajko, line 3: Please go first with functional changes (and keep them minimal) -  after moving and re-formatting it's hard to review what's changed i.e.  `isItemSelectable` is gone. It was not very useful in our app but still this change should be easy to notice and acknowledge

client/src/app/components/ToolbarBulkSelector.tsx
sjd78: Yes, the "all filtered" and the "all" actions are optional and are only included if the handler is provided.

client/src/app/hooks/table-controls/useSelectionState.ts
sjd78: Two things were pulled out `isItemSelectable` (as mentioned in the PR description) and the `externalState`.  Neither were used and the dead code was removed.

The rest of the changes swapped the change functions out to better match up to the needs of UI components.

client/src/app/hooks/table-controls/useLocalTableControls.ts
sjd78: I'm not quite sure what you mean with this comment.  I'll explicitly call out each prop going to the hook.  But this isn't a change to how it was working before.

sjd78: BTW -- trying to figure out if the items or a filtered set of items should be used is strangely difficult.  The full item list is good enough for now since the selecting functions in the component helper props take care of which set of records to select for different callbacks.

client/src/app/hooks/table-controls/useSelectionState.ts
sjd78: And yeah -- git decided that I changed the file enough that it wasn't a rename and update, it was a remove and this is a new file.  I really don't want to just rearrange the code to make the diff smaller and just be a rename plus change.

client/src/app/hooks/table-controls/useLocalTableControls.ts
sjd78: Ok I understand what you're talking about now.  If the items in the selectionState were exactly the max items that are safe to select, then the core bug could be solved. 

I'd like to just leave it as the full set of items available (for client tables == all items, for server tables == current page) and have the table props hook setup the various selection handlers as they make sense given a wider view of the data.

client/src/app/hooks/table-controls/useSelectionState.ts
sjd78: Rearranged things, diffs should look better now.  New comments in a new thread.

client/src/app/hooks/table-controls/useLocalTableControls.ts
sjd78: Specifically, the selection vs item work is done in `useTableControlProps()`...

```tsx
  const toolbarBulkSelectorProps: PropHelpers["toolbarBulkSelectorProps"] = {
    areAllSelected,
    itemCounts: {
      selected: selectedItems.length,
      page: currentPageItems.length,
      filtered: totalItemCount, // TODO: Adjust when enable onSelectAllFiltered
      items: totalItemCount, // TODO: Adjust when enable onSelectAll
    },
    onSelectNone: () => selectAll(false),
    onSelectCurrentPage: () => selectOnly(currentPageItems),
    // TODO: send these functions only if we know local filtering is active. They don't make
    //       sense in server side pagination/filter/sort.
    // onSelectAllFiltered: () => selectOnly(filteredItems),
    // onSelectAll: () => selectAll(true),
  };
```

sjd78:
> Looks good but the changes in the `useSelectionState` hook need to be traceable. I would also re-consider the default selection mode in `useLocalTableControls()` - it's safer to be the current page (with all items being an extra option)

I undid the move of `useSelectionState()` hook so the diffs look reasonable now.

client/src/app/hooks/useSelectionState/useSelectionState.ts
```diff
@@ -4,99 +4,89 @@ export interface ISelectionStateArgs<T> {
...
-    [selectableItems, setSelectedItems]
+    (isSelecting) => setSelectedItems(isSelecting ? items : []),
+    [items, setSelectedItems]
   );
-  const areAllSelected = selectedItems.length === selectableItems.length;
+  const areAllSelected = selectedItems.length === items.length;
 
-  // Preserve original order of items
-  const selectedItemsInOrder = React.useMemo(() => {
+  const selectedItemsInItemsOrder = React.useMemo(() => {
```
rszwajko, line 73: WDYT about combining this step with setting the selection state? i.e. 
This would ensure that we have one consistent selection state.
```
setSelectedInItemsOrder= (toBeSelected) => setSelectedItems(
    items.filter(it => toBeSelected.some(selected => ...)
)
```

client/src/app/hooks/table-controls/useTableControlProps.ts
```diff
@@ -113,14 +111,27 @@ export const useTableControlProps = <
...
+    onSelectNone: () => selectAll(false),
+    onSelectCurrentPage: () => selectOnly(currentPageItems),
+    // TODO: send these functions only if we know local filtering is active. They don't make
+    //       sense in server side pagination/filter/sort.
+    // onSelectAllFiltered: () => selectOnly(filteredItems),
+    // onSelectAll: () => selectAll(true),
+  };
+
+  const toolbarBulkExpanderProps: PropHelpers["toolbarBulkExpanderProps"] = {
+    // areAllExpanded,
```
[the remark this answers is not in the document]
sjd78: They never existed to begin with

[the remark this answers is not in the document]
sjd78: i.e. bulk expansion never worked

client/src/app/hooks/useSelectionState/useSelectionState.ts
```diff
@@ -4,99 +4,89 @@ export interface ISelectionStateArgs<T> {
...
 
   const selectAll = React.useCallback(
-    (isSelecting = true) =>
-      setSelectedItems(isSelecting ? selectableItems : []),
-    [selectableItems, setSelectedItems]
+    (isSelecting) => setSelectedItems(isSelecting ? items : []),
+    [items, setSelectedItems]
   );
-  const areAllSelected = selectedItems.length === selectableItems.length;
+  const areAllSelected = selectedItems.length === items.length;
```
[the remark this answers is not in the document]
sjd78: I noticed issues with this also.  I updated this to actually cross-check the check arrays.

client/src/app/pages/applications/applications-table/applications-table.tsx
```diff
@@ -577,15 +579,22 @@ export const ApplicationsTable: React.FC = () => {
...
-
-    selectionState: { selectedItems: selectedRows },
     columnState,
   } = tableControls;
 
+  const {
+    selectedItems: selectedRows,
+    propHelpers: { toolbarBulkSelectorProps, getSelectCheckboxTdProps },
+  } = useBulkSelection({
+    items: applications,
```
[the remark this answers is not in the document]
rszwajko: current page

client/src/app/pages/migration-waves/components/manage-applications-form.tsx
```diff
@@ -194,6 +193,17 @@ export const ManageApplicationsForm: React.FC<ManageApplicationsFormProps> = ({
     expansionDerivedState: { isCellExpanded },
   } = tableControls;
 
+  const {
+    selectedItems,
+    propHelpers: { toolbarBulkSelectorProps, getSelectCheckboxTdProps },
+  } = useBulkSelection({
+    items: availableApplications,
```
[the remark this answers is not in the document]
rszwajko: current page

client/src/app/pages/migration-waves/migration-waves.tsx
```diff
@@ -253,6 +255,16 @@ export const MigrationWaves: React.FC = () => {
     expansionDerivedState: { isCellExpanded, setCellExpanded },
   } = tableControls;
 
+  const {
+    selectedItems,
+    propHelpers: { toolbarBulkSelectorProps, getSelectCheckboxTdProps },
+  } = useBulkSelection({
+    items: migrationWaves,
```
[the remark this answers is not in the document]
rszwajko: current page

rszwajko (commented):
Looks good! A lot of `fixme` and `todo` removed !
Regarding the selection scope:  I would vote for current page items everywhere.
This seems the safest option - user sees the items that will be used in a given action.
The disadvantage is that filtering/sorting/paging may clear selection.  However to safely allow selecting items not visible on the screen we would need re-think our actions i.e. they could display a warning if items not visible on the screen will be impacted.

client/src/app/hooks/table-controls/useLocalTableControls.ts
```diff
@@ -24,26 +23,24 @@ export const useLocalTableControls = <
     TFilterCategoryKey,
     TPersistenceKeyPrefix
   >
-): ITableControls<
-  TItem,
-  TColumnKey,
-  TSortableColumnKey,
-  TFilterCategoryKey,
-  TPersistenceKeyPrefix
+): ReturnType<
```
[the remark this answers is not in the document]
sjd78: No, but I like the extra info.

client/src/app/pages/applications/applications-table/applications-table.tsx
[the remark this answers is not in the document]
sjd78: For sure that's the safest way to go.

rszwajko (approved):
Looks good! The effect of restoring the selection after going back (paging, filtering) works similar to popular mail clients.

Issue #2198, opened 2025-03-13
On the analysis wizard, the targets field shows as empty if only a repo custom target is selected


sjd78:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Konveyor version

0.6

### Priority

Major

### Current Behavior

- Make sure a custom target that is based on a repository is available.
- Step through the analysis wizard.
- On the Set Targets step, select the repository Custom Target
- Step forward to the Advanced, Options and then Review step...
- Notice no targets are listed

### Expected Behavior

There should be an indication that a target has been selected if it does not have an associated set of labels.

### How Reproducible

Always (Default)

### Environment

```markdown
- OS:
```

sjd78:
Some background information: https://issues.redhat.com/browse/MTA-3438?focusedId=26776825&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-26776825

sjd78:
See related tickets:
 - https://issues.redhat.com/browse/MTA-4475
 - https://issues.redhat.com/browse/MTA-4795
 - https://issues.redhat.com/browse/MTA-4585

PR #2200, opened 2025-03-13
:seedling: Refactor wizard form field name to improve readability

sjd78:
The wizard form field `formLabels` is confusing and misleading. The field has been renamed `selectedTargetLabels`.  Variables have been renamed as appropriate.

Util function `getUpdatedFormLabels()` changed to `updateSelectedTargetLabels()` and refactored to streamline its functionality.

Pre-work for #2198 
Needs #2325 for unit tests

### DIFF
--- a/client/public/locales/en/translation.json
+++ b/client/public/locales/en/translation.json
@@ -605,9 +605,9 @@
       "scope": "Scope",
       "setTargets": "Set targets",
       "source": "Source",
-      "sources": "Sources",
+      "source_plural": "Sources",
       "target": "Target",
-      "targets": "Targets",
+      "target_plural": "Targets",
       "transactionReport": "Transaction Report",
       "autoTagging": "Automated tagging"
     },
--- a/client/src/app/pages/applications/analysis-wizard/utils.ts
+++ b/client/src/app/pages/applications/analysis-wizard/utils.ts
@@ -73,34 +73,31 @@ export const toggleSelectedTargets = (
   return toggle(selectedTargets, target, (t) => t.id);
 };
 
-export const getUpdatedFormLabels = (
+export const updateSelectedTargetLabels = (
   isSelecting: boolean,
   selectedLabelName: string,
   target: Target,
-  formLabels: TargetLabel[]
+  currentSelectedTargetLabels: TargetLabel[]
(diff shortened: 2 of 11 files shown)


client/public/locales/en/translation.json
```diff
@@ -607,9 +607,9 @@
       "scope": "Scope",
       "setTargets": "Set targets",
       "source": "Source",
-      "sources": "Sources",
+      "source_plural": "Sources",
       "target": "Target",
-      "targets": "Targets",
+      "target_plural": "Targets",
```
rszwajko, line 612: interesting!

client/src/app/pages/applications/analysis-wizard/utils.ts
```diff
@@ -73,34 +73,31 @@ export const toggleSelectedTargets = (
   return toggle(selectedTargets, target, (t) => t.id);
 };
 
-export const getUpdatedFormLabels = (
+export const updateSelectedTargetLabels = (
```
rszwajko, line 76: the refactoring looks good but to be on the safe side I would create some unit tests before this PR as safety net.  This way we could verify that that after refactoring the flow is still the same.

client/src/app/utils/rules-utils.ts
```diff
@@ -5,11 +5,13 @@ type RuleFileType = "YAML" | "XML" | null;
 
 export const checkRuleFileType = (filename: string): RuleFileType => {
   const fileExtension = filename.split(".").pop()?.toLowerCase();
-  if (fileExtension === ("yaml" || "yml")) {
```
rszwajko, line 8: good catch!

client/public/locales/en/translation.json
sjd78: Yep, just need to include `{ count }` in the intl call to have it be able to make the choice between plural forms.

client/src/app/pages/applications/analysis-wizard/utils.ts
sjd78: See #2325!

rszwajko (approved):
Looks good! there are some minor merge conflict after the PR with tests was merged

PR #2203, opened 2025-03-20
:bug: Add ID attribute to archetype kebab menu

nachandr:


client/src/app/pages/archetypes/archetypes-page.tsx
```diff
@@ -484,7 +484,7 @@ const Archetypes: React.FC = () => {
                             </Tooltip>
                           </Td>
                         )}
-                        <Td isActionCell>
+                        <Td isActionCell id="row-actions">
```
rszwajko, line 487: The `id` attribute should be unique (within the document which in our case is the whole page).
If we need to use it in tests please consider the alternatives i.e. custom attributes (`data-*`).

sjd78: In this case, it should be fine.  The testing could look for the correct `Tr` first (with `data-item-id` or `data-item-name` attributes and then find its `Td[id="row-actions"]`.

sjd78 (approved):

Issue #2176, opened 2025-01-29
Allow the user to add custom targets for supported providers other than Java and Go


kpunwatk:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Konveyor version

main

### Priority

Major

### Current Behavior

Right now we don't allow the user to add custom targets for anything outside of java and Go, for example python or nodejs.

![Image](https://github.com/user-attachments/assets/b06fa99f-102d-4373-9282-c1bd711b477d)

### How Reproducible

Always (Default)

### Environment

```markdown
- OS:
```

sjd78:
For adding custom targets, the provider dropdown can match to one of the tags in the "Language" tag category.

That way, two nice things happen.  First, if a new Language is needed, it can be added by creating a new Tag.  Second, the Target provider name can now match up with the application language tags to maintain consistency.

sjd78:
See enhancement https://github.com/konveyor/enhancements/issues/218

PR #2230, opened 2025-04-17
:sparkles: Update custom migration targets filtering/creation

sjd78:
Resolves: #2176

On the custom migrations target page, update the page to better support provider types. Summary of changes:
  - Add support for target filtering by Name and Provider
  - The filter provider list is pulled from the existing targets' providers so all targets are all always visible to filtering
  - The provider type is now selected on the create modal
  - The provider type can be edited for any custom target
  - The list of providers is contained within the hook `useMigrationProviderList()`.  This allows the list to be changed with little disruption to other components

Other related changes:
  - Resolved linting warnings
  - Fixed types for Target queries

### DIFF
--- a/client/src/app/pages/migration-targets/components/custom-target-form.tsx
+++ b/client/src/app/pages/migration-targets/components/custom-target-form.tsx
@@ -1,4 +1,4 @@
-import React, { useContext, useEffect, useState } from "react";
+import React, { useContext, useEffect, useState, useMemo } from "react";
 import {
   ActionGroup,
   Alert,
@@ -28,14 +28,7 @@ import {
 } from "@app/components/HookFormPFFields";
 import { getAxiosErrorMessage } from "@app/utils/utils";
 import { useCreateFileMutation } from "@app/queries/targets";
-import {
-  IReadFile,
-  New,
-  ProviderType,
-  Rule,
-  Target,
-  TargetLabel,
-} from "@app/api/models";
+import { IReadFile, New, Rule, Target, TargetLabel } from "@app/api/models";
 import { getParsedLabel, parseRules } from "@app/utils/rules-utils";
 import { OptionWithValue, SimpleSelect } from "@app/components/SimpleSelect";
 import { toOptionLike } from "@app/utils/model-utils";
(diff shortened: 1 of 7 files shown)


sjd78:
Old view:
image

New view:
image

New create target:
image

client/src/app/pages/migration-targets/migration-targets.tsx
```diff
@@ -124,22 +134,39 @@ export const MigrationTargets: React.FC = () => {
...
         ]);
       }
-    }
 
+      // Make sure the new target's provider is part of the provider's filter so it can be seen
+      if (filterState.filterValues["provider"]) {
+        let newProviderFilter: FilterValue = null;
+
+        const targetProvider = response.data.provider;
+        if (targetProvider) {
```
rszwajko, line 143: this part looks useful but tricky
1. if there is no `targetProvider` (never?) we should just return
2. `newProviderFilter` may be const

client/src/app/pages/migration-targets/migration-targets.tsx
```diff
@@ -124,22 +134,39 @@ export const MigrationTargets: React.FC = () => {
...
+          const fv = filterState.filterValues["provider"];
+          newProviderFilter = fv.includes(targetProvider)
+            ? fv
+            : [...fv, targetProvider];
+        }
+
+        filterState.setFilterValues({
+          name: filterState.filterValues["name"],
```
rszwajko, line 151: can we just unpack all the filters to be future proof ? i.e. `{... filterState.filterValues, provider: newProviderFilter}`

rszwajko (commented):
Looks good!

client/src/app/pages/migration-targets/migration-targets.tsx
sjd78: If there is no provider (which is permissible on hub), then the only way to see it is to remove the filter by setting the filterState to null.

client/src/app/pages/migration-targets/components/custom-target-form.tsx
```diff
@@ -75,9 +73,23 @@ export const CustomTargetForm: React.FC<CustomTargetFormProps> = ({
   target: initialTarget,
   onSaved,
   onCancel,
-  providerType,
 }) => {
   const { pushNotification } = useContext(NotificationsContext);
+  const baseProviderList = useMigrationProviderList();
+  const providerList = useMemo(
+    () =>
+      unique([initialTarget?.provider, ...baseProviderList]).filter(Boolean),
```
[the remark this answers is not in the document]
sjd78: The useMemo is more in anticipation that the list won't be static in the near future and come from a tag category or something similar.

rszwajko (approved):

PR #2231, opened 2025-04-17
:seedling: Upgrade rollup and plugins

sjd78:
Update rollup to version to `4.40.0`.  There were some issues with the native builder on qemu builds for some platforms, but those have been addressed or the platforms are no longer part of the builds.

Updated rollup plugins to `latest` versions.

Add some JSDoc based typing to the rollup config files.


sjd78:
Replaces #2227

rszwajko (approved):

PR #2232, opened 2025-04-18
:seedling: Update ID locator for actions element in application tables

nachandr:
The actions element has a different ID locator on the Controls pages (Eg: Stakeholder groups, stakeholders etc.) .
So, I'm updating the ID locator for the element in application tables to keep it consistent. This will also help us with automation code maintenance .


nachandr:
@sjd78 , Could you review this PR ?

sjd78 (approved):
LGTM

PR #2240, opened 2025-04-23
:bug: Analysis config excluded labels are labels not tags

sjd78:
When creating the Taskgroup's Data to run an analysis, the user entered values for the "Excluded rules labels" field should be passed along as labels to be excluded, not tags to be excluded.

Fixes: https://issues.redhat.com/browse/MTA-4973


rszwajko (approved):

sjd78:
Related PR: https://github.com/konveyor/tackle2-addon-analyzer/pull/150

PR #2313, opened 2025-05-20
:seedling: Upgrade webpack, express, proxies, dev mode

sjd78:
Upgrade and refactor the packages to use the latest express server, http proxy, webpack and related.  Some of the upgrades has breaking changes that needed to be addressed.

Summary of changes:

  - In `client`, upgrade webpack and related plugins.  Fixup configs as necessary.

  - Move proxy definitions to the server package

  - In dev mode, put the express server in front of the `webpack-dev-server` (proxying `/` to the dev server) so the proxies run the same way in dev mode and prod mode

  - Fixed breaking changes for express and `http-proxy-middleware` with major version upgrades.

  - Add a setting to `.npmrc` to try to make sure dependency resolved and integrity fields are included in the lock file.

  - Update `README.md` with the new dev server structure

  - `cookies.ts` was removed since it wasn't used

  - `rules-utils.ts` fixed lint error on a string compare conditional

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,7 +92,7 @@ The React and Patternfly based UI is composed of web pages served by an http ser
 - In **production** mode, the application code is served as statically built UI assets. A small
   handler inserts relevant environment information on the root page. This configures how the UI
   runs in the browser. The server's listener port is configurable via environment variables and
-  defaults to **:9000**.
+  defaults to **:8080**.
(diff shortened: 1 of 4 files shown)


rszwajko (approved):
Nice improvement! The server part will be now used more frequently.

PR #2361, opened 2025-06-11
:bug: Fix Configuring a ssh connection to a git repository fails

marcoklaassen:
Resolves: https://issues.redhat.com/browse/MTA-5385
https://github.com/konveyor/operator/issues/441

Covers repository URL strings in:
  - Application source code, source repository
  - Application analysis wizard, custom rules step, repository tab, source repository
  - Custom migration targets, Retrieve from a repository, source repository

Use [git-url-parse](https://github.com/IonicaBizau/git-url-parse) to handle git URLs.  Subversion URLs are handled by the same RegEx and previous.

### DIFF
--- a/client/src/app/pages/applications/components/application-form/application-form.tsx
+++ b/client/src/app/pages/applications/components/application-form/application-form.tsx
@@ -8,11 +8,10 @@ import {
   Popover,
   PopoverPosition,
 } from "@patternfly/react-core";
-import { useForm } from "react-hook-form";
+import { useForm, useWatch } from "react-hook-form";
 import { yupResolver } from "@hookform/resolvers/yup";
 
 import { SimpleSelect, OptionWithValue } from "@app/components/SimpleSelect";
-// import { SimpleSelectTypeahead } from "@app/components/SimpleSelectTypeahead";
 import { DEFAULT_SELECT_MAX_HEIGHT } from "@app/Constants";
 import { Application, New, TagRef } from "@app/api/models";
 import { duplicateNameCheck, getAxiosErrorMessage } from "@app/utils/utils";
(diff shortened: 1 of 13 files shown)


client/src/app/utils/utils.ts
```diff
@@ -110,17 +110,21 @@ export const getValidatedFromError = (error: unknown | undefined) => {
...
 export const standardStrictURLRegex =
   /https:\/\/(www\.)?[-a-zA-Z0-9@:%._\\+~#=]{2,256}\.[a-z]{2,4}\b([-a-zA-Z0-9@:%_\\+.~#?&//=]*)/;
 
 export const svnUrlRegex = /^svn:\/\/[^\s/$.?#].[^\s]*$/;
 
 export const customURLValidation = (schema: yup.StringSchema) => {
   const containsURL = (string: string) =>
-    gitUrlRegex.test(string) ||
+    gitRegularUrlRegex.test(string) ||
+    gitSshUrlRegex.test(string) ||
```
sjd78, line 127: That the right thing to do for sure.  I'm looking at a library that has a more comprehensive set of checks.  I'll push a change to your PR if it works nicely.

sjd78 (commented):
Change is the right approach.  I'll update with a bit more robust checking and unit tests.

client/src/app/utils/utils.test.ts
```diff
@@ -76,12 +77,16 @@ describe("utils", () => {
 
     it("Regex should validate git URLs", () => {
       const testGitURLs = [
+        "ssh://git@github.com:konveyor/tackle2-ui.git",
```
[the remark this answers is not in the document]
sjd78: Turns out the `git@` prefix defaults to the ssh protocol.

sjd78 (approved):

client/src/app/pages/applications/components/application-form/application-form.tsx
```diff
@@ -166,25 +162,27 @@ export const useApplicationFormHook = ({
       comments: string()
         .trim()
         .max(250, t("validation.maxLength", { length: 250 })),
+
+      // source code fields
+      kind: string().oneOf(["git", "subversion"]),
```
rszwajko, line 167: @sjd78 
It seems that the validation is too restrictive right now - you cannot create an app without a source repo.
Although the repository is marked as required in the form it's effectively optional as you can decide to use only binaries.

rszwajko (changes requested):

client/src/app/pages/applications/components/application-form/application-form.tsx
sjd78: Fixed finally by rearranging the schema and using the value of the kind field to indicate if the repository field is required.

sjd78:
(closed/reopened to refresh links to jira)

Issue #2283, opened 2025-05-08
Create source platform page and handling


sjd78:
## Summary
For a [source platform](https://github.com/konveyor/tackle2-ui/tree/main/docs/enhancements/platform-awareness-asset-generation#source-platform), create a new page, table, detail drawer, and add/edit modal.

## Enabling Work 
HUB enablement issues:
- https://github.com/konveyor/tackle2-hub/issues/809
- https://github.com/konveyor/tackle2-hub/issues/810

UI enablement issues:
- #2284 

## New Work
Page:
- New navigation location: Administration / Source Platforms
- The page will be a standard table view with a detail drawer and add/edit modal

Table view:
- Columns:
  - Name
    - Task status popover (for discover applications tasks)
  - Provider type (based on available discovery providers rest endpoint)
  - Credentials attached?
  - :thinking: Link to platform coordinates schema defined fields
  - :thinking: Count of associated applications
- Toolbar actions:
  - Create
- Record kebab actions:
  - Edit
  - Delete
  - Discover applications

Select item drawer:
- Base source platform information
- View of the platform coordinates schema defined fields
- View of the application coordinates schema

Add/edit modal:
- Name input
- Provider type dropdown (based on available discovery providers rest endpoint)
- Credentials dropdown
  - Note: Include "None" as a valid option since it's not required.
  - Note: The credentials must already exist.  In future, the ability to create a credential inline would be nice-to-have.
- Expandable Section for the platform coordinates [schema defined fields](#2284)

## Actions
- Discover applications: #2289

PR #2368, opened 2025-06-12
:sparkles: add source platforms page to the ui

ElayAharoni:
https://issues.redhat.com/browse/MTA-5243
Part of: #2283

Added the source platform page to the UI, still there is more work to do there but this is the initial one.

## Summary by CodeRabbit

- **New Features**
  - Introduced a new "Platforms" section to manage source platforms, accessible from the sidebar navigation.
  - Added functionality to create, edit, and delete platforms, with forms supporting validation and feedback.
  - Enabled viewing platform details, including coordinates and associated applications, in a dedicated drawer UI.
  - Provided a table view for platform applications with filtering, sorting, and pagination.
  - Added the ability to trigger application discovery tasks for platforms.
  - Included a JSON editor component for editing platform configuration metadata with schema validation.

- **Improvements**
  - Enhanced localization with new terms, messages, and validation error handling related to platforms.
  - Updated navigation and routing to support the new platforms feature.

- **User Interface**
  - Added new components for platform details, forms, and application tables.
  - Improved UI feedback through notifications, loading states, and empty state messages.

### DIFF
--- a/client/src/app/pages/source-platforms/components/platform-applications-table.tsx
+++ b/client/src/app/pages/source-platforms/components/platform-applications-table.tsx
@@ -1,0 +1,170 @@
+import * as React from "react";
+import { useTranslation } from "react-i18next";
+import { Toolbar, ToolbarContent, ToolbarItem } from "@patternfly/react-core";
+import { Table, Tbody, Td, Th, Thead, Tr } from "@patternfly/react-table";
+import spacing from "@patternfly/react-styles/css/utilities/Spacing/spacing";
+import { Ref } from "@app/api/models";
+import {
+  useTableControlState,
+  useTableControlProps,
+  deserializeFilterUrlParams,
+} from "@app/hooks/table-controls";
+import { TablePersistenceKeyPrefix } from "@app/Constants";
+import {
+  ConditionalTableBody,
+  TableHeaderContentWithControls,
+  TableRowContentWithControls,
+} from "@app/components/TableControls";
+import { SimplePagination } from "@app/components/SimplePagination";
+import { FilterToolbar, FilterType } from "@app/components/FilterToolbar";
+import { useHistory } from "react-router-dom";
+import { useFetchApplications } from "@app/queries/applications";
+
+export interface IPlatformAppsTableProps {
+  platformApplications?: Ref[];
+}
+
+const PlatformAppsTable: React.FC<IPlatformAppsTableProps> = ({
+  platformApplications,
+}) => {
+  const { t } = useTranslation();
+  const history = useHistory();
+
+  const urlParams = new URLSearchParams(window.location.search);
+  const filters = urlParams.get("filters");
+  const deserializedFilterValues = deserializeFilterUrlParams({ filters });
+
+  const {
+    data: baseApplications,
+    isFetching: isFetchingApplications,
+    error: applicationsFetchError,
+  } = useFetchApplications();
+
+  const getPlatformApplication = () => {
+    const platformApplicationIds = platformApplications?.map(
+      (platformApplication) => platformApplication.id
+    );
+    return baseApplications?.filter((app) =>
+      platformApplicationIds?.includes(app.id)
+    );
+  };
+
+  const tableApplications = getPlatformApplication();
+
+  const tableControlState = useTableControlState({
+    tableName: "platform-apps-table",
+    persistTo: "urlParams",
+    persistenceKeyPrefix: TablePersistenceKeyPrefix.platformApplications,
+    columnNames: {
+      name: "Application",
+    },
+    isFilterEnabled: true,
+    isSortEnabled: true,
+    isPaginationEnabled: true,
+    sortableColumns: ["name"],
+    initialSort: { columnKey: "name", direction: "asc" },
+    initialFilterValues: deserializedFilterValues,
+    filterCategories: [
+      {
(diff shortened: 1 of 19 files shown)


client/src/app/pages/source-platforms/components/platform-detail-drawer.tsx
```diff
@@ -0,0 +1,125 @@
...
+              </DescriptionListGroup>
+            </DescriptionList>
+          </Tab>
+          <Tab
+            eventKey={TabKey.Coordinates}
+            title={<TabTitleText>{t("terms.coordinates")}</TabTitleText>}
+          >
+            <DescriptionList className="platform-coordinates-drawer-list">
+              <SchemaDefinedField />
+            </DescriptionList>
```
sjd78, line 100: Does the `DescriptionList` wrapper need to be there?  It makes sense in the Details tab, but not so much in the Coordinates or Applications tabs.

client/src/app/pages/source-platforms/components/platform-applications-table.tsx
```diff
@@ -0,0 +1,221 @@
...
+
+  const urlParams = new URLSearchParams(window.location.search);
+  const filters = urlParams.get("filters");
+  const deserializedFilterValues = deserializeFilterUrlParams({ filters });
+
+  const {
+    data: baseApplications,
+    isFetching: isFetchingApplications,
+    error: applicationsFetchError,
+  } = useFetchApplications();
```
sjd78, line 41: It would be really great if this hook accepted a filter/param to only return applications that match the platform's application refs.  That would probably need to be added later.  Best to create a new issue referencing this idea so we could do it in future.

client/src/app/pages/source-platforms/components/platform-applications-table.tsx
```diff
@@ -0,0 +1,221 @@
...
+//       {mavenCentralLink ? (
+//         <ExternalLink isInline href={mavenCentralLink}>
+//           {version}
+//         </ExternalLink>
+//       ) : (
+//         <Text>{version}</Text>
+//       )}
+//     </TextContent>
+//   );
+// };
```
sjd78, line 221: Don't forget to remove these comments.

client/src/app/pages/source-platforms/source-platforms.tsx
```diff
@@ -0,0 +1,437 @@
...
+  }, [basePlatforms]);
+
+  const onError = (error: AxiosError) => {
+    pushNotification({
+      title: getAxiosErrorMessage(error),
+      variant: "danger",
+    });
+  };
+
+  const { mutate: deleteplatform } = useDeletePlatformMutation(
```
sjd78, line 103:
```suggestion
  const { mutate: deletePlatform } = useDeletePlatformMutation(
```

Camel casing on variable names keeps my code spell checker happy.

client/src/app/pages/source-platforms/source-platforms.tsx
```diff
@@ -0,0 +1,437 @@
...
+  useEffect(() => {
+    if (basePlatforms) {
+      setPlatforms(
+        basePlatforms.map((platform) => ({
+          ...platform,
+          discoverApplicationsState: "No task" as TaskState,
+        }))
+      );
+    }
+  }, [basePlatforms]);
```
sjd78, line 94: I suggest extracting this to a hook so the complexity doesn't live here.  The `useDecoratedApplication` hook is overkill in this scenario, but it is an example of further enhancing the base query hook.

sjd78 (changes requested):
- Source Platform (SP) empty state looks good
- SP create form functions properly, some visual tweaks needed
- SP detail drawer:
  - Details tab looks good
  - Coordinates/schemas tab needs to be clarified (see below), second viewing of the tab and the code editor doesn't display properly, first view should have the editor height take up available space without having the scroll the drawer.  Content of the editor may be scrolled.
  - Applications tab looks good
- SP table
  - layout looks good
  - row kebab looks good
  - delete confirm looks good


Each Source Platform can potentially have 2 sets of schemas:
  1. Platform coordinates jsonschema -- Define a document stored with the SP instance _IF_ we end up needing more than the single URL to define connection parameters to the platform itself.  I don't think we're doing this right now.
  2. Application coordinates jsonschema -- Schema is provided by the SP provider (its kind) to define the "platform coordinates" that an application needs to have defined to be able to run the retrieve configurations task.  This schema itself may be useful to view on the SP details drawer but will not be editable.  Every Application associated to the SP will have a coordinates document at least at the time of starting the retrieve configurations task.

In future we may need to review the labels for the kind field (currently Provider Type).

client/src/app/pages/source-platforms/source-platforms.tsx
ElayAharoni: you sure it will be right to extract it just for this logic? if we will have more properties to add i would agree

sjd78 (approved):
Coordinates tab/SchemaDefinedFields...
  - The details drawer coordinates schema render needs to address code-editor height issues. This can be addressed in a followup PR when adding proper schema defined field handling.
  - Can add json schema validation directly to the editor.  See the comment inline on how to do that. This can also be done in a followup PR.
  - The padding on the Coordinate tab content is off.

Source platform table...
  - When implementing the application discovery workflow, the task status part of the name column will need to be updated.

Issue #2285, opened 2025-05-08
Create application discovery manifest view and handling


sjd78:
## Summary
For a [discovery manifest](https://github.com/konveyor/tackle2-ui/tree/main/docs/enhancements/platform-awareness-asset-generation#discovery-manifest), which is the result of retrieving configurations from a source platform for a single application, enable views and handling.  Since this manifest only relates to applications, the view and actions are all attached to an application.

## Enabling Work
HUB enablement issues:
- https://github.com/konveyor/tackle2-hub/issues/808

UI enablement:
- The view can be based on #2284's components.

## Changes
On the application table:
- :thinking: Add an icon to indicate the status of the "discovery manifest" for the application's source platform:
  - "not ready": Application is not ready to retrieve configurations (no source platform assigned, or application platform coordinates aren't valid)
  - "ready": Application is ready to retrieve configurations (application platform coordinates are valid)
  - "failed: Configuration retrieval attempted but failed
  - "discovered": Configurations have been retrieved
  - :spiral_notepad: This _could_ be redundant info if the retrieval task is reported in the application task info popover

On application edit:
- If the application's source platform selection is changed, the manifest needs to be cleared

Options to view the manifest:
- Tab on the application edit modal
- Separate link to open the manifest in its own modal using #2284's components from the kebab or from a tab in the details drawer

PR #2424, opened 2025-06-30
:sparkles: create application discovery manifest view and handling

ElayAharoni:
https://issues.redhat.com/browse/MTA-5244
https://issues.redhat.com/browse/MTA-5601

part of: #2285

added the application discovery manifest API endpoints, and the veiw of it inside the application side drawer, as a readonly code editor

## Summary by CodeRabbit

* **New Features**
  * Added a "Manifest" tab to the application detail drawer, displaying the application's manifest in a formatted JSON view.
  * Introduced the ability to fetch and display manifest data for each application.
  * Enhanced translation support with a new "manifest" key in English.

* **Improvements**
  * The SchemaDefinedField component now supports optional class names and improved read-only behavior based on the presence of a save handler.

* **Bug Fixes**
  * Corrected a conditional expression in the application detail drawer for accurate archetype display.

### DIFF
--- a/client/src/app/api/models.ts
+++ b/client/src/app/api/models.ts
@@ -875,6 +875,8 @@ export interface GroupedStakeholderRef extends Ref {
   uniqueId: string;
 }
 
+export interface JsonDocument extends Record<string, unknown> {}
+
 export interface SourcePlatform {
   id: number;
   kind: string;
@@ -882,7 +884,7 @@ export interface SourcePlatform {
   url: string;
   identity?: Ref;
   applications?: Ref[];
-  coordinates?: Record<string, any>;
+  coordinates?: JsonDocument;
   discoverApplicationsState?: TaskState;
 }
 export interface ManifestDeployment {
@@ -894,16 +896,10 @@ export interface ManifestSecret {
   user?: string;
   password?: string;
 }
-export interface ManifestContent {
-  name: string;
-  password?: string;
-  deployment?: ManifestDeployment;
-  service?: { name: string; other?: number };
-}
 
 export interface Manifest {
   id: number;
-  content: ManifestContent;
+  content: JsonDocument;
   secret: ManifestSecret;
   application?: Ref;
 }
--- a/client/src/app/api/models.ts
+++ b/client/src/app/api/models.ts
@@ -885,3 +885,25 @@ export interface SourcePlatform {
   coordinates?: Record<string, any>;
   discoverApplicationsState?: TaskState;
 }
+export interface ManifestDeployment {
+  name: string;
+  other?: number;
+}
+
+export interface ManifestSecret {
+  user?: string;
+  password?: string;
+}
+export interface ManifestContent {
+  name: string;
+  password?: string;
+  deployment?: ManifestDeployment;
+  service?: { name: string; other?: number };
+}
+
+export interface Manifest {
+  id: number;
+  content: ManifestContent;
+  secret: ManifestSecret;
+  application?: Ref;
+}
--- a/client/src/app/api/rest.ts
+++ b/client/src/app/api/rest.ts
@@ -47,8 +47,10 @@ import {
   SourcePlatform,
   AnalysisDependency,
   AnalysisAppDependency,
+  Manifest,
 } from "./models";
 import { serializeRequestParamsForHub } from "@app/hooks/table-controls";
+import { template } from "radash";
 
 // endpoint paths
 export const HUB = "/hub";
@@ -88,6 +90,10 @@ export const TICKETS = HUB + "/tickets";
 export const TRACKER_PROJECT_ISSUETYPES = "issuetypes"; // TODO: ????
 export const TRACKER_PROJECTS = "projects"; // TODO: ????
 export const TRACKERS = HUB + "/trackers";
+
+export const MANIFESTS = HUB + "/manifests";
+
+export const APPLICATION_MANIFEST = HUB + "/applications/{{id}}/manifest";
 
 const jsonHeaders: RawAxiosRequestHeaders = {
   Accept: "application/json",
@@ -275,6 +281,13 @@ export const deleteBulkApplications = (ids: number[]): Promise<Application[]> =>
 
 export const getApplicationById = (id: number | string): Promise<Application> =>
   axios.get(`${APPLICATIONS}/${id}`).then((response) => response.data);
+
+export const getApplicationManifest = (
+  applicationId: number | string
+): Promise<Manifest> =>
+  axios
+    .get<Manifest>(template(APPLICATION_MANIFEST, { id: applicationId }))
+    .then((response) => response.data);
 
 export const getApplications = (): Promise<Application[]> =>
   axios.get(APPLICATIONS).then((response) => response.data);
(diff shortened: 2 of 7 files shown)


client/src/app/api/models.ts
```diff
@@ -885,3 +886,25 @@ export interface SourcePlatform {
...
+  deployment?: ManifestDeployment;
+  service?: { name: string; other?: number };
+}
+
+export interface Manifest {
+  id: number;
+  content: ManifestContent;
+  secret: ManifestSecret;
+  application?: Ref;
+}
```
sjd78, line 910: The `content` and `secret` attributes can just be `unknown` (instead of string) at this point.

See: https://github.com/konveyor/tackle2-hub/blob/main/api/manifest.go#L290-L295

client/src/app/api/rest.ts
```diff
@@ -846,3 +854,23 @@ export const updatePlatform = (platform: SourcePlatform) =>
...
+export const createManifest = (manifest: New<Manifest>) =>
+  axios.post<Manifest>(MANIFESTS, manifest).then((res) => res.data);
+
+// success with code 204 and therefore no response content
+export const updateManifest = (manifest: Manifest): Promise<void> =>
+  axios.put(`${MANIFESTS}/${manifest.id}`, manifest);
+
+// success with code 204 and therefore no response content
+export const deleteManifest = (id: number): Promise<void> =>
+  axios.delete(`${MANIFESTS}/${id}`);
```
sjd78, line 876: Frontend will not need to deal with these at all.  we'll just access via the `/applications/:id/manifest` endpoint.

client/src/app/api/rest.ts
```diff
@@ -284,6 +287,11 @@ export const deleteBulkApplications = (ids: number[]): Promise<Application[]> =>
 export const getApplicationById = (id: number | string): Promise<Application> =>
   axios.get(`${APPLICATIONS}/${id}`).then((response) => response.data);
 
+export const getApplicationManifestById = (
+  id: number | string
+): Promise<Manifest> =>
+  axios.get(`${APPLICATIONS}/${id}/manifest`).then((response) => response.data);
```
sjd78, line 293:
```suggestion
export const getApplicationManifest = (
  applicationId: number | string
): Promise<Manifest> =>
  axios.get(`${APPLICATIONS}/${applicationId}/manifest`).then((response) => response.data);
```

client/src/app/components/schema-defined-fields/SchemaDefinedFields.tsx
```diff
@@ -13,32 +13,31 @@ export { Language } from "@patternfly/react-code-editor";
...
 
   console.log("SchemaDefinedField rendered with jsonDocument:", jsonDocument);
+
   const initialCode = React.useMemo(
     () => JSON.stringify(jsonDocument, null, 2),
     [jsonDocument]
   );
-  const [currentCode, setCurrentCode] = React.useState(initialCode); // State to hold editable content
-
+  const [currentCode, setCurrentCode] = React.useState(initialCode);
```
sjd78, line 40: The `jsonDocument` param is speced to be an object or a string.  If it is a string, we don't want to stringify it.  Add a typeof test to handle the variations properly.

client/src/app/components/schema-defined-fields/SchemaDefinedFields.tsx
```diff
@@ -61,7 +60,9 @@ export const SchemaDefinedField = ({
   const handleSave = () => {
     if (editorRef.current) {
       const contentToSave = editorRef.current.getValue();
-      onSchemaSaved(JSON.parse(contentToSave));
+      onSchemaSaved
+        ? onSchemaSaved(JSON.parse(contentToSave))
```
sjd78, line 64: Similar comment as above.  If the jsonDocumet (the thing being edited) is an object, probably want to send an object.  If it is a string, probably want to send a string.

client/src/app/pages/applications/components/application-detail-drawer/application-detail-drawer.tsx
```diff
@@ -182,6 +184,14 @@ export const ApplicationDetailDrawer: React.FC<
               <TabTasksContent application={application} task={task} />
             </Tab>
           )}
+          {!application || !application.manifest ? null : (
+            <Tab
+              eventKey={TabKey.Manifest}
+              title={<TabTitleText>{t("terms.manifest")}</TabTitleText>}
+            >
+              <SchemaDefinedField jsonDocument={application.manifest} />
```
sjd78, line 192: Need to make sure the margins/padding around the component is consistent with the other tab contents

client/src/app/pages/applications/useDecoratedApplications.ts
```diff
@@ -60,6 +62,8 @@ export interface DecoratedApplication extends Application {
     identities?: Identity[];
     archetypes?: Archetype[];
   };
+
+  manifest?: Manifest;
```
sjd78, line 66: Since this is only useful to the detail drawer tab that would take at least 2 user clicks to get to, there is no need to fetch the manifest for all applications here.

Instead, have the tab contents component pull the application manifest directly and pass it along to the editor.

client/src/app/queries/applications.ts
```diff
@@ -71,6 +72,24 @@ export const useFetchApplicationById = (id?: number | string) => {
   };
 };
 
+export const useFetchApplicationManifestById = (id?: number | string) => {
```
sjd78, line 75: The manifest is being fetched by the application id, not the manifest id.

```suggestion
export const useFetchApplicationManifest = (applicationId?: number | string) => {
```

client/src/app/queries/manifests.ts
sjd78, line 1: I don't think these queries are needed.  We will only need to access the manifest read-only via the application path.  Only the Task addOn will need to use the manifest endpoints.

sjd78 (changes requested):
In summary, focus on the `/applications/:id/manifest` endpoint and keep it as light as possible.

client/src/app/components/schema-defined-fields/SchemaDefinedFields.tsx
```diff
@@ -13,32 +13,31 @@ export { Language } from "@patternfly/react-code-editor";
...
-  // If you wanted to make it dynamic to load other schemas:
   jsonDocument: object | string;
-  jsonSchema: object | string;
+  jsonSchema?: object | string;
   onSchemaSaved?: (newSchemaContent: object) => void;
 }
 
 export const SchemaDefinedField = ({
-  onSchemaSaved = () => {},
+  onSchemaSaved,
```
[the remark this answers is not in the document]
ElayAharoni: changed that on a different pr

client/src/app/components/schema-defined-fields/SchemaDefinedFields.tsx
ElayAharoni: changed on different pr

client/src/app/queries/manifests.ts
ElayAharoni: i can see that all endpoints are using the queries, maybe it is good that all will be aligned the same?

client/src/app/api/models.ts
ElayAharoni: why do you think that it should be unknown?

sjd78: Use `unknown` because we don't know the shape of the content.  It could be `object` or it could be `Record<string, string>` if we know it'll be a flat map.

Really the `unknown` just forces a cast of the type before using it.

sjd78 (changes requested):
Two things and it's ready to merge:
  - Drop the unused manifest rest functions and queries.
  - Add a `TabManifestContent` component to the drawer.

sjd78 (approved):
LGTM

Issue #2284, opened 2025-05-08
Enable working with a a set of schema defined fields


sjd78:
Since platform awareness requires the ability to attach a dynamic set of fields to at least a platform entity, being able to handle a set [schema defined fields](https://github.com/konveyor/tackle2-ui/tree/main/docs/enhancements/platform-awareness-asset-generation#schema-defined-fields) is required.

From a json-schema associated to an entity through an entity type (e.g. Source platform coordinates through the platform's type), we need to be able to support:

- Display of a schema
- View a document with a schema via patternfly components
- View a document with a schema via the `@patternfly/react-code-editor`
- Edit a document with a schema via patternfly components
- Edit a complex document with a schema via the `@patternfly/react-code-editor`
- Know when to choose the patternfly render/edit vs the react-code-editor render/edit (probably based on schema complexity)
- 🤔 Allow the user to select the render/edit view

Note: This is similar to how OCP Console handles entities at various points -- some as form fields, some as the yaml document.

PR #2432, opened 2025-07-02
:sparkles: Enable working with a set of schema defined fields

ElayAharoni:
https://issues.redhat.com/browse/MTA-5242

Resolves: #2284

## Summary by CodeRabbit

* **New Features**
  * Introduced a form-based editor that dynamically generates fields from a provided JSON schema, supporting validation and internationalization.
  * Added a toggle to switch between a code editor view and a form view for editing JSON documents, with automatic disabling for complex schemas.
  * Implemented a new JSON schema code editor with save functionality and user-friendly features such as copy, download, and loading states.
  * Added a save control button for improved user interaction during document editing.
  * Enhanced input components to support an "integer" type for more precise number handling.

* **Bug Fixes**
  * Updated JSON schema to remove obsolete fields and adjust required properties for improved accuracy.

* **Chores**
  * Added new validation message translations and a "Reset" action label in the English locale.
  * Updated component props to use consistent naming for JSON documents across the app.
  * Changed proxy form port inputs to use the new "integer" input type for better data consistency.
  * Added accessibility improvements with IDs for helper text in form components.

### DIFF
--- a/client/src/app/api/models.ts
+++ b/client/src/app/api/models.ts
@@ -912,6 +912,7 @@ export interface Manifest {
   application?: Ref;
 }
 
+// Could use https://www.npmjs.com/package/@types/json-schema in future if needed
 export interface JsonSchemaObject {
   type: "string" | "integer" | "number" | "boolean" | "object" | "array";
   title?: string;
--- a/client/src/app/components/schema-defined-fields/utils.tsx
+++ b/client/src/app/components/schema-defined-fields/utils.tsx
@@ -1,12 +1,17 @@
 import { JsonSchemaObject } from "@app/api/models";
 import { yupResolver } from "@hookform/resolvers/yup";
+import { TFunction } from "i18next";
 import * as yup from "yup";
 
 export const jsonSchemaToYupSchema = (
   jsonSchema: JsonSchemaObject,
-  t: (key: string, options?: Record<string, unknown>) => string = (k, v) =>
-    `${k}: ${JSON.stringify(v)}`
+  translate?:
+    | TFunction<"translation", undefined>
+    | ((k: string, v?: object) => string)
 ): yup.AnySchema => {
+  const fallbackT = (k: string, v?: object) => `${k}: ${JSON.stringify(v)}`;
+  const t = translate || fallbackT;
+
   if (jsonSchema.type === "array") {
     let schema = yup.array();
     if (jsonSchema.items) {
@@ -21,7 +26,10 @@ export const jsonSchemaToYupSchema = (
   }
 
   if (jsonSchema.type === "object" && jsonSchema.properties) {
-    const props: Record<string, any> = {};
+    const props: Record<
+      string,
+      yup.AnySchema | ReturnType<typeof yup.lazy<yup.AnySchema>>
+    > = {};
 
     for (const [key, prop] of Object.entries(jsonSchema.properties)) {
       const propSchema = jsonSchemaToYupSchema(prop, t);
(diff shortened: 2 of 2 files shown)


client/src/app/api/models.ts
```diff
@@ -885,3 +885,11 @@ export interface SourcePlatform {
...
   discoverApplicationsState?: TaskState;
 }
+
+export interface JsonSchemaObject {
+  type: string;
+  properties?: { [key: string]: any };
+  required?: string[];
+  title?: string;
+  description?: string;
+}
```
sjd78, line 895: There is https://www.npmjs.com/package/@types/json-schema.  It only covers up to draft 7, but may still be useful

client/src/app/components/schema-defined-fields/SchemaAsFields.tsx
```diff
@@ -0,0 +1,238 @@
...
+}
+
+export const SchemaAsFields: React.FC<SchemaAsFieldsProps> = ({
+  jsonDocument,
+  jsonSchema,
+  onDocumentSaved,
+}) => {
+  const { t } = useTranslation();
+
+  const jsonSchemaToYupSchema = (
```
sjd78, line 43: This should be a util function that can have some unit tests.  We'd want a few unit tests that can prove...

Given a simple schema (one property in an object), turn it into a yup validator and then run validation against a good and a bad sample.  That will let us be reasonable sure that what we say can work will work w/o having to manually test everything.

All that being said, looks like decent coverage.

There is always something like https://www.npmjs.com/package/json-schema-yup-transformer that could be adopted if more exotic use cases come up.

client/src/app/components/schema-defined-fields/SchemaAsFields.tsx
```diff
@@ -0,0 +1,238 @@
...
+              )}
+            </CardBody>
+          </Card>
+        );
+    }
+  };
+
+  const onValidSubmit = (values: JsonSchemaValues) => {
+    onDocumentSaved && onDocumentSaved(values);
+  };
```
sjd78, line 205: Integration back with a larger form will probably be needed.

For example, when we want to have platform coordinates on an application and we want that to be on the standard application edit form.  In that case, the document will live on the application but be managed by this component.  So easiest is to post document changes all the way out to the containing form and have that form somehow know that the document is valid according to the schema.

Really what does that mean right now?  Probably:
  - the submit, cancel, reset buttons shouldn't be here
  - every field change should call a `onDocumentChange()`

Changes to the data would be validated locally (by the field or the code editor) for display AND the containing form would also want to validate the document as a whole.  It's double validation but that's ok.  We don't have anywhere to live test it now other than just to console.log changes somewhere.

I expect the data flow and display in this component will be tweaked over time as we get further away from happy path schemas and data.

sjd78 (changes requested):
Lots of comments but in summary, you're on the right path.  Restructuring a few things to support unit testing of core util functions is necessary.

Also, the view path is ok but the edit path will need work to properly integrate with a containing form.

client/src/app/components/schema-defined-fields/utils.tsx
```diff
@@ -0,0 +1,149 @@
...
+      } else {
+        // For optional object properties with properties, use lazy validation
+        // to avoid validating internal structure when the object is undefined
+        if (prop.type === "object" && prop.properties) {
+          props[key] = yup.lazy((value) => {
+            if (value === undefined || value === null) {
+              return yup.mixed().optional();
+            }
+            return propSchema;
+          }) as unknown as yup.AnySchema;
```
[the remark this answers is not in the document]
sjd78: Close.  With my version of yup, that typing doesn't work.  But `ReturnType<typeof yup.lazy<yup.AnySchmea>>` does.

PR #2501, opened 2025-07-29
:sparkles: Finalize Source Platforms

sjd78:
Follow-up: #2368

Close gaps on source platform to match what is in the operator for konveyor-0.8:
  - provider list is just: [cloud-foundry]
  - platforms do no have schema defined coordinates


ibolton336 (approved):
VISACK

PR #2594, opened 2025-09-02
:ghost: Fix KAI proxy configuration 

ibolton336:
- Update KAI proxy to use KAI_URL environment variable instead of TACKLE_HUB_URL
- Remove /services/kai path prefix for direct KAI service communication


server/src/proxies.js
```diff
@@ -108,12 +108,12 @@ export default {
 
   kai: {
     pathFilter: "/kai",
-    target: KONVEYOR_ENV.TACKLE_HUB_URL || "http://localhost:9002",
+    target: KONVEYOR_ENV.KAI_URL || "http://localhost:9003",
```
sjd78, line 111: FWIW, in the current setup on the dev server, ports go like this:

:9000 -- server proxy to webpack dev server
:9001 -- port forward to keycloak
:9002 -- port forward to hub
:9003 -- webpack dev server

I swapped the server-in-dev-mode around so the base port :9000 is always used and always is handled by the express server first.

Issue #2602, opened 2025-09-10
Missing tooltip text on 'Create new platform' wizard


mguetta1:
https://issues.redhat.com/browse/MTA-6124

PR #2604, opened 2025-09-10
:bug: Fix missing tooltip text in create platform wizard

mguetta1:
Resolves: https://issues.redhat.com/browse/MTA-6124
Resolves: #2602 

  - Use currently selected platform kind instead of platform prop for URL tooltip
  - Add fallback tooltip text for unknown platform kinds
  - Fixes https://issues.redhat.com/browse/MTA-6124 where tooltip was missing in create wizard but present in update wizard

[Screencast_20250910_153810.webm](https://github.com/user-attachments/assets/16ae8b51-bdca-43ac-a24c-d5f721ac05ba)

## Summary by CodeRabbit

* **New Features**
  * Platform URL field shows a dynamic tooltip that updates immediately based on the selected platform type.
  * Clear fallback tooltip text is displayed when the platform type is unknown.

* **Chores**
  * Added English translation for the platform URL tooltip to improve clarity and i18n coverage.
  * Form now tracks the selected platform type in real time for a more responsive UI.

### DIFF
--- a/client/public/locales/en/translation.json
+++ b/client/public/locales/en/translation.json
@@ -291,6 +291,11 @@
     "retain": "Retain",
     "retire": "Retire"
   },
+  "repositoryKind": {
+    "git": "Git",
+    "subversion": "Subversion",
+    "none": "None"
+  },
   "risks": {
     "high": "High",
     "low": "Low",
--- a/client/src/app/hooks/useRepositoryKind.ts
+++ b/client/src/app/hooks/useRepositoryKind.ts
@@ -1,0 +1,28 @@
+import { useMemo } from "react";
+import { useTranslation } from "react-i18next";
+
+import { OptionWithValue } from "@app/components/SimpleSelect";
+
+export type RepositoryKind = "git" | "subversion" | "" | null;
+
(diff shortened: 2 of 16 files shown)


client/src/app/pages/source-platforms/components/platform-form.tsx
```diff
@@ -212,7 +214,7 @@ const PlatformFormRenderer: React.FC<PlatformFormProps> = ({
         fieldId="url"
         isRequired
         labelIcon={
-          <Popover bodyContent={<div>{getUrlTooltip(platform?.kind)}</div>}>
+          <Popover bodyContent={<div>{getUrlTooltip(selectedKind)}</div>}>
```
sjd78, line 217: What about hiding the popover if there is no tool tip to display?

sjd78 (approved):
This is ok.

I added [useRepositoryKind()](https://github.com/konveyor/tackle2-ui/blob/main/client/src/app/hooks/useRepositoryKind.ts) that uses a different approach for static list with translated keys.  It would be good to harmonize around a common approach if/when we have to update again, or if there is another hook that provides a set of static data.

PR #2610, opened 2025-09-12
:bug: Add identity role handling to application manage credentials

sjd78:
Part 1 of 2: https://issues.redhat.com/browse/MTA-6122
Part 1 of 2: https://github.com/konveyor/tackle2-ui/issues/2599
Requires hub changes in: https://github.com/konveyor/tackle2-hub/pull/911

Use kind=source identities for asset repositories by adding a role to the `application.identities` array. Removal of the management of kind=asset identities will be done in a later PR. The identity `Ref` roles will be used by HUB and AddOns as appropriate instead of relying on the identity's kind.

Manage Credentials modal:
  - Remains unchanged visually
  - Will retain any identities that do not have a role of source, maven, or asset
  - Each drop down assigns a role to the selected identity

## Summary by CodeRabbit

* **New Features**
  * Role-based identity management for applications with role-aware defaults and selection (source, maven, asset).

* **Bug Fixes**
  * Preserves existing non-managed identities when updating.
  * Improved validation and clearer error feedback to prevent duplicate or mismatched identities.

* **Refactor**
  * Consolidated identities into a unified top-level structure and streamlined identity fetching for more consistent form behavior.

### DIFF
--- a/client/src/app/pages/applications/application-identity-form/application-identity-form.tsx
+++ b/client/src/app/pages/applications/application-identity-form/application-identity-form.tsx
@@ -42,11 +42,11 @@ export interface FormValues {
   asset: number | null;
 }
 
-const MANAGED_IDENTITY_ROLES: ManagedIdentityRole[] = [
+const MANAGED_IDENTITY_SET: Set<IdentityRole> = new Set([
   "source",
   "maven",
   "asset",
-] as const;
+]);
 
 function identitiesToOptions(
   identities?: Identity[]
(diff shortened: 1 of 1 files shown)


client/src/app/api/models.ts
```diff
@@ -90,6 +90,10 @@ export interface IdRef {
   id: number;
 }
 
+export interface RefWithRole<RoleType = string> extends Ref {
```
rszwajko, line 93: Not sure if it's beneficial (long term) to create another type of `Ref`. This requires special handling all over the code base - i.e. in methods like `toRef` which narrow down to pure Ref (in Konveyor almost everything is RefLike). Consider the options:
1. extend  `Ref` - since the prop is optional this should work with the existing code base (but you need to handle all the code that converts the Refs)
2. switch to composition i.e. ` RefWithRole = { ref: Ref, role: string}`. If the type is used only locally the the overhead is minimal and type system will help to avoid bugs.

PR #2626, opened 2025-09-22
:bug: Update assessment form IDs with UUID

aufi:
Resolves: https://issues.redhat.com/browse/MTA-6143
Resolves: #2653

Updating assessment questionnaires to not construct fields IDs/names with questions&answers tests, but use UUIDs.

Fixes misfunctions caused by too long questions&answers strings.

Used claude code to help create this PR.

## Summary by CodeRabbit

- Bug Fixes
  - Stabilized Assessment Wizard steps and questionnaire items to prevent jumps, duplicates, or missing steps.
  - Preserves answers, input focus, and selection state when navigating or when question/option order changes.
  - Ensures consistent option behavior and avoids accidental resets during edits.

- Performance
  - Smoother, more reliable rendering with fewer re-renders and reduced flicker across the wizard and forms.
  - Faster navigation between sections and questions.

- Other
  - No visual design changes; improvements focus on stability and consistency of the assessment experience.


aufi:
> I installed the questionnaire attached to MTA-6143, and took the assessment for an application with the PR running. I could click all the way through the wizard but the assessment actions page shows the questionnaire as "Continue" instead of the "Retake" I expected:
> 
That is right, I can confirm this, the original questionnaire had a wrong order number for second section (already mentioned it to reporter on slack), that causes this behaviour. With questionnaire data fix (in the original file, change line 142 oder: 1->2), it works well for me locally.

image

Issue #2629, opened 2025-09-25
Add source platform discover applications wizard to the application inventory page


sjd78:
Reference: https://issues.redhat.com/browse/MTA-6120

Add source platform application discovery along side the existing "Import Applications" action on the application inventory page level action kebab.

Location:

<img width="1677" height="1180" alt="Image" src="https://github.com/user-attachments/assets/1c07b5b9-27bd-491f-b45b-951babd3df13" />

PR #2646, opened 2025-10-01
:bug: Add discover application wizard to the applications table

sjd78:
Resolves: https://issues.redhat.com/browse/MTA-6120
Resolves: #2629

## Summary
The source platform discovery applications wizard is now available in the applications table. A new step to select the source platform has been added to the wizard. The wizard itself is now a shared component that is both from the applications table and the source platforms page.

Now there are 3 import/discover actions on the application inventory:
  - Import applications from CSV
  - Manage application imports
  - Discover applications

Summary of changes:
  - Move the discover import wizard to a shared component
  - Add a reducer value/setter for the platform
  - Allow easy initialization of the wizard state
  - Add a select platform step to the wizard if no platform is provided
  - Add `usePlatformKindList` hook
  - Update `useRepositoryKind` hook to match i18n use
  - Update the action labels for import/discover actions on application inventory

## Demo

https://github.com/user-attachments/assets/4b2f6dcd-05eb-49ce-80b8-74f4155bcf40

## Summary by CodeRabbit

* **New Features**
  * Added a Platform selection step, a Select Platform UI, and a screen shown when no platform is selected.
  * Discovery Import Wizard can be opened from the Applications table.

* **Refactor**
  * Replaced legacy no-platform modal with inline step flow and gated steps until a platform is chosen.
  * Wizard state now tracks the selected platform and exposes platform setter/reset behavior.

* **Style**
  * Updated copy to use “source platform”; added step titles/descriptions and action label updates (e.g., “Import applications from CSV”, “Manage application imports”).

* **Chores**
  * Standardized imports, consolidated exports, removed debug logging, and added new runtime dependencies.

### DIFF
--- a/client/src/app/components/discover-import-wizard/useWizardReducer.ts
+++ b/client/src/app/components/discover-import-wizard/useWizardReducer.ts
@@ -33,6 +33,7 @@ type WizardAction =
 
 const updateIsReady = (draft: WizardState) => {
   draft.isReady = !!draft.platform && draft.filters.isValid;
+  return draft;
 };
 
 const wizardReducer: WizardReducer = (draft, action) => {


client/src/app/components/discover-import-wizard/discover-import-wizard.tsx
```diff
@@ -129,41 +110,64 @@ const DiscoverImportWizardInner: React.FC<IDiscoverImportWizard> = ({
...
           }}
         >
-          {!results ? (
-            <Review platform={platform} filters={filters} />
-          ) : (
-            <Results results={results} />
-          )}
+          {platform ? (
+            !results ? (
+              <Review platform={platform} filters={filters} />
```
rszwajko, line 166: I think it would be more readable to handle missing platform/results inside `Review` component (return null ?). This way we could just:
```
 <Review platform={platform} filters={filters}  results={results}/>
```

client/src/app/components/discover-import-wizard/select-platform.tsx
```diff
@@ -0,0 +1,65 @@
...
+  platform: SourcePlatform | null;
+  onPlatformSelected: (platform: SourcePlatform | null) => void;
+}
+
+export const SelectPlatform: React.FC<SelectPlatformProps> = ({
+  platform,
+  onPlatformSelected,
+}) => {
+  const { t } = useTranslation();
+  const { platforms } = useFetchPlatforms();
```
rszwajko, line 22: are we sure the results are in cache? if not we should handle the loading state

client/src/app/components/discover-import-wizard/useWizardReducer.ts
```diff
@@ -20,17 +24,20 @@ const INITIAL_WIZARD_STATE: WizardState = {
...
 type WizardReducer = (state: WizardState, action: WizardAction) => WizardState;
 type WizardAction =
+  | { type: "SET_PLATFORM"; payload: SourcePlatform | null }
   | { type: "SET_FILTERS"; payload: FilterState }
   | { type: "SET_RESULTS"; payload: ResultsData | null }
   | { type: "RESET" };
 
 const validateWizardState = (state: WizardState): WizardState => {
-  const isReady = state.filters.isValid;
+  const isReady = state.platform !== null && state.filters.isValid;
```
rszwajko, line 33: just in case I would use a wider condition for the platform ` platform !== null => !!platform`

client/src/app/components/discover-import-wizard/useWizardReducer.ts
```diff
@@ -45,14 +52,47 @@ const wizardReducer: WizardReducer = (state, action) => {
 const validatedReducer: WizardReducer = (state, action) =>
   validateWizardState(wizardReducer(state, action));
 
-export const useWizardReducer = () => {
+// Helper function to create initial state using the same reducer actions
+const createInitialState = (
```
rszwajko, line 56: 1. Are those conditions mutually exclusive?
2. early return would simplify reasoning

rszwajko (commented):
Looks good! some minor comments below

client/src/app/components/discover-import-wizard/select-platform.tsx
sjd78: Never sure sure that results are in cache, but really I'm not concerned with loading state here.

client/src/app/components/discover-import-wizard/discover-import-wizard.tsx
sjd78: I want to keep the Review and Results components separate.  I can push the null handling of platform inside the Review component.  That does make sense.

client/src/app/components/discover-import-wizard/useWizardReducer.ts
sjd78: Humm, I'll try a different way.

PR #2804, opened 2025-12-16
:test_tube: Fix analysis wizard test flake with next button

sjd78:
When running the an analysis wizard test, the next button on the custom rules step was not immediately enabled.  The form requires a few renders to properly validate and enable the Next button.  The tests would click right away, and sometimes it would be enabled and sometimes not.  This would cause the next "Next" click to hit the custom rules step instead of the advanced options step.  In that case, the test would fail.

This fix adds a wait for the next button to be enabled before clicking it.  This wait allows the form to validate, update state, and the next buttons to be enabled.

First noticed in CI working on #2747

## Summary by CodeRabbit

* **Tests**
  * Updated test automation to enhance step progression and logging in the analysis workflow.
  * Improved test utility for better button state verification.

* **Chores**
  * Enhanced form field wiring for improved consistency.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>


sjd78:
Merged with the failing e2e testing to unblock other PRs.  The test fail is being addressed in #2805.

Issue #2882, opened 2026-01-27
[BUG] Profile selection list is not scrollable, preventing access to all options


sshveta:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Konveyor version

upstream

### Priority

Major

### Current Behavior

Description:
The profile selection dropdown/list does not provide a scrollbar when the number of profiles exceeds the visible area. As a result, users cannot scroll down to view or select profiles that appear beyond the initially visible portion of the list.

<img width="2113" height="1122" alt="Image" src="https://github.com/user-attachments/assets/9abfc5b5-2474-4db1-97ab-7436aa21a497" />

### How Reproducible

Always (Default)

### Steps To Reproduce

1. Create an application, Navigate to the profile selection screen.
2. Open the profile selection dropdown/list.(Make sure you have 12/13 profiles)
3. Observe the list when there are more profiles than can fit in the visible area.

The list is fixed in size and does not scroll, making profiles beyond the visible area inaccessible.

### Environment

```markdown
- OS:
```

PR #2948, opened 2026-02-05
:bug: Use isScrollable by default in SimpleSelect

rszwajko:
Fixes:  https://github.com/konveyor/tackle2-ui/issues/2882

## Summary by CodeRabbit

* **New Features**
  * Enabled scrollable dropdown options for improved navigation through long option lists.


sjd78:
@coderabbitai Can you check the repo for use of this component and try to evaluate if turning on scrolling by default will break use anywhere else?  Is the component used in a static height scenario that should be address here as well?

PR #2959, opened 2026-02-06
:test_tube: Mark tests with Upstream bugs

sshveta:
Tests can be marked with upstream bugs now .
pattern "Bug Tackle-XXX"

## Summary by CodeRabbit

* **Tests**
  * Updated many test titles to add or remove bug IDs and include issue links; minor reliability tweaks (added brief waits).
* **Chores**
  * Restricted bug-reference extraction to specific project prefixes; made download cleanup more robust by ensuring the downloads folder exists.
* **Fixtures**
  * Normalized an excludePackages field to an array and adjusted an effort value in YAML.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,7 +9,7 @@
 #   - https://github.com/konveyor/tackle2-ui/pull/1781
 
 # Builder image
-FROM registry.access.redhat.com/ubi9/nodejs-20:9.7-1770021187 AS builder
+FROM registry.access.redhat.com/ubi9/nodejs-20:9.7-1770626460 AS builder
 
 USER 1001
 COPY --chown=1001 . .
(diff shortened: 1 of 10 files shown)


cypress/scripts/bug_summary.mjs
```diff
@@ -4,7 +4,7 @@ import { readFile, writeFile } from "fs/promises";
 
 import { table } from "table";
 
-const bugPattern = /Bug\s+([A-Z]+-\d+)/i;
+const bugPattern = /Bug\s+((?:MTA|Tackle|TACKLE)-\d+)/i;
```
nachandr: How about we add tackle2-hub, tackle2-ui for upstream issues?

The it() block would then look like this :
 it("Bug Tackle2-hub-966: Create test data", function () {}

PR #3005, opened 2026-02-19
:bug: Memoize validateSettingsXml

rszwajko:
Partial fix for:  https://issues.redhat.com/browse/MTA-6007

Due to memoization the extra requests are fired only when xml is changed.

## Summary by CodeRabbit

* **Refactor**
  * Validation now caches results so repeated validation of the same settings returns cached outcomes, reducing repeated work and improving responsiveness.
  * Cache handling is improved to avoid stale results by clearing cache when the input value changes, while preserving existing validation behavior and error handling.


rszwajko:
## After


https://github.com/user-attachments/assets/5eccaf64-bcb1-4741-866a-208f6fdee824

PR #3175, opened 2026-04-08
:test_tube: Removed task manager unstable test (#3174)

sshveta:
* **Tests**
  * Removed an end-to-end test covering task manager flows.
* Test cleanup now uses full bulk deletion of related entities instead of list-driven deletion.

* **Chores**
* Simplified test runner configuration by removing a file-preprocessor setup.

---------


cypress/cypress.config.ts
```diff
@@ -124,14 +123,6 @@ export default defineConfig({
       }
 
       // Plugins
-      // Esbuild bundler configured with sourcemaps for better stack traces
-      on(
```
nachandr, line 128: I'm wondering why this has been removed.

sshveta: This is throwing error in InterOp tests reported by them .

nachandr (approved):

PR #3201, opened 2026-04-21
:test_tube: Removed npm ci

sshveta:


nachandr (approved):

cypress/dockerfiles/interop/Dockerfile
```diff
@@ -18,7 +18,7 @@ RUN apt-get update && \
 RUN mkdir /tmp/tackle-ui-tests
 WORKDIR /tmp/tackle-ui-tests
 COPY . .
-RUN npm ci
+RUN npm install .
```
sjd78, line 21: Just curious -- why isn't the clean-install working?  Does the lock file not get copied into the container?

Does the interop build not run from the repo root?

sshveta: yes , they are running from cypress dir

Issue #3238, opened 2026-05-06
UI | Permissions instead of roles.


jortel:
## Migration from Role-Based to Scope-Based Permissions                                                                                                                  
                                         
  ### Overview                                                                                                                                                             
  Migrated all permission checks throughout the UI from role-based access control to scope-based access control to align with OIDC standards where permissions are encoded
  as OAuth2 scopes in JWT tokens.                                                                                                                                          
                                                            
  ### Background                                                                                                                                                           
  With the migration to OIDC, user permissions are now represented as scopes in the JWT token's `scope` claim (space-separated string) rather than as roles in a `roles`
  claim (array). All permission checks needed to be updated to extract and validate scopes instead of roles.                                                               
   
  ### Key Changes                                                                                                                                                          
                                                            
  #### JWT Token Extraction                                                                                                                                                
  Changed from extracting `roles` array to extracting `scope` string from JWT claims:
                                                                                                                                                                           
  **Before:**                                                                                                                                                              
  ```typescript                                                                                                                                                            
  const roles = claims?.roles; // Array of role names                                                                                                                      
                                                            
  After:                                                                                                                                                                   
  const scopeString = claims?.scope as string | undefined;
  const scopes = scopeString.split(" "); // Space-separated scopes                                                                                                         
                                                                                                                                                                           
  Modified Files                                                                                                                                                           
                                                                                                                                                                           
  client/src/app/rbac.ts                                                                                                                                                   
                                                                                                                                                                           
  - Deprecated role constants:                                                                                                                                             
    - devRoles = ["admin", "architect", "migrator"]
    - adminRoles = ["admin"]                                                                                                                                               
  - Added scope constants for route-level access:           
  export const administrationScopes = ["settings:put"];                                                                                                                    
  export const migrationScopes = ["applications:get"];                                                                                                                     
  - Added scope constants for analysis profiles:                                                                                                                           
  export const analysisProfilesWriteScopes = [                                                                                                                             
    "analysis.profiles:post",                                                                                                                                              
    "analysis.profiles:put",                                                                                                                                               
    "analysis.profiles:delete",                                                                                                                                            
  ];                                                                                                                                                                       
  export const analysisProfilesReadScopes = ["analysis.profiles:get"];
  - Updated RBAC component:                                                                                                                                                
    - Changed extractRolesFromToken() to extract from roles claim
    - Added extractPermissionScopes() to extract from scope claim                                                                                                          
[...]

PR #3260, opened 2026-05-14
 :sparkles: Migrate auth from @react-keycloak/web to oidc-client-ts + react-oidc-context

### LINKED ISSUE #3202: Migrate to current OIDC library
As per https://github.com/konveyor/enhancements/issues/268, migrate off the existing keycloak library use to a modern OIDC library for authentication.

sjd78:
Implements #3202 per the ui-oidc enhancement proposal:
https://github.com/konveyor/enhancements/tree/master/enhancements/ui-oidc

Removes the unmaintained keycloak-js and @react-keycloak/web packages
in favor of the actively-maintained oidc-client-ts and react-oidc-context
libraries. No operator or deployment changes are required. The existing
KEYCLOAK_REALM, KEYCLOAK_CLIENT_ID, and /auth base env vars are reused
as-is.

Key changes:
- New `@app/auth` module as the single source of truth for all auth
  state and authorization. Feature code imports only from this barrel.
  Direct access to oidc-client-ts or token internals is prohibited.

- **Strategy-pattern AuthProvider** selects the concrete auth strategy
  at runtime. All three strategies publish their state through a shared
  `AuthStateContext`, and each strategy owns its own masthead toolbar UI
  (via the `ToolbarContent` field on `AuthState`):
    - `OidcAuthStrategy` — live OIDC provider session (Authorization
      Code + PKCE, sessionStorage token persistence, automatic sign-in
      redirect on cold load, silent token renewal via signinSilent()).
    - `MasqueradeAuthStrategy` — dev-mode persona switcher backed by
      localStorage; swaps admin/architect/migrator presets without a
      page reload. Tree-shaken from production bundles.
    - `NoAuthStrategy` — hardcoded full-admin roles for production
      when auth is disabled.

- Clean hook API: `useAuth()`, `useHasRealmRoles()`, `useHasScopes()`,
  `useIsArchitect()` replace all prior direct reads of
  `keycloak.tokenParsed`. Hooks read from `AuthStateContext` and never
  call `useOidcAuth()` directly, avoiding circular dependencies.

- Realm roles correctly decoded from the access token JWT (Keycloak
  places `realm_access.roles` in the access token, not the ID token).

- Axios auth interceptors are only registered when `AUTH_REQUIRED=true`,
  double-registration guarded to survive React StrictMode, and include
  an `_authRetry` flag to prevent infinite 401 retry loops. On 401 the
  interceptor attempts `signinSilent()` → retry once → `signinRedirect()`
  on failure.

- `RouteWrapper` simplified to use `useHasRealmRoles()` directly instead
  of the old `checkAccess()` helper.

Behavior matrix (AUTH_REQUIRED × NODE_ENV):
  true  + any env      → OidcAuthStrategy — real OIDC session
  false + development  → MasqueradeAuthStrategy — persona panel visible
  false + production   → NoAuthStrategy — roles pinned to full-admin

Removed:
  - `keycloak-js`, `@react-keycloak/web` dependencies
  - keycloak specific components and singleton

Added:
  - `oidc-client-ts ^3.5.0`, `react-oidc-context ^3.3.1`
  - `client/src/app/auth/` module (AuthProvider, OidcAuthStrategy,
    MasqueradeAuthStrategy, NoAuthStrategy, MasqueradeDevPanel,
    OidcToolbarItem, hooks, userManager, masquerade, types, barrel index)
  - `client/src/app/layout/HeaderApp/AuthProviderToolbarItem.tsx`

Cypress e2e tests may need updates for changed redirect URIs or session
storage keys — tracked as follow-up work.

Fixes: #3202

Assisted-by: goose:claude-sonnet-4-6
Assisted-by: Cursor:claude-opus-4-6

## Summary by CodeRabbit

* **New Features**
  * Migrate authentication to OpenID Connect (OIDC) with a unified AuthProvider, runtime strategy selection, and developer masquerade personas for non-production.

* **Updates**
  * Replace SSO toolbar with a user account menu (username, Manage account, Logout).
  * RBAC and UI permissions now use centralized auth hooks; HTTP requests handle token injection and silent refresh.

* **Tests**
  * Test setup updated to mock the new auth/OIDC layer for reliable CI.

[Review Change Stack](https://app.coderabbit.ai/change-stack/konveyor/tackle2-ui/pull/3260?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


client/src/app/pages/applications/applications-table/applications-table.tsx
```diff
@@ -339,6 +337,7 @@ export const ApplicationsTable: FC = () => {
         }),
         variant: "success",
       });
+    // eslint-disable-next-line react-hooks/immutability
```
rszwajko: at least in my setup the warning displayed here is a different one `Error: Cannot access variable before it is declared` and can be fixed with moving entire `tableControls` block up.

client/src/app/layout/HeaderApp/tests/__snapshots__/HeaderApp.test.tsx.snap
```diff
@@ -336,7 +389,60 @@ exports[`Test snapshot 1`] = `
               <div
                 class="pf-v6-c-toolbar__group pf-m-hidden pf-m-visible-on-md"
                 id="header-toolbar-sso"
-              />
+              >
+                <div
+                  class="pf-v6-c-toolbar__item"
+                >
+                  <button
```
rszwajko: I think we should not test dev tools here - can we set AUTH_REQUIRED or NODE_ENV to disable the masquerade dev panel?

client/src/app/auth/AuthProvider.tsx
```diff
@@ -0,0 +1,110 @@
+/**
+ * auth/AuthProvider.tsx
```
rszwajko: 1. I'd rather avoid file names in the comments as it complicates refactoring.
2. the comments refer to the code that it's no longer available - this makes sense only during migration. 
3. the comments are not linked to the actual function/component - if they are general design then let's put them in the *.md file

rszwajko (commented):
The direction looks good! I like the concept of Masquarade Dev Panel. 
Some high level comments below.

client/src/app/axios-config/apiInit.ts
```diff
@@ -1,41 +1,66 @@
...
 
-import keycloak from "@app/keycloak";
+import { isAuthRequired } from "@app/Constants";
+import { userManager } from "@app/auth/userManager";
+
+// Guard against duplicate interceptor registration (e.g. React StrictMode double-invoke).
+let _initialized = false;
 
 export const initInterceptors = () => {
+  if (_initialized) return;
```
rszwajko, line 21: we could skip the initialization for no-auth case here - no need to add any interceptors.

client/src/app/axios-config/apiInit.ts
```diff
@@ -1,41 +1,66 @@
...
   axios.interceptors.response.use(
-    (response) => {
-      return response;
-    },
+    (response) => response,
     async (error) => {
-      if (error.response && error.response.status === 401) {
+      if (!isAuthRequired) return Promise.reject(error);
+
+      if (error.response?.status === 401) {
```
rszwajko, line 41: 1. with one condition we can use early return
2. do we need to guard against redirect loop?

client/src/app/auth/AuthProvider.tsx
sjd78: I have addresses the comments in general as Claude did things a bit weird and used British spellings in places.  JSDocs around will be updated across the PR.

client/src/app/layout/HeaderApp/tests/__snapshots__/HeaderApp.test.tsx.snap
sjd78: I consider this a low priority concern.  Maybe have the masquerade panel only render in dev and use the noauth=admin strategy in test and prod builds.

In this PR, I'm not super concerned on getting this right and having a "what is useful" discussion and adjust from there.

client/src/app/pages/applications/applications-table/applications-table.tsx
sjd78: I'm just going to remove the comment and leave it as-is. Fixing this is a bit beyond the scope of the PR.

client/src/app/axios-config/apiInit.ts
```diff
@@ -1,41 +1,65 @@
...
-                Authorization: `Bearer ${keycloak.token}`,
+                Authorization: `Bearer ${refreshedUser.access_token}`,
               },
             };
             return axios(retryConfig);
           }
         } catch {
-          keycloak.login();
+          // Refresh failed — redirect to login.
+          await userManager.signinRedirect();
```
rszwajko: the call can throw - we probably need a catch here

client/src/app/auth/hooks.ts
```diff
@@ -0,0 +1,90 @@
...
+  }
+  return authState;
+};
+
+// ── useHasRealmRoles ──────────────────────────────────────────────────────────
+
+/**
+ * Returns true when the current user has at least one of the supplied realm roles.
+ *
+ * When auth is disabled, masquerade roles are checked so RBAC behaves the same
```
rszwajko: 1. statement is incorrect
2.  `isAuthRequired` should be used sparingly - all current usages should be reviewed

rszwajko (approved):
Let's address the remaining issues in the follow-ups

client/src/app/axios-config/apiInit.ts
sjd78: Changed to be a single refresh/retry chain.  If the silent refresh fails, or the retry fails, then send the user to login. Bad things could still happen, but this shouldn't loop.

client/src/app/auth/hooks.ts
sjd78: I cleaned up the hooks.  More refactoring will be done with #3238.

Issue #2498, opened 2025-07-27
Update the Generator parameter handling to view, edit and produce a simple JSON schema document


sjd78:
Following up on #2291, the generator parameters need to be upgraded.  The parameters field under a generator needs to live as a simple json schema document.  This will allow us to use the `SchemaDefinedFields` technique to capture the user input when a user is walking through the generate assets workflow.

How the parameters form should work:
  - The form should still allow add and delete of rows.
  - Each row should have a "Name" and a type dropdown.
  - The allowed types should be "string", "number", "integer" or "boolean".
  - Each type _could_ allow further restrictions such as:
    - required/optional
    - string: min length, max length, regex, well known regex formats (ipv4 address, url)
    - number: min, max
    - integer: min, max

The existing `SchemaDefiendFields` component should be able to properly render all of those types as PF input fields and allow easy capture of data from the user during the workflow.  

The `combineSchemas()` function used in `step-capture-parameters` of the genearte assets wizard will also need to be aligned with these changes.

PR #3283, opened 2026-05-25
:ghost: Remove export modifier from local-only code

### LINKED ISSUE #3280: Remove unused exports
Knip command:
```
 npx knip --exports
```
reports:
1. Unused exports (270)
2. Unused exported types (105)
3. Unused exported enum members (9)
4. Duplicate exports (18)

rszwajko:
Remove the code that was detected as unused after
removing export modifier.

Fixes: #3280 
Reference-Url: https://knip.dev/
Assisted-by: Cursor:claude-4.6-sonnet

## Release Notes

* **Refactor**
  - Consolidated internal type exports and reduced the public API surface to improve code maintainability.
  - Removed unused internal constants, query keys, and helper functions throughout the application.
  - Reorganized module exports to better distinguish between public interfaces and internal implementation details.

[Review Change Stack](https://app.coderabbit.ai/change-stack/konveyor/tackle2-ui/pull/3283?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


eslint.config.mjs
```diff
@@ -66,6 +66,7 @@ export default defineConfig([
       ],
       "@typescript-eslint/no-explicit-any": "warn",
       "@typescript-eslint/no-redeclare": "error",
+      "no-unused-vars": "off",
```
sjd78: I thought the `tseslint.configs.recommended` disabled all the conflicting eslint base rules.  Maybe it actually does not since this one is explicitly enabled in the block above?

client/src/app/queries/tasks.ts
sjd78, line 1: Does it really make sense to drop these queries even if they're not currently being used?

client/src/app/queries/schemas.ts
sjd78, line 1: Similar to the Tasks, does it make sense to remove the queries even if they're not currently being used?

client/src/app/pages/applications/generate-assets-wizard/step-capture-parameters.tsx
sjd78, line 1: We want to keep this file as it would come back with #2498

sjd78 (changes requested):
A few questions....

client/src/app/queries/tasks.ts
rszwajko: I would drop all dead code. If we really need it than we have the git history.

sjd78 (approved):
LGTM

PR #3303, opened 2026-06-04
:seedling: Replace webpack with rspack and rollup with rslib

sjd78:
## Summary
Bundeler migration:
  - webpack to [rspack](https://rspack.rs/)
  - common:rollup to [rslib](https://rslib.rs/)
  - server:rollup to [rslib](https://rslib.rs/)

## Why? 
Migrating to the rspack/rslib platform reduce the amount of time needed to build/package significantly.

Before:
  - common build: 3.1 seconds
  - server build: 1.9 seconds
  - client build: 106 seconds
  - total `npm run build` time: 110 seconds

After
  - common build: 2.4 seconds
  - server build: 0.6 seconds
  - client build: 7.7 seconds
  - total `npm run build` time: 26 seconds

Note: OpenShift console dynamic plugins will be able to use rspack soon (it supports the required module federation), so wider adoption should happen in the near future.

Assisted-by: Cursor:claude-opus-4-6

## Summary by CodeRabbit

* **New Features**
  * Development builds now automatically restart the server after rebuilding.
  * Branding content is included consistently across library outputs.
  * Added a command-line entry point for starting the server.

* **Bug Fixes**
  * Clarified server startup and shutdown messages across environments.

* **Chores**
  * Updated client, server, and shared-library build tooling for more consistent builds.
  * Improved packaging of branding assets, source maps, and TypeScript declarations.

* **Tests**
  * Improved Jest support for selected modern module dependencies.


server/src/index.js
```diff
@@ -22,7 +22,7 @@ debugMode && console.log("KONVEYOR_ENV", KONVEYOR_ENV);
 const __dirname = fileURLToPath(new URL(".", import.meta.url));
 const pathToClientDist = path.join(__dirname, "../../client/dist");
 
-const port = parseInt(KONVEYOR_ENV.PORT, 10) || developmentMode ? 9000 : 8080;
+const port = parseInt(KONVEYOR_ENV.PORT, 10) || (developmentMode ? 9000 : 8080);
```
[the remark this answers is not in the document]
sjd78: PORT 0 doesn't make sense, so the expression does make sense

client/config/rspack.prod.mts
```diff
@@ -0,0 +1,70 @@
...
+      ),
+      favicon: faviconPath,
+      minify: {
+        collapseWhitespace: false,
+        keepClosingSlash: true,
+        minifyJS: true,
+        removeEmptyAttributes: true,
+        removeRedundantAttributes: true,
+      },
+    }),
```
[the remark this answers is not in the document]
sjd78: The server expects the `index.html.ejs` and will render the template with the correct at runtime.  This issue is not an issue.

package.json
```diff
@@ -83,8 +76,7 @@
     "mocha": {
       "serialize-javascript": "^7.0.6"
     },
-    "webpack-dev-server": {
-      "express": "$express"
-    }
+    "follow-redirects": "^1.16.0",
+    "qs": "^6.14.1"
```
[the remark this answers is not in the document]
sjd78: Will be handled in another PR

common/rslib-plugin-build-marker.ts
```diff
@@ -0,0 +1,108 @@
...
+            order: "post",
+            handler: ({ isFirstCompile }) => {
+              if (
+                isFirstCompile &&
+                totalEnvironments > 0 &&
+                finishedEnvironments.size === totalEnvironments
+              ) {
+                api.logger.success("first build complete, writing marker file");
+                writeMarker();
+              }
```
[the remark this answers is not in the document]
sjd78: Given the plugin and build lifecycles of rslib, the best way to determine "on the first successful build, and only on the first successful build, create the marker file". This way other builds that are waiting for the marker don't start if the build fails, and also only start once the full build is complete.  If there are gaps in understanding of the lifecycles, what would be a better way to handle this?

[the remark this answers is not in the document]
sjd78: This produces the marker file BEFORE the type definitions are created.  The type definitions are needed by the dependent modules to type check and build properly. The "virtual modules" also build after the marker file. There needs to be another check somehow.  The isFirstCompile flag is putting the marker generation in the proper place

client/config/rspack.common.mts
```diff
@@ -0,0 +1,127 @@
...
+      },
+
+      // Monaco CSS -- always style-loader (separate from PF CSS pipeline)
+      {
+        test: /\.css$/,
+        include: [pathTo("../../node_modules/monaco-editor")],
+        use: ["style-loader", "css-loader"],
+      },
+
+      // XSD schemas and YAML imported as raw strings
```
rszwajko, line 69: It seems we miss ` resourceQuery: /raw/,`  (check [this guide](https://rspack.rs/guide/features/asset-module#replacing-raw-loader))
(applies also to svg "raw" import above)

client/config/rspack.common.mts
```diff
@@ -0,0 +1,127 @@
...
+
+      // Fonts (PF + Monaco) -- .woff2 referenced via CSS @font-face url()
+      { test: /\.(woff2?|ttf|eot)$/, type: "asset/resource" },
+
+      // App-local SVGs imported as raw strings
+      { test: /\.svg$/, include: [pathTo("../src")], type: "asset/source" },
+
+      // Images -- inline if small, emit as file if large
+      {
+        test: /\.(jpg|jpeg|png|gif)$/i,
```
rszwajko, line 57: in old config we filtered images/fonts by location (basically only PF and local). Is it safe to remove this requirement?

client/config/rspack.common.mts
```diff
@@ -0,0 +1,127 @@
...
+              parser: { syntax: "typescript", tsx: true },
+              transform: { react: { runtime: "automatic" } },
+            },
+            detectSyntax: "auto",
+          },
+        },
+        type: "javascript/auto",
+      },
+
+      // Fonts (PF + Monaco) -- .woff2 referenced via CSS @font-face url()
```
rszwajko, line 49: it seems we are processing all fonts regardless of location (contrary to the comment).

client/config/rspack.common.mts
```diff
@@ -0,0 +1,127 @@
...
+    extensions: [".js", ".ts", ".tsx", ".jsx"],
+    tsConfig: {
+      configFile: pathTo("../tsconfig.json"),
+    },
+    symlinks: false,
+    fallback: { crypto: false, fs: false, path: false },
+  },
+
+  // monaco-editor: ignore warnings from static code analysis, allow mocks without warning
+  ignoreWarnings: [/Critical dependency: require function is used in a way/],
```
rszwajko, line 120: why we need this ignore? can we limit it to monaco editor only (as the comment claims)?

common/rslib.config.ts
```diff
@@ -0,0 +1,86 @@
...
+const __dirname = fileURLToPath(new URL(".", import.meta.url));
+
+const baseBrandingPath = process.env.BRANDING ?? "./branding";
+const brandingPath = path.resolve(__dirname, "../", baseBrandingPath);
+const jsonStrings = JSON.parse(
+  readFileSync(path.resolve(brandingPath, "./strings.json"), "utf8")
+);
+const stringsModule = render(
+  `
+  export const strings = ${JSON.stringify(jsonStrings)};
```
rszwajko, line 19: can we use ordinary loaders and  import the json file in a standard way?

common/rslib-plugin-build-marker.ts
```diff
@@ -0,0 +1,122 @@
...
+function createPrefixLogger(api: RsbuildPluginAPI, prefix: string) {
+  return createLogger({
+    ...api.logger.options,
+    prefix: color.dim(prefix),
+    level: api.logger.level,
+  });
+}
+
+/**
+ * A rslib plugin which will create/touch a marker file once the bundles are first generated. This
```
rszwajko, line 41: 1. seems overkill - can we wait on some real emitted file instead? maybe we could clean the files before the build and wait on first file emitted?
2. rollup version seems simpler - maybe we can keep it? (applies also to the server package)

server/rslib-plugin-run.ts
```diff
@@ -0,0 +1,93 @@
...
+      return new Promise<void>((resolve) => {
+        child!.on("close", () => resolve());
+        controller!.abort();
+      });
+    };
+
+    // Redirect SIGINT/SIGTERM to a clean process.exit(0) so rslib's normal
+    // shutdown runs and triggers api.onExit below. Without this, Node's default
+    // handler exits with code 130 and npm reports "npm error code 130".
+    process.on("SIGINT", () => process.exit(0));
```
rszwajko, line 50: it feels low level - I'd rather use some library or keep rollup here.

client/config/rspack.common.mts
sjd78: Yes. That was a flourish in an older PF example project.

client/config/rspack.common.mts
sjd78: We don't _need_ to ignore it, but there is no fix since it comes from monaco sources before the warned code gets shaken out. This is much nicer in the build logs.  It could go in the dev config instead, but then there would be a log difference between dev and prod builds.  I'd rather keep those as close to the same as possible.

client/config/rspack.common.mts
sjd78: The `resourceQuery` would be for matching the rule to an import with a query string.  We don't have any `import foo from "someSchema.xsd?raw"`.

The imports we do use just go `import foo from "someSchema.xsd"` and the test captures that.

common/rslib.config.ts
sjd78: No. The module is a typed struct and this is actually an easy way to setup the module dynamically before the build without having to modify any other config file.  It could be done by updating the tsconfig.json with a specific alias to a file generated before the build, but that is much messier.

Besides, this was debated when I initially did the branding work years ago.

common/rslib-plugin-build-marker.ts
sjd78: This is a very different discussion over all.

I went with the marker since there wasn't a single file consistently emitted at the right time.  And it is so there is no hacky "start the server after a length of time the common normally takes to build" config.

server/rslib-plugin-run.ts
sjd78: There isn't a library for this already available.

sjd78:
Scope reduced and replaced by #3553

common/rslib.config.ts
rszwajko: @sjd78 
yeah ...this is a clear sign that we need a comment here :)

PR #3351, opened 2026-06-15
:test_tube: Fixed some failing tests due to OIDC in auth in nightly runs

sshveta:
## Summary by CodeRabbit

* **Tests**
  * Ensured e2e scenarios authenticate before reading custom metrics, downloading static reports, or capturing dependent state.
  * Reworked migration and dependency tests to perform cleanup via authenticated API deletes with auth headers.
  * Updated RBAC migration coverage to use the correct architect workflow, including expanded stakeholder cleanup.

* **Bug Fixes**
  * Strengthened session validation during login by verifying OIDC session data and failing fast when access tokens are missing or expired.

### DIFF
--- a/cypress/e2e/tests/migration/applicationinventory/analysis/source_analysis_without_creds.test.ts
+++ b/cypress/e2e/tests/migration/applicationinventory/analysis/source_analysis_without_creds.test.ts
@@ -21,7 +21,6 @@ import {
   cleanupDownloads,
   deleteAllAnalysisProfiles,
   deleteApplicationTableRows,
-  deleteBulkApplicationsByApi,
   exists,
   getProfileNameFromApp,
   getRandomAnalysisData,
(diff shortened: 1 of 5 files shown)


cypress/e2e/tests/migration/applicationinventory/analysis/source_analysis_without_creds.test.ts
```diff
@@ -187,10 +190,11 @@ describe(["@tier0"], "Tier 0 Analysis and Static Report validation ", () => {
...
   after("Perform test data clean up", function () {
     login();
-    cy.visit("/");
-    deleteBulkApplicationsByApi(applicationIds);
-    credentialsList.forEach((credential) => credential.delete());
-    profilesToDelete.forEach((profile) => profile.delete());
+    getAuthHeaders().then((headers) => {
+      Application.deleteAllViaApi(headers);
+      Credentials.deleteAllViaApi(headers);
+      AnalysisProfile.deleteAllViaApi(headers);
```
sjd78, line 196: The old deletes were based off a list of things to delete, presumably matching things created during the tests.  Do the new deletes still use those lists or do they really delete ALL of those entities?

Deleting ALL of those entities while running in a local dev environment would be bad.

ibragins (approved):
LGTM

cypress/e2e/tests/migration/dynamic-report/dependencies/filter_sorting_pagination.test.ts
```diff
@@ -73,6 +76,9 @@ describe(
       cy.visit("/");
       deleteAllMigrationWaves();
       deleteApplicationTableRows();
+      deleteAllStakeholders();
+      deleteAllStakeholderGroups();
+      deleteAllBusinessServices();
```
sjd78, line 81: Seeing other things just straight delete everything and I suppose a developer would never want to run CI locally on an instance where they have test data.  Is this mentioned in any docs?

sjd78 (approved):
Looks ok to me.

I'm not a fan of having tests delete all records in general. Any developer doing that in a dev environment with existing testing data will loose their data. Since it is not a new thing in the tests, and keeping it that way is not a change, I'm ok with letting the PR continue.

PR #3358, opened 2026-06-16
:seedling: Bump react-error-boundary to 6.1.2

sjd78:
- react-error-boundary to 6.1.2
- update ErrorFallback component to use new API types
- ErrorFallback useEffect instead of useRef to handle notification

Follow up to: #3357

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error display and notification behavior so users receive clearer, more reliable feedback when errors occur.
* **Chores**
  * Updated the error-handling library to a newer compatible minor version to keep error boundary handling current.

### DIFF
--- a/client/package.json
+++ b/client/package.json
@@ -49,7 +49,7 @@
     "radash": "^12.1.0",
     "react": "^18.3.1",
     "react-dom": "^18.3.1",
-    "react-error-boundary": "^6.0.0",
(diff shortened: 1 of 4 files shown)


rszwajko (approved):
Looks good! We can lower the number of lint warnings:)
```
20 problems (0 errors, 20 warnings)
```

PR #3368, opened 2026-06-18
:test_tube: Automate user creation from User Management UI

sshveta:
## Summary by CodeRabbit

* **Tests**
  * Added end-to-end coverage for User Management (create, edit, role assignment, and deletion), including verifying the user can’t log in after deletion.
  * Added auth-required E2E suites for Migrator and Architect RBAC operations, plus a task-cancellation permission check.
  * Removed older legacy RBAC spec variants to prevent duplicate/redundant runs.
* **CI / Test Automation**
  * Updated Cypress UI test selection/exclusion logic to rely on auth tagging when auth is not required.
  * Aligned nightly E2E matrix to include auth-required tagging (and skip the non-auth variant).
  * Increased Docker `nofile` limit for CI E2E runs.


.github/actions/run-ui-tests/action.yml
```diff
@@ -133,11 +133,11 @@ runs:
...
 
-        # Exclude rbac tests if auth is not required
+        # Exclude rbac and oidc tests if auth is not required
         if [[ "$AUTH_REQUIRED" != "true" ]]; then
-          echo "Auth is disabled - excluding rbac tests"
-          # Filter out any rbac test files from the spec
-          SPEC=$(echo "$SPEC" | tr ',' '\n' | { grep -v '/rbac/' || true; } | tr '\n' ',' | sed 's/,$//')
+          echo "Auth is disabled - excluding rbac and oidc tests"
+          # Filter out any rbac and userManagement test files from the spec
+          SPEC=$(echo "$SPEC" | tr ',' '\n' | { grep -v -e '/rbac/' -e '/userManagement/' || true; } | tr '\n' ',' | sed 's/,$//')
```
rszwajko: can we add a new tag i.e. `@tierX_authNeeded` ? 
i.e. similar to how we handle secrets 
```
describe(
  ["@tier3", "@tier3_secretsNeeded"],
  "Jira connection filter validations",
```
then we could either filter them out  with `findTierFiles.mjs` or even better use [GH action matrix exclude feature](https://docs.github.com/en/actions/how-tos/write-workflows/choose-what-workflows-do/run-job-variations#excluding-matrix-configurations):
```
exclude:
      - test_tags: '@tierX_authNeeded'
        auth_required: true 

```

cypress/e2e/models/administration/userManagement/user.ts
```diff
@@ -0,0 +1,210 @@
+/*
+Copyright © 2021 the Konveyor Contributors (https://konveyor.io/)
```
rszwajko: 1. In general the `LICECNSE` file in the root folder should take care of that. 
2. If you want to add the comment please correct the year -[ it should be the current year](https://reuse.software/faq/#years-copyright).

Note that `LICENSE`  file is not fully correct i.e. it contains  line `Copyright [yyyy] [name of copyright owner]` which is clearly a template. We could also consider migrating to  [SPDX tags](https://reuse.software/spec-3.3/#comment-headers).
I think it's worth a follow-up discussion.

cypress/e2e/models/administration/userManagement/user.ts
```diff
@@ -0,0 +1,210 @@
...
+      cy.visit(User.fullUrl, { timeout: 35 * SEC }).then((_) => {
+        cy.get("h1", { timeout: 10 * SEC }).should("contain", "Users");
+      });
+      return;
+    }
+
+    cy.url().then(($url) => {
+      if ($url != User.fullUrl) {
+        selectUserPerspective(administration);
+        clickByText(commonView.navLink, userManagement);
```
rszwajko: on each page change we always start with all the nav groups expanded so there is no need to click that group (actually I think it will collapse the sub list)

cypress/e2e/models/administration/userManagement/user.ts
```diff
@@ -0,0 +1,210 @@
...
+        clickByText(userManagementView.usersSubmenuLink, users);
```
rszwajko: why do we need a new selector? I think it should work with regular one i.e. `clickByText(commonView.navLink, users)`

cypress/e2e/views/user-management.view.ts
```diff
@@ -0,0 +1,37 @@
...
+export const loginInput = "#login";
+export const nameInput = "#name";
+export const emailInput = "#email";
+export const passwordInput = "#password";
+export const confirmPasswordInput = "#confirmPassword";
+export const rolesSelectToggle = "#roles-select-toggle-input";
+export const rolesMenuItem = ".pf-v6-c-menu__item-text";
+
+// Buttons
+export const createButton = ".pf-v6-c-modal-box__footer button.pf-m-primary";
```
rszwajko: please add `ouiaId` to create/save buttons and migrate selectors to those data IDs

cypress/e2e/views/user-management.view.ts
```diff
@@ -0,0 +1,37 @@
...
+export const saveButton = ".pf-v6-c-modal-box__footer button.pf-m-primary";
+
+// Actions
```
rszwajko: the selectors below seems unused

cypress/e2e/models/administration/userManagement/user.ts
```diff
@@ -0,0 +1,210 @@
...
+    inputText(userManagementView.passwordInput, this.password);
+  }
+
+  protected fillConfirmPassword() {
+    inputText(userManagementView.confirmPasswordInput, this.password);
+  }
+
+  protected assignRoles() {
+    if (this.roles.length > 0) {
+      click(userManagementView.rolesSelectToggle);
```
rszwajko: Check out how we add source labels  are added in https://github.com/konveyor/tackle2-ui/blob/a76c60df74362d4bf37dd140d4e4f575b4ceee23/cypress/e2e/models/migration/applicationinventory/analysis.ts#L222
This should be the same component as the one used here for roles. Note that selection is scoped to the opened menu (not the entire body).

cypress/e2e/types/types.ts
```diff
@@ -173,6 +173,14 @@ export type UserData = {
   roles?: string[];
 };
 
+export type OidcUserData = {
```
rszwajko, line 176: you should be able to re-use the existing `UserFormValues` type

cypress/e2e/views/user-management.view.ts
```diff
@@ -0,0 +1,37 @@
...
+  "section.pf-v6-c-nav__subnav a.pf-v6-c-nav__link";
+
+// Form inputs
+export const loginInput = "#login";
+export const nameInput = "#name";
+export const emailInput = "#email";
+export const passwordInput = "#password";
+export const confirmPasswordInput = "#confirmPassword";
+export const rolesSelectToggle = "#roles-select-toggle-input";
+export const rolesMenuItem = ".pf-v6-c-menu__item-text";
```
rszwajko: no needed if we move to the same pattern as in `labelsToInclude()`

rszwajko (changes requested):

cypress/e2e/types/types.ts
sshveta: UserFormValues has confirmPassword which we don't need in the model
  and it has roles: User["roles"] (HubRef[]) while we just need role names as strings for simplicity

.github/actions/run-ui-tests/action.yml
rszwajko: IMHO we should be able to remove entire ` if [[ "$AUTH_REQUIRED" != "true" ]]; then` section.
Filtering by tags only has the benefit that we have single source of truth. The extra filtering based on path may break with any refactoring.

cypress/e2e/tests/administration/userManagement/user.test.ts
```diff
@@ -0,0 +1,85 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+/// <reference types="cypress" />
+
+import { getRandomOidcUserData } from "../../../../utils/data_utils";
+import { exists, login, logout, notExists } from "../../../../utils/utils";
+import { User } from "../../../models/administration/userManagement/user";
+
+describe(["@tier0", "@tier0_authNeeded"], "User CRUD operations", () => {
```
rszwajko: `@tier0` (and probably `@tier1`) should be split into mutually exclusive groups similar how `@tier2` and `@tier3` are split:
```
@tier2 === @tier2_A + @tier2_B + @tier2__secretsNeeded
```
If the chunks overlap the `exclude` won't work properly as the tests will be selected  in 2 places:
1. `@tier0` - the whole tier (including @tier0_authNeeded)
2. `@tier0_authNeeded` (only the chunk that requires auth)

If you find splitting too wide change we might just use a dedicated tag i.e. 
```
describe(["@authNeeded"]...
```

In the future  chunks in `@tier2` and `@tier3` may also start to overlap by accident/user error. We might need a test to verify that each test belongs to one chunk.

rszwajko (commented):
Overall looks good - check my comments below for issue with tags.

.github/actions/run-ui-tests/action.yml
rszwajko: @sshveta ^^

sshveta: Yes forgot to remove this

rszwajko (approved):

PR #3403, opened 2026-06-25
:test_tube: Automated token generation from UI 

sshveta:
## Summary by CodeRabbit

* **New Features**
  * Added unique UI automation identifiers for the token creation controls to improve reliable UI targeting (including the primary create button).

* **Quality & Testing**
  * Expanded end-to-end coverage for administrative API token authentication and lifecycle flows, including creation, validation against key endpoints, custom lifespan scenarios, listing verification, and revocation.
  * Added reusable token test helpers and view selectors to support these scenarios.


cypress/e2e/models/administration/userManagement/token.ts
```diff
@@ -0,0 +1,173 @@
...
+  }
+
+  /**
+   * Revoke (delete) a token via the UI.
+   */
+  static revokeTokenByLogin(login: string) {
+    Token.openList();
+
+    // Find the row with the login and click the revoke action
+    cy.contains("td", login)
```
rszwajko: we have nice utility functions for that - the preferred is  `performRowActionBySelector()`. If adding data test IDs is not feasible you can try other versions like `performRowAction`

cypress/e2e/tests/administration/userManagement/token.test.ts
```diff
@@ -0,0 +1,170 @@
...
+        );
+        expect(createdToken).to.exist;
+      });
+    });
+  });
+
+  it("Should reject invalid token", function () {
+    const invalidToken = "invalid-token-12345";
+
+    Token.testTokenViaAPI(invalidToken).then((response) => {
```
rszwajko, line 75: The test is not verifying the UI.  I assume we have dedicated API tests to verify the backend.
IMHO we should use the API sparingly i.e. to improve performance or for extra verification.

rszwajko (commented):
Looks good! Small comments below.

cypress/e2e/tests/administration/userManagement/token.test.ts
sshveta: UI only allows creating token but to check if token can be used to login I have used API .
If these tests are covered somewhere else I can remove it from here @rszwajko .

rszwajko: OK. Just saying that we are not testing UI here. We can keep the test as it's a small one.

rszwajko (approved):

PR #3421, opened 2026-06-29
:test_tube: Automate role OIDC tests

sshveta:
## Summary by CodeRabbit

* **New Features**
  * Added end-to-end role management coverage for creating, editing, duplicating, and deleting roles, including permission UI validations.
  * Added UI selectors and deterministic role save/create identifiers to improve reliable automation.
  * Introduced a helper to read trimmed table cell text by row identifier and column name.
* **Bug Fixes**
  * Improved token cleanup to handle multiple API response formats and safely delete only matching tokens with valid IDs.
  * Tightened kebab-menu row lookups to avoid selecting unintended text outside the target table.
* **Tests**
  * Added a dedicated roles test suite and updated the token authentication assertion to expect the correct response property.

### DIFF
--- a/cypress/e2e/models/administration/userManagement/role.ts
+++ b/cypress/e2e/models/administration/userManagement/role.ts
@@ -13,7 +13,7 @@ import * as userManagementView from "../../../views/user-management.view";
 
 export interface RoleData {
   name: string;
-  permissions?: string[]; // Permission scopes like "addon.create", "application.delete"
+  permissions?: string[];
 }
 
 export class Role {
@@ -55,11 +55,9 @@ export class Role {
   protected assignPermissions() {
     if (this.permissions.length > 0) {
       this.permissions.forEach((permissionScope) => {
-        // Find the permission in the available (left) list
         cy.contains(userManagementView.permissionItem, permissionScope).click();
       });
 
-      // Click "Add selected" button (right arrow)
       cy.get(userManagementView.addSelectedButton).click();
     }
   }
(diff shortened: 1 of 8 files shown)


cypress/e2e/views/user-management.view.ts
```diff
@@ -33,3 +33,13 @@ export const createTokenButton =
 export const tokenCreateButton =
   '[data-ouia-component-id="token-create-button"]';
 export const lifespanInput = "#lifespan";
+
+// Role management
+export const roleNameInput = "#name";
+export const rolesMenuItem = '[role="menuitem"]';
+export const createRoleButton = 'button[aria-label="Create role"]';
+export const roleDialog = 'div[role="dialog"]';
+export const permissionItem = ".pf-v6-c-dual-list-selector__item-text";
```
rszwajko, line 42: can we avoid using CSS class as selector?

cypress/e2e/models/administration/userManagement/role.ts
```diff
@@ -0,0 +1,179 @@
...
+      clickByText(button, "Cancel");
+      notExists(newRoleName);
+      return null;
+    } else {
+      cy.get(userManagementView.roleCreateButton).click();
+
+      cy.get(userManagementView.roleDialog).should("not.exist");
+      exists(newRoleName);
+
+      return new Role({ name: newRoleName, permissions: [] });
```
rszwajko: the permissions are empty - in duplicate scenario I would expect the permissions from the cloned role.

cypress/e2e/views/user-management.view.ts
sshveta: @rszwajko I checked  PatternFly DualListSelectorListItem doesn't support OUIA IDs . For now I will have to use this only .

rszwajko (approved):

PR #3447, opened 2026-07-02
:bug: Use revoke endpoint to revoke tokens

### LINKED ISSUE #3428: Use tokens/:id/revoke
When revoking a token, instead of DELETE /auth/tokens/:id
Use: POST /auth/tokens/:id/revoke.

This will delete the GRANT as well.  Else, for access tokens, the refresh token is still valid and can be renewed to obtain another access token.

rszwajko:
Fixes: #3428 

## Summary by CodeRabbit

* **New Features**
  * Added a dedicated token revocation action, with the UI now calling the new revoke flow directly.
  * Introduced updated token mutation handling so revocations refresh token data after success.

* **Bug Fixes**
  * Token removal behavior now clearly revokes the token as part of the action, aligning the interface and backend flow.

### DIFF
--- a/client/src/app/api/rest/tokens.ts
+++ b/client/src/app/api/rest/tokens.ts
@@ -22,12 +22,10 @@ export const createToken = (
 ): Promise<PersonalAccessToken> =>
   axios.post<PersonalAccessToken>(TOKENS, token).then((r) => r.data);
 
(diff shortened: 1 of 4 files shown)


sjd78 (approved):
LGTM

PR #3457, opened 2026-07-08
:seedling: Update Dockerfile labels

### LINKED ISSUE #3151: Revisit the Dockerfile labels
The LABELs used in our `Dockerfile` have not been updated in a few years.  Revisit the labels and update the contents, add new labels and remove old labels as necessary or appropriate for the current ecosystem (k8s, OpenShift, OLM, CNCF, etc)

sjd78:
Resolves: #3151

Update Dockerfile labels to:
  - Add new and useful labels (OCI standard)
  - Remove deprecated labels (OpenShift 3 specific)
  - Update values of existing labels to current state of the project

Note, the `image-build.yaml` workflow calls the shared `konveyor/release-tools/.github/workflows/build-push-images.yaml@main` workflow. This workflow will generate OCI labels based on the checked out repository used to build the image. If any labels that overlap between the Dockerfile and input at build time, the build time label values will be used.

Assisted-by: Cursor:claude-opus-4.6-high

## Summary by CodeRabbit

* **Documentation**
  * Updated container metadata with clearer product descriptions, links, licensing, and vendor information.
  * Improved container usage instructions, including port mapping and image tag examples.
  * Simplified and standardized platform labeling for better container cataloging.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -54,7 +54,6 @@ LABEL name="konveyor/tackle2-ui" \
 COPY --from=builder /opt/app-root/src/dist /opt/app-root/dist/
 
 ENV DEBUG=1
-ENV NODE_EXTRA_CA_CERTS=/opt/app-root/src/ca.crt
 
 WORKDIR /opt/app-root/dist
 ENTRYPOINT ["./entrypoint.sh"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -32,24 +32,29 @@ RUN microdnf -y install tar procps-ng && microdnf clean all
 USER 1001
 
 LABEL name="konveyor/tackle2-ui" \
-      description="Konveyor for Tackle - User Interface" \
-      help="For more information visit https://konveyor.io" \
-      license="Apache License 2.0" \
-      maintainer="gdubreui@redhat.com,ibolton@redhat.com" \
-      summary="Konveyor for Tackle - User Interface" \
+      description="Konveyor - User Interface" \
+      summary="Konveyor UI provides the web-based frontend for managing application modernization workflows" \
       url="https://quay.io/konveyor/tackle2-ui" \
-      usage="podman run -p 80 -v konveyor/tackle2-ui:latest" \
+      help="For more information visit https://konveyor.io" \
+      license="Apache-2.0" \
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -32,24 +32,29 @@ RUN microdnf -y install tar procps-ng && microdnf clean all
...
+      org.opencontainers.image.url="https://konveyor.io" \
+      org.opencontainers.image.source="https://github.com/konveyor/tackle2-ui" \
+      org.opencontainers.image.documentation="https://konveyor.io/docs" \
+      org.opencontainers.image.licenses="Apache-2.0" \
+      org.opencontainers.image.vendor="Konveyor"
 
 COPY --from=builder /opt/app-root/src/dist /opt/app-root/dist/
 
 ENV DEBUG=1
+ENV NODE_EXTRA_CA_CERTS=/opt/app-root/src/ca.crt
```
rszwajko: why we need it at build time?

entrypoint.sh
```diff
@@ -24,6 +24,7 @@ if [[ $AUTH_REQUIRED != "false" ]]; then
 fi
 
 # Copy the Kube API and service CA bundle to /opt/app-root/src/ca.crt if they exist
+NODE_EXTRA_CA_CERTS=${NODE_EXTRA_CA_CERTS:-/opt/app-root/src/ca.crt}
```
rszwajko: I'd rather skip the default - just pass the env variable as it is.

entrypoint.sh
rszwajko, line 30: 1. NODE_EXTRA_CA_CERTS can be provided as read-only volume mount.
2. why we need to copy the certs?

Dockerfile
sjd78: Dropped

entrypoint.sh
sjd78: Dropped

entrypoint.sh
sjd78: Dropped. I will create a new PR with some changes to the handling of NODE_EXTRA_CA_CERTS.  There are extra considerations, and better ways to handle it.

rszwajko (approved):

PR #3474, opened 2026-07-15
:seedling: Update entrypoint.sh ca-certs handling

sjd78:
Update the entrypoint script to build up a ca-certs file to contain certs potentially available in the running container and make it available to the nodejs process.

This is a workaround for the nodejs process not using the system trust store by default. It will use the certs in the NODE_EXTRA_CA_CERTS pointed file for the extra certs.

The NODE_EXTRA_CA_CERTS environment variable is not required to be set, but if the file needs to be stored in a different location in the running container, the environment variable should be set to the desired location.

## Summary by CodeRabbit

* **New Features**
  * Improved Node.js TLS certificate handling in container deployments by combining available trusted CA certificates automatically.
  * Added support for custom `.crt` and `.pem` certificates mounted into the container.
  * Added support for common Kubernetes and OpenShift certificate sources.

* **Documentation**
  * Added configuration guidance and deployment examples for custom CA certificates.
  * Documented operator follow-up steps for the updated certificate injection workflow.

### DIFF
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -23,11 +23,9 @@ if [[ $AUTH_REQUIRED != "false" ]]; then
   fi
 fi
 
-# Build a combined CA bundle for Node.js, which does not use the system trust store.
-# NODE_EXTRA_CA_CERTS adds CAs on top of Node's compiled-in Mozilla root CAs.
-# If NODE_EXTRA_CA_CERTS is already set to an existing file, it is treated as
-# an additional input source rather than the output path, so read-only mounts
-# (Secrets, ConfigMaps) are safe to use.
+# Nodejs does not use the system trust store by default. We build a combined CA
+# bundle from known sources and set NODE_EXTRA_CA_CERTS to point at it. This adds
+# CAs on top of Node's compiled-in Mozilla root CAs.
 ca_bundle="/tmp/node-ca-bundle.pem"
 
 ca_sources=(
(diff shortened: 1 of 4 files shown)


entrypoint.sh
```diff
@@ -23,21 +23,31 @@ if [[ $AUTH_REQUIRED != "false" ]]; then
...
+  # Kubernetes API server CA (standard SA volume mount)
+  /var/run/secrets/kubernetes.io/serviceaccount/ca.crt
+  # OpenShift service-serving CA (requires operator-managed ConfigMap mount)
+  /var/run/secrets/kubernetes.io/serviceaccount/service-ca.crt
+  # Cluster-injected custom/proxy CAs via RHEL trust store
+  /etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem
+)
+
+found=false
+touch "${NODE_EXTRA_CA_CERTS}"
```
rszwajko: NODE_EXTRA_CA_CERTS may point to a read-only path.
If we want to merge the certs then lets use a path that is under our control.
I'm thinking about flow like this:
```
KONVEYOR_CERTS=/opt/app-root/src/ca.crt
ca_sources=(
   ...
   ${NODE_EXTRA_CA_CERTS}
)

or src in "${ca_sources[@]}"; do
  if [ -f "$src" ]; then
    cat "$src" >> "${KONVEYOR_CERTS}"
  fi
done

export NODE_EXTRA_CA_CERTS = KONVEYOR_CERTS;
```

sjd78: I refactored the handling to be more generic and allow for mounting of certs to a ca-certs.d and have things included from there.  It'll need an operator change to go with it, so now this PR is draft and I'll deal with it full after 0.10.0 GA

sjd78:
This PR is not needed considering #3580.

PR #3535, opened 2026-08-10
:test_tube: Increase Node.js max HTTP header size for OIDC cookie support (#v7)

sshveta:
cypress-io/github-action@v7 runs on Node.js 20, which enforces a
  stricter default maxHeaderSize (16KB). OIDC/Keycloak session cookies
  exceed this limit, causing 400 Request Header Or Cookie Too Large
  errors in auth-enabled nightly runs.


jortel (approved):
LGTM

Issue #3302, opened 2026-06-03
Include offline_access scope to OIDC client request.


jortel:
### Offline Access Scope (`offline_access`)

**Request**: Add support for the `offline_access` scope.

#### Description
The `offline_access` scope is used in **OpenID Connect** to request a Refresh Token that can be used to obtain new Access Tokens even when the user is not present (offline / long-lived access).  Without it, the login page will pop up when the access-token expires (Every 5 min) instead of refreshing.

#### Specification
- Defined in: **[OpenID Connect Core 1.0](https://openid.net/specs/openid-connect-core-1_0.html#OfflineAccess)** (Section 12.2)
- Not defined in any core OAuth 2.0 RFC (e.g. RFC 6749), but is a standard OIDC extension.

#### Expected Behavior
When `offline_access` is requested (along with `openid`), the authorization server should return a `refresh_token` that remains valid beyond the user's current session.

**Note**: Keycloak is less strict about this.

---

**Use Case**: Required for mobile/desktop apps, background sync, service integrations, etc.

```
diff --git a/client/src/app/auth/userManager.ts b/client/src/app/auth/userManager.ts
index bd036804d..77af2988c 100644
--- a/client/src/app/auth/userManager.ts
+++ b/client/src/app/auth/userManager.ts
@@ -34,7 +34,7 @@ const settings: UserManagerSettings = {
   redirect_uri: window.location.origin,
   post_logout_redirect_uri: window.location.origin,
   response_type: "code", // Authorization Code + PKCE (oidc-client-ts default for SPAs)
-  scope: "openid profile email",
+  scope: "openid profile email offline_access",
   userStore: new WebStorageStateStore({ store: window.sessionStorage }),
 
   // Automatically try a silent renew 60 s before expiry.
```

jortel:
The hub can issue the refresh token without the explicit scope.

PR #3545, opened 2026-08-20
:bug: Request offline_access so the builtin OP issues a refresh token

ibolton336:
UI sessions against a hub with auth required die ~5 minutes after login: every hub call starts returning 401 and the app eventually bounces to the login form. The chain, verified live:

- Hub access tokens live 300 s (`Settings.Token.Lifespan` default).
- The UI requests `scope: "openid profile email"`. The builtin OP (zitadel/oidc) issues a refresh token only when `offline_access` is requested — so none is issued.
- `automaticSilentRenew` then falls back to hidden-iframe renew, which cannot work here: the OP answers `prompt=none` with a login redirect (not `error=login_required`), and the SPA registers no silent-callback route. Every renew times out, and the expired bearer keeps being sent.

The fix is one line: also request `offline_access`, so renewal rides the refresh grant.

This is the ask from #3302, which was closed as unnecessary because the hub can issue refresh tokens without the explicit scope — with Keycloak that's true, but the builtin OP enforces the OIDC-spec requirement (live-verified: the token response carries no `refresh_token` without the scope). The hub's own Go binding has always requested `openid profile email offline_access` (`shared/binding/auth/oidc.go`), so any hub that works with addons already accepts it.

### DIFF
--- a/client/src/app/auth/oidc/userManager.ts
+++ b/client/src/app/auth/oidc/userManager.ts
@@ -11,7 +11,7 @@ export const userManager = new UserManager({
   redirect_uri: window.location.origin,
   post_logout_redirect_uri: window.location.origin,
   response_type: "code", // Authorization Code + PKCE (oidc-client-ts default for SPAs)
-  scope: "openid profile email",
+  scope: "openid profile email offline_access",
   userStore: new WebStorageStateStore({ store: window.sessionStorage }), // sessionStorage == new login for each tab
 
   // Automatically try a silent renew before expiry.


ibolton336:
`@tier3_C` on attempt 1 failed in `controls/stakeholdergroups/filter.test.ts › Member filter validations` — unrelated to this change, and a latent defect in the spec:

- The e2e matrix runs with `feature_auth_required: false` ([ci-e2e-tests.yml:137](https://github.com/konveyor/tackle2-ui/blob/main/.github/workflows/ci-e2e-tests.yml#L137)). In that mode `AuthProvider` mounts `NoAuthStrategy` and `axios-auth.ts` never installs its interceptors, so the `scope` string this PR edits is never sent anywhere.
- The test only asserts group 1's *member name* when the two faker names share group 0's first three characters (here `Eli`: *Elizabeth Hammes* / *Elijah Lowe*). Member names are rendered only inside an expanded row, the table is single-expand, and row 1 is never expanded — so that branch cannot pass (`exists()` then does its `cy.visit` recovery, which is the collapsed, filter-less screenshot). ≈0.7 % of runs with `faker@5.5.3`; first sighting in the last 200 CI runs. One-line fix for the spec: assert the row (`exists(stakeholderGroupsList[1].name)`) instead of the member name — tracked in #3551.
- Attempt 2 (`--failed` rerun) could not pass on principle: the e2e stack pulls `ttl.sh/tackle2-ui-<sha>:2h`, which had expired, so `tackle-ui` never started. Full rerun ([attempt 3](https://github.com/konveyor/tackle2-ui/actions/runs/32383561361/attempts/3), same head): all nine tiers green, `@tier3_C` included.

sjd78:
Why is the needed? If the web UI is open it'll keep the token alive / renewed just based on the background data fetching. Is in not behaving that way?

ibolton336:
Closing: this is a no-op against current upstream. The builtin OIDC provider force-appends `offline_access` to every auth request in `CreateAuthRequest` (`internal/auth/storage.go`, builtin-OP PR konveyor/tackle2-hub#1042, commit `623f9d5b`), so the hub issues a refresh token regardless of the scope the UI requests — verified live on a current hub (pre-change UI gets a working refresh token). So @jortel's rationale on #3302 holds for the merged hub. May revisit purely as spec-correctness hardening (request `offline_access` explicitly rather than rely on the hub's forced-append), but nothing is broken today.

Issue #3526, opened 2026-08-07
Continue-work affordance on a completed run


ibolton336:
A "Continue work" affordance on a completed run launches a continuation run with branch and context prefilled. Behind the #3522 feature gate.

Cut-line note: high demo value, not load-bearing — first to slip if the sprint math fails.

Done when:
- A completed run offers "Continue work" and the resulting run targets the same branch with context prefilled.

Depends on: https://github.com/konveyor/agentic-controller/issues/128 (continuation runs) and #3524 (the run-creation form).

Part of the Konveyor agentic dev-preview (v0.11.0) scope.

Issue #3524, opened 2026-08-07
Run-creation forms: namespaced workflow/agent params and mode selection


ibolton336:
Generate the run-creation form from declared params: a workflow section and an agent section, mirroring ADR 0009's namespacing. Execution knobs per ADR 0011: runs expose mode selection only, validated against the Agent's `allowedModes`; limits (`maxTurns`/`maxCost`) belong on the Agent/workflow authoring forms — runs cannot override them in dev preview. Feeds both the single-run and bulk-run flows. Behind the #3522 feature gate.

Done when:
- The form renders from declarations, namespaces workflow vs agent params, and offers only in-set modes.

Depends on: https://github.com/konveyor/agentic-controller/issues/115 (the CRD fields) and the current shim/hub serving the declarations.

Part of the Konveyor agentic dev-preview (v0.11.0) scope.

Issue #3523, opened 2026-08-07
Runs pages: server-side application/archetype filter chips


ibolton336:
Add server-side application/archetype filter chips to the agent-runs and workflow-runs list pages, backed by the Hub's label-selector run filters and matching the hub filter conventions used elsewhere in the console.

Done when:
- Both runs lists filter by application (and archetype, once available) server-side via chips consistent with existing console filters.

Depends on: hub run-list filters (https://github.com/konveyor/tackle2-hub/issues/1112) and https://github.com/konveyor/agentic-controller/issues/107 (stage runs must inherit labels, or filtered lists miss every stage run).

Part of the Konveyor agentic dev-preview (v0.11.0) scope.

Issue #3504, opened 2026-07-28
Agent runs UI: agents, runs, playbooks, and interactive ACP session panel (agentic-controller integration)


ibolton336:
## Context

konveyor/agentic-controller#22 defines the client contract for agent runs, and a 2026-07-27 design sync settled placement: the Hub exposes the REST surface (R1) and a WebSocket proxy to the run pod (R2) under an `/agent/…` route namespace. Full contract and responsibility split: [consolidated contract comment](https://github.com/konveyor/agentic-controller/issues/22#issuecomment-4905804098).

This issue tracks the **tackle2-ui** side: a new "agent runs" section plus an interactive session panel. The Hub absorbs all Kubernetes access, secret/key handling, and credential resolution — the browser's only credential is the hub token, so the UI work is purely presentational. A working reference implementation (plain React, pre-PatternFly) exists in [ibolton336/agentcontroller-client](https://github.com/ibolton336/agentcontroller-client) (`ui/src/components`), verified end-to-end against the real controller with a goose+Bedrock agent.

## Depends on

- Hub R1: konveyor/agentic-controller#72 (Hub integration — the `/hub/agent/…` route table), implemented by konveyor/tackle2-hub#1112
- Hub R2 (WS proxy): tracked as a checkbox inside konveyor/agentic-controller#72; the write-stream authorization design is konveyor/tackle2-hub#1116 — blocks only the session panel (item 6)

## Current work breakdown (dev preview / v0.11.0)

Scoped issues carved out of this umbrella: #3520, #3521, #3522, #3523, #3524, #3525, #3526, #3527. Working branch: [`ibolton336/tackle2-ui` `feature/agent-runs`](https://github.com/ibolton336/tackle2-ui/tree/feature/agent-runs).

## Scope

- [ ] **1. REST layer** — types (`Agent`, `AgentRun`, `AgentPlaybook`, `AgentPlaybookRun`; read-only `LLMProvider`/`SkillCard`/`SkillCollection`), axios endpoints for the Hub's `/agent/…` routes, react-query hooks. The Hub is fire-and-forget (no reconciliation, no watch/SSE), so run queries poll via `refetchInterval` to track phase transitions.
- [ ] **2. Routes, navigation, RBAC** — new sidebar section and route registration, gated by the standard scopes the Hub R1 issue defines (architect/migrator scoping). Existing tackle2-ui patterns throughout.
[...]

PR #3546, opened 2026-08-20
:sparkles: Agentic console: agent runs, workflows, and interactive session panel (gated dev preview)

### LINKED ISSUE #3520: Start agent workflow runs for selected applications from the application inventory
## Problem

Agent runs and agent workflow runs can only be started from the Agent runs section, where the user picks an application out of a dropdown. The natural entry point is the opposite direction: select applications in the inventory and run something against them — the way analysis, credential management, and configuration retrieval already work.

## Proposal

Add a bulk item to the applications table toolbar kebab (`applications-table.tsx`, `toolbarKebabItems`), next to "Retrieve configurations" — whose gating (`selectedRows.length < 1 || !selectedRows.some(...)`) is already the eligibility shape needed here.

It opens a new modal:

- Choose the agent workflow (or agent) to run
- **Target branch**, templated per application (e.g. `agentic/{app}-{date}`) — a single literal branch name will collide with *some* applications' source branch, which the create path rejects
- **Eligibility summary** — "10 of 12 applications eligible; 2 excluded (no repository URL)" — rather than a blocking form error
- Create, then report per-application success and failure

## Fan-out

There is no server-side fan-out primitive for agentic runs. Bulk analysis works because the Hub has taskgroups: one POST, server fans out to N applications. `POST /api/agentruns` and `POST /api/agentworkflowruns` each create exactly one run against exactly one `applicationRef`.

So this ships as client-side fan-out — `Promise.allSettled` over the selection — following the precedent already in this same table: `useStartFetchApplicationManifest` returns a `{ success, failure }` split for exactly this reason. Keep it behind a single mutation hook so a future server-side `applicationRefs[]` can replace it without touching the modal. Shape discussion: ibolton336/agentcontroller-client#4.

Worth noting the already-commented-out bulk "Generate assets" item in the same kebab — *"Add this back when we can handle the generate operation in bulk"* — hit the same wall.

## Notes

- **Application ids line up for free.** The agentic inventory reports `id: String(hubApplication.id)`, so `applicationRef` is just `String(application.id)` off an inventory row. No lookup, no mapping.
- **N applications means N runs and N sandbox pods.** The confirm step should show the count plainly; concurrency limits are part of the fan-out discussion above.
- Once runs can be started from here, users will expect to see them here. Status feedback is tracked separately (agent runs drawer tab) and should land close behind this, or the bulk action is fire-and-forget into a different nav section.

## Dependencies

- **#3522** — the agentic console is not feature-gated today. This item appears in the toolbar kebab of the application inventory, which every deployment uses, so it must be hidden when there is no agentic backend configured.
- **ibolton336/agentcontroller-client#5** — the create-run modals currently own their own application picker and must accept applications as props before they can be driven from here.

Part of #3504.

### LINKED ISSUE #3521: Application detail: agent runs tab and per-application run action
## Problem

Two gaps, both about the single-application case:

1. **Starting a run for one application** still means leaving the inventory, opening the Agent runs page, and finding that application in a dropdown.
2. **Nothing shows what an application has run.** Agent runs live in a different backend from Hub tasks, so nothing decorates inventory rows the way `useDecoratedApplications` / `tasks.currentAnalyzer` does for analysis. After starting runs from the inventory, the user gets a toast and then has to navigate to a different nav section to find out what happened.

## Proposal

- **Row kebab action** in `applications-table.tsx` — pre-select the application and open the create modal with its picker hidden. Nearly free once the modal refactor lands, and probably the more common flow than bulk.
- **"Agent runs" tab in the application detail drawer**, joining Details / Tags / Reports / Reviews / Tasks / Platform: the runs for this application with phase, timestamps, and a link through to run detail.

## Filtering runs by application

The application link currently rides only in the run's `spec.env.APP_ID` — no label, no annotation. `runHubCoordinates()` already decodes it, so the tab can be built today by fetching all runs and filtering client-side.

That is acceptable at current scale but does not hold at inventory scale. ibolton336/agentcontroller-client#3 covers stamping the application id as a label so this becomes a real selector; the tab should move to it once available.

## Out of scope

A per-row status column or decoration in the inventory table itself. Worth doing eventually — it is what makes runs visible without opening the drawer — but it needs the label-selector work first, and probably a decorator alongside `useDecoratedApplications`.

## Dependencies

- **#3522** — the agentic console is not feature-gated today. Both surfaces here (a row action and a drawer tab) live in the application inventory, which every deployment uses, so they must be hidden when there is no agentic backend configured.
- **ibolton336/agentcontroller-client#5** — the create-run modals must accept an application as a prop before the row action can drive them.

Part of #3504.

### LINKED ISSUE #3522: Feature-gate the agentic console (agent runs, agents, skills, workflows)
## Problem

The agentic console — the Agent runs / Agents / Skills / Workflows nav section being built for #3504 — is not gated. `SidebarApp.tsx` renders the `NavExpandable` unconditionally, the routes are registered unconditionally in `Routes.tsx`, and `ClientEnv` (`common/src/env-types.ts`) carries no agentic entry.

The only thing between a stock deployment and a broken nav section today is `AGENTIC_SHIM_URL`, the **server-side** proxy target for `/agentic`. When it is unset the proxy has no target and every agentic call fails — so the feature is effectively off, but it fails *loudly* rather than hiding. A user on a deployment with no agentic backend still sees four nav items that error on click.

## Proposal

- Add an agentic flag to `ClientEnv` (e.g. `AGENTIC_ENABLED: "true" | "false"`), defaulting off, serialized through `window._env` alongside the existing flags (`RWX_SUPPORTED`, `AUTH_REQUIRED`, …).
- Gate the agentic `NavExpandable` in `SidebarApp.tsx` and the agentic route registrations in `Routes.tsx` on it.
- Consider defaulting it from whether `AGENTIC_SHIM_URL` is configured server-side, so an operator that deploys the shim gets the console without having to set a second knob — and a deployment without one never renders the section.

## Why now

Any integration of agentic features into *existing shipped screens* depends on this. #3520 adds a bulk item to the application inventory toolbar, and #3521 adds a row action and a tab to the application detail drawer. Those surfaces are used by every deployment, agentic or not, and must not show agentic affordances when there is no backend behind them.

Blocks #3520 and #3521. Part of #3504.

### LINKED ISSUE #3525: Run a standalone agent (AgentRun) from the console
From the scope call: "run agent, not just run workflow." Create and launch a single standalone AgentRun from the console using the generated param form; the runs list/detail pages already exist, and the shim already accepts AgentRun creates. Behind the #3522 feature gate.

Done when:
- A standalone agent can be selected, parameterized, and run from the console, landing on the existing run detail view.

Depends on: #3524 (the run-creation form).

Part of the Konveyor agentic dev-preview (v0.11.0) scope.

### LINKED ISSUE #3527: HITL read-only run view; steer behind a default-off feature flag
The ChatPanel live run view ships read-only for dev preview: status frames, plan ladder, and usage from the merged ACP tee (https://github.com/konveyor/agentic-controller/pull/96). Steer UI exists only behind a default-off feature flag wired to the backend's `HARNESS_HITL_STEER` kill switch; write-to-stream stays off until the Hub authorization design lands. Behind the #3522 feature gate.

Done when:
- The live view renders read-only with the flag off.
- Flipping the flag (dev only) exposes steer end to end.

Depends on: the merged backend (PR 96 above); the steer authorization design is deliberately post-dev-preview.

Part of the Konveyor agentic dev-preview (v0.11.0) scope.

ibolton336:
The agentic console from #3504 as a single PR: a new gated section of the UI for launching and watching agentic-controller runs through the hub — agent runs, agent workflows, agents/skills/workflows pages (skill authoring included: inline, image and git-sourced skill cards with client-side frontmatter validation, skill collections, and detail drawers), run creation from the application inventory, and a live interactive session panel (ACP over the hub's WebSocket proxy) with a read-only HITL posture. Skills are also visible run-side: what an agent brings at launch, a Skills group on run detail, and `load_skill` tool calls rendered as "Loaded skill …" rows in the transcript.

**Everything is dark by default.** The console renders only with `AGENTIC_ENABLED=true` (deployment opt-in in `serverConfig.js`; functional only against a hub serving the agentic endpoints — konveyor/tackle2-hub#1119). Free-text steering of a live run is additionally gated by `AGENTIC_STEER_ENABLED` (default off — the dev-preview posture is read-only: watch the transcript, answer the agent's questions, don't inject instructions mid-run). With the flags off, nothing changes for existing deployments: no routes, no sidebar entry, no requests.

**Shape of the diff:** ~+14k lines, almost purely additive — 48 new files under `client/src/app` (37 pages, 6 query modules, 3 api, 2 utils) plus new locale keys and one new dependency (`@patternfly/chatbot`). Pre-existing files are touched only lightly; the spots worth real review:

- `server/src/index.js`, `server/src/proxies.js` — the hub proxy gains `ws: true` and an upgrade handler scoped to `/hub/agentic/` paths (ACP WebSocket; the socket itself is authorized by the hub's one-time-nonce two-step, dialed from `api/agentic/acp.ts`)
- `client/src/app/Routes.tsx` + `layout/SidebarApp` — gated route list and sidebar group
- `applications-table` / `application-detail-drawer` — run-from-inventory kickoff and a per-application runs deep link

**Verified:** end-to-end on a live cluster (real goose+Bedrock agent behind a hub with auth required), and against a scripted mock of the hub contract (boot races, socket drops, steer/elicitation flows); jest suite green; zero new lint warnings (`eslint --max-warnings=20` — main already sits at exactly 20); tsc clean. The authoring surfaces were additionally driven end-to-end against a real hub with auth on (disposable minikube rig running the hub agent-endpoints build): skill card create/edit/delete, agent create/edit/delete (including nested gateway/skill refs and a required param), and two-stage workflow create/delete all round-trip to the CRs exactly as entered, with list invalidation after every mutation and zero browser console errors. The skills rework that followed (inline/image/git card authoring with validation, both collection modes, detail drawers, designer summary, launch-time "Skills this agent brings", run-detail Skills group, `load_skill` transcript rows over a live ACP socket) was verified against the scripted hub mock with every request body logged and checked against the expected CR shape; that live-hub pass predates the rework. `api/agentic/contract.ts` tracks the skill-card shape of konveyor/agentic-controller#157 (still open) and will be re-synced if it moves before merge.

Closes #3520, closes #3521, closes #3522, closes #3525, closes #3527.
Advances #3523 (client-side application/workflow filter chips with URL persistence; server-side filtering follows the hub's `?application=` support), #3524 (typed param forms + launch preflight; namespaced sections and mode selection to follow), and #3526 (not started). Umbrella: #3504.


client/config/rspack.dev.mts
```diff
@@ -32,6 +32,12 @@ const devClientEnv = (env: ClientEnv = process.env as unknown as ClientEnv) =>
       UI_INGRESS_PROXY_BODY_SIZE: env.UI_INGRESS_PROXY_BODY_SIZE ?? "500m",
       RWX_SUPPORTED: env.RWX_SUPPORTED ?? "true",
       AUTH_REQUIRED: env.AUTH_REQUIRED ?? "false",
+      // On by default in dev — the standalone dev server has no proxy config
+      // to derive from, and dev is where the console is being built.
+      AGENTIC_ENABLED: env.AGENTIC_ENABLED ?? "true",
```
rszwajko: I'd rather use `false` to minimize impact on other use cases. Otherwise in all places we would need to explicitly disable agentic console.

client/src/app/pages/applications/applications-table/applications-table.tsx
```diff
@@ -799,6 +817,21 @@ export const ApplicationsTable: FC = () => {
...
         ),
+        // Only offered where an agentic backend is configured — see
+        // AGENTIC_ENABLED. The harness clones from each application's Hub
+        // record, so a selection with no repository anywhere can't run.
+        isAgenticEnabled && agentWorkflowRunAccess && (
+          <DropdownItem
+            key="run-agent-workflow-bulk"
+            isDisabled={
+              selectedRows.length < 1 ||
+              !selectedRows.some((app) => app.repository?.url)
```
rszwajko: IMHO we should enable the action if all apps are valid targets

client/src/app/pages/applications/applications-table/applications-table.tsx
```diff
@@ -1251,6 +1284,16 @@ export const ApplicationsTable: FC = () => {
                                       application,
                                     ]),
                                 },
+                              // The harness clones from the Hub record, so
+                              // an application with no repository can't run.
+                              isAgenticEnabled &&
+                                agentWorkflowRunAccess &&
+                                !!application.repository?.url && {
```
rszwajko: I'd rather keep access checks separate from app-based validations. In this case I would use similar approach as for `"actions.retrieveConfigurations"` (see below):
1. hide if insufficient permissions 
2. disable if no repository

client/src/app/pages/applications/applications-table/applications-table.tsx
```diff
@@ -1341,6 +1384,53 @@ export const ApplicationsTable: FC = () => {
           }}
         />
       )}
+      {runForApplication && (
+        <CreateWorkflowRunModal
+          application={runForApplication}
+          onClose={() => setRunForApplication(null)}
+          onCreated={(runName) => {
+            setRunForApplication(null);
+            history.push(formatPath(Paths.workflowRunDetails, { runName }));
```
rszwajko: the workflow details might not yet exist at this point - it would be safer to use the same approach as with bulk actions - open the list page with filters (`onStarted`).

rszwajko (commented):
The shared parts look OK. Ideally when disabled the agentic console should have no impact on the rest of the product. Currently it's enabled in all scenarios. I think that we should change the defaults for  AGENTIC_ENABLED to false. Check also my other comments below.

I haven't reviewed the parts ported from editor extension or general the parts specific to agentic console. Until it's PoC behind feature flag we can keep it as-it-is. However at some point that code will need some refactorings,

ibolton336:
@rszwajko thanks for the review — all four points addressed in `fe30cf4d`:

- **Default agentic off in dev** (`rspack.dev.mts`): `AGENTIC_ENABLED` and `AGENTIC_STEER_ENABLED` now default to `false`, matching `serverConfig.js`. Local console work opts in explicitly. The console has no impact on other use cases unless enabled.
- **Bulk run enablement** (`applications-table.tsx` ~828): switched `some` → `every`, so the bulk action is only enabled when every selected application is a valid target (has a repository).
- **Separate access checks from app validation** (~1291): now follows the `retrieveConfigurations` pattern — hidden without the feature flag / run access, shown-but-`isDisabled` when the application has no repository.
- **Post-create navigation** (~1393): `onCreated` now lands on the runs list pre-filtered to the application (via `workflowRunsPath`) instead of pushing to the run detail page, since the run object may not be readable yet (mirrors the bulk `onStarted` approach).

CI is green across all tiers. Re-requesting your review.

sjd78 (approved):
Structure looks well contained.  Not sure the server ws upgrade handler is really needed, but it is ok there.  The server's different proxy backend should handle ws connections much more consistently now

sjd78:
FWIW, the large components are not testable as they are written.  Pulling complex state out into a hook and having the react part just drive the display based on data from the hook would let us have unit test on the hook and component tests on the render.

rszwajko (approved):

PR #3580, opened 2026-09-08
[experimental] :sparkles: Replace express server/proxy with Caddy

sjd78:
- Drop the server module
- Use Caddy to replace the server module as the file/proxy server
- Add entrypoint go app to replace entrypoint.sh that cannot run in the hardened image
- Move the env handling to the client app
  - In prod, the `/.env` endpoint is used to inject the env
  - In dev, the `ClientEnvInjectPlugin` injexts the env
  - Cypress tests now check `/.env` instead of parsing the index.html
- `npm run start:dev` now runs Caddy via podman in place of the old express server (see `scripts/run-proxy.sh`)

## Summary by CodeRabbit

* **New Features**
  * Added Caddy-based routing for authentication, API, OIDC, and AI service requests.
  * Added runtime environment configuration through the `/.env` endpoint.
  * Added Caddy-powered development and production serving, including SPA fallback routing and compressed assets.
  * Added container startup support for trusted Kubernetes and OpenShift certificates.

* **Improvements**
  * Client configuration now uses structured environment values with sensible defaults.
  * Local development now supports dedicated proxy workflows across macOS and Linux.


rszwajko:
@sjd78 
Looks good! One question - why do we need the entrypoint to be a Go app? can we (re)use shell script used in dev mode?

sjd78:
> @sjd78 Looks good! One question - why do we need the entrypoint to be a Go app? can we (re)use shell script used in dev mode?

The PR is using the [RH hardened image](https://images.redhat.com/) "distro-less" Caddy container. There is no shell included. Other than using a side car image, using a statically linked go app to do that same work is the easiest way to do it. That is also why the entrypoint/main.go app doesn't use anything other than standard libs to minimize go build needs.

PR #3609, opened 2026-09-17
:sparkles: Create-run modal: let the creator opt in to agent questions

ibolton336:
**Draft — do not merge before its two dependencies** (see "Order" below).
Merged early, the checkbox would do nothing and say nothing.

## What

konveyor/agentic-controller#242 makes the agent's `ask_user` tool opt-in via
`spec.execution.askUser`. The tool lets an agent stop mid-turn and put a
question to whoever is attached to the run; by design an unanswered question
**fails the run**. It used to be on for every run, so a run nobody was watching
could be failed by a question nobody saw (seen on the ROKS demo cluster: same
Agent and prompt Succeeded, then Failed at turn 38). After #242 a run only gets
the tool if it asks for it — this is how the console asks.

## Changes

- **Create-run modal:** a checkbox under *Supervision mode* — "Let the agent ask
  me questions", with helper text that says the cost plainly: stay on the run's
  details page, a question nobody answers within a few minutes fails the run.
  Unchecked by default, and deliberately **not** prefilled on re-run: every run
  opts in on its own.
- **`createAgentRun`:** sends `execution.askUser: true` only when ticked; the
  key is absent otherwise (so the body is unchanged for everyone who never
  touches it).
- **Contract types:** `ExecutionSpec.askUser`, `CreateRunInput.askUser`.

Workflow runs are not covered: `askUser` there is per stage on the
AgentWorkflow, not on the run.

## Order

1. konveyor/agentic-controller#242 — adds the field, flips the default. **Merged.**
2. **konveyor/tackle2-hub#1139** — bumps
   `github.com/konveyor/agentic-controller/api` past #242. Hub binds the create
   body straight into that Go type (`type AgentRun = agent.AgentRun`, pinned at
   `22fee49`, 2026-08-26) and its `BindJSON` is strict
   (`DisallowUnknownFields`), so until that merges a ticked box would **400 the
   whole create**, not be quietly ignored. This is why the PR is a draft.
3. This PR.

## Verification

- `tsc`, `eslint` clean; `npm run test -w client` — 48 suites / 559 tests pass
- New: `createAgentRun` sends `askUser` only when opted in (and keeps
  `execution: { mode }` byte-identical otherwise); the modal's checkbox is
  unchecked by default and reaches the mutation as `askUser: true` when ticked
- Not yet seen in a browser or against a hub — nothing upstream accepts the
  field yet


ibolton336:
Closing — the bug this chain was about is already fixed without it.

The defect was an unwatched run being failed by an `ask_user` question nobody saw. konveyor/agentic-controller#242 (merged) fixed that on its own by making the tool opt-in, default off. This PR is the other half — a way to opt back *in* from the console — and nobody is asking for that yet. No shipped Agent or AgentWorkflow sets `askUser`, and the checkbox reintroduces the same failure as a self-inflicted one: tick it, walk away, and an unanswered question fails the run.

It also needs two more changes to land first, in order (Hub API bump, operator CRD sync), which is a lot of machinery for something off by default.

Consequence worth knowing: the console's elicitation card in `ChatPanel` stays on `main` but is unreachable for console-created runs, since nothing can turn the tool on through Hub. `ask_user` is `kubectl`-only until this is revived. The branch is kept if that day comes.

Two corrections to this description for whoever picks it up: an unbumped Hub does **not** drop `askUser` silently — `BindJSON` is strict, so the request fails with `400 json: unknown field "askUser"`. And the Order list is missing a step: the operator's AgentRun CRD needs the field too (konveyor/tackle2-operator#623), or the API server prunes it — that is the actual silent failure.

PR #3610, opened 2026-09-17
:bug: Do not offer edits the operator will revert

ibolton336:
## Problem

Overriding the image on a shipped Agent in the UI appears to work, then silently undoes itself.

The operator owns that content. In `konveyor/tackle2-operator`, `roles/tackle/tasks/agentic-default.yml` applies each curated default with:

```yaml
server_side_apply:
  field_manager: agentic-controller-defaults
  force_conflicts: true
```

`force_conflicts: true` re-asserts `spec.image` on every reconcile rather than erroring on the competing field manager, and the value it writes comes from the object's `konveyor.io/related-image` annotation (e.g. `RELATED_IMAGE_AGENT_JAVA`) — so the edit isn't reverted to its previous value, it's replaced by the operator's configured release image. Deletes are re-created, and a companion prune task deletes by the `app.kubernetes.io/managed-by=agentic-controller-defaults` label.

This is deliberate on the operator side — the task file calls it the "CLOBBER POLICY" and says customers wanting their own content should create a new object rather than edit these. The bug is that the UI offered a write the cluster is designed to undo, with no feedback.

## Change

Recognize the marker label and stop offering what cannot stick:

- **Agents and Workflows** — the row action becomes **View**, opening the designer/composer read-only with an inline note explaining the operator owns the object. Deliberately not a disabled Edit: that form is the only place to read an Agent's prompt and params, or a workflow's stage instructions, so it stays reachable.
- **SkillCards and SkillCollections** — Edit is disabled with the same explanation in a tooltip; those tables already have their own View action for inspection.
- **Delete** — disabled for every managed object.
- User-created objects carry no such label and are completely unaffected.

One trap worth calling out for reviewers: `konveyor.io/managed` is a **different** label already in this codebase — it marks an Agent as requiring Hub context, not as operator-owned. Keying off it would have locked down the wrong set of objects. The new helper uses `app.kubernetes.io/managed-by`, and a test pins that distinction so the two don't get conflated later.

## Scope

This is a UI-only fix. Hub still accepts a `PUT` on a managed Agent and returns 200, with the revert happening silently afterwards; closing that properly belongs in Hub (reject the write, or surface ownership so any client can react). Happy to open a tracking issue against `konveyor/tackle2-hub`.

## Verification

- `npx tsc --noEmit` clean
- `npm run lint` clean
- `npm run test` — 50 suites / 564 tests pass, no regressions

New tests: a unit test for the label helper (including that `konveyor.io/managed` does *not* trigger it), and four component tests driving the real kebab menu on the Agents page — View replaces Edit for a shipped default, the editor opens with the image field disabled and no Save button, a user-created Agent stays fully writable, and the disabled Delete does not even open the confirm dialog (asserting only "mutate wasn't called" would have passed vacuously).

## Summary by CodeRabbit

- **New Features**
  - Operator-managed agents and workflows can now be viewed in read-only mode.
  - Operator-managed skills, skill collections, agents, and workflows display explanatory tooltips and prevent editing or deletion.
  - Regular, user-created resources remain fully editable and removable.
  - Added localized labels and guidance for operator-managed resources and defaults.

- **Tests**
  - Added coverage verifying read-only behavior, disabled actions, and continued editing for user-created resources.


ibolton336:
Closing this — it fixes the wrong layer.

The underlying problem is real: the operator applies its curated defaults with `force_conflicts: true` and re-asserts `spec.image` from the `konveyor.io/related-image` annotation on every reconcile, so editing a shipped Agent in the UI silently reverts. But Hub's `PUT /agentic/agents/{name}` returns 204 for that write regardless, and this PR only taught one client to avoid it — `kubectl`, scripts and any future client still hit the same trap.

It also encoded an operator implementation detail (the `agentic-controller-defaults` field-manager name) in the frontend, where it would go stale without notice if the operator's clobber policy changed.

Moving the fix to Hub: reject writes and deletes on operator-owned objects with a 409 naming the reason, so every client gets a real answer, and expose ownership in the GET response so a UI can grey out the action by reading Hub's contract rather than guessing from a label. Tracking in konveyor/tackle2-hub.

